PR #154, opened 2022-05-31
✨ Added Search View for fonts

andretortolano:
Adding a static search view on fonts page (debugger).

Spent a long time trying to get an animated version of this working properly but unfortunately its very difficult to animate content size changes with shadow it seems, so I decided to fall back to more of a simple version of what I was going for.


andretortolano:
should be working now. 

for some reason on API 31 there is an Extra call back from onValueChange on TextField that happens after we lose focus of the TextField, this was setting the previous value of "mono" back to the textField right after we set that to an empty String

iujames (approved):
looks good to me! just a detekt method length issue to clear up.

PR #157, opened 2022-06-06
✨ Added Tests for Analytics

andretortolano:
Included tests for the desired classes in this task.

some changes made in order to better unit test logic.

Edit:

Ended up breaking apart AnalyticsTracker into smaller pieces. 

1. AnalyticsTracker is the original class much simplified
2. AnalyticsQueueProcessor is responsible for holding the pendingActivities and scheduling/cancelling the call to the repository
3. QueueScheduler which handles the scheduler logic (harder to test without it)


appcues/src/test/java/com/appcues/analytics/AutoPropertyDecoratorTest.kt
```diff
@@ -0,0 +1,201 @@
...
+            with(attributes["_identity"] as Map<String, Any>) {
+                assertThat(get("_sessionRandomizer")).isNotEqualTo(sessionRandomizer)
+                sessionRandomizer = get("_sessionRandomizer") as Int
+            }
+        }
+        // when
+        with(autoPropertyDecorator.decorateTrack(event)) {
+            // then
+            with(attributes["_identity"] as Map<String, Any>) {
+                assertThat(get("_sessionRandomizer")).isNotEqualTo(sessionRandomizer)
```
iujames, line 170: suppose there is a very slim chance that this test will randomly fail?

appcues/src/test/java/com/appcues/analytics/AutoPropertyDecoratorTest.kt
```diff
@@ -0,0 +1,201 @@
...
+        val activityRequest = ActivityRequest(userId = "test userId", accountId = "test accountId")
+        // when
+        with(autoPropertyDecorator.decorateIdentify(activityRequest)) {
+            // then
+            assertThat(profileUpdate).hasSize(21)
+        }
+    }
+
+    @Test
+    fun `decorateIdentity SHOULD put autoProperties to profileUpdate map`() {
```
iujames, line 187: does this mean should put `custom properties` to the profileUpdate map?

iujames (approved):
nice! looks like this is about a ~5% bump in code coverage from my local run

appcues/src/main/java/com/appcues/AppcuesCoroutineScope.kt
```diff
@@ -1,17 +1,20 @@
...
 import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlin.coroutines.CoroutineContext
 
 internal class AppcuesCoroutineScope(
-    logcues: Logcues
+    logcues: Logcues,
+    dispatcher: CoroutineDispatcher = Dispatchers.Main
```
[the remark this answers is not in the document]
andretortolano: I will look into it, but I think I it better this way because we can just set the Dispatcher we want.

appcues/src/test/java/com/appcues/analytics/AutoPropertyDecoratorTest.kt
```diff
@@ -0,0 +1,201 @@
...
+import com.google.common.truth.Truth.assertThat
+import io.mockk.every
+import io.mockk.mockk
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
+import org.junit.Test
+
+class AutoPropertyDecoratorTest {
+
+    private val config: AppcuesConfig = mockk<AppcuesConfig>(relaxed = true).apply {
```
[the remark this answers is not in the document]
andretortolano: I'm actually using it to test correct parameter passing on `decorateTrack SHOULD decorate context properties`

appcues/src/test/java/com/appcues/analytics/AutoPropertyDecoratorTest.kt
andretortolano: okay so I did some refactoring, please take a look and see if it makes sense.

I split your rule into two rules, this is how I used in previous jobs too, so they work basically as small "plugins" to test files.


sessonRandomizer test is a tricky one, let me know if you have a better idea how we can do it.

appcues/src/test/java/com/appcues/analytics/AutoPropertyDecoratorTest.kt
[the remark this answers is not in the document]
iujames: oh yeah i didn't mean get rid of the block contents, i just meant the `apply` could be removed (i think) - so like 
```
private val config = mockk<AppcuesConfig>(relaxed = true) {
    every { applicationId } returns "applicationId"
}
```

appcues/src/test/java/com/appcues/rules/MainDispatcherRule.kt
```diff
@@ -0,0 +1,24 @@
...
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
+import kotlinx.coroutines.test.resetMain
+import kotlinx.coroutines.test.setMain
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+
+@Suppress("MemberVisibilityCanBePrivate")
+@OptIn(ExperimentalCoroutinesApi::class)
+class MainDispatcherRule(
+    val dispatcher: TestDispatcher = UnconfinedTestDispatcher(),
```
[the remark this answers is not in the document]
iujames: ah I see - thanks!

appcues/src/main/java/com/appcues/analytics/AnalyticsQueueManager.kt
```diff
@@ -0,0 +1,90 @@
+package com.appcues.analytics
+
+import com.appcues.data.remote.request.ActivityRequest
+import com.appcues.data.remote.request.EventRequest
+import java.util.Timer
+import java.util.TimerTask
+import kotlin.concurrent.schedule
+
+internal class AnalyticsQueueManager {
```
andretortolano, line 9: here I extracted everything related to queue in the analyticsTracker class and made it in its own class. let me know if you would change anything, and also if there is any scenario you remember that you can test this version against to make sure its working the same that would be helpful, but I basically just moved code around to make it cleaner on AnalyticsTracker and also easier to test there. (can comeback to this and also test this class)

appcues/src/main/java/com/appcues/analytics/AnalyticsTracker.kt
```diff
@@ -19,16 +15,9 @@ internal class AnalyticsTracker(
     private val activityBuilder: ActivityRequestBuilder,
     private val experienceLifecycleTracker: ExperienceLifecycleTracker,
     private val analyticsPolicy: AnalyticsPolicy,
+    private val analyticsQueueManager: AnalyticsQueueManager,
```
andretortolano, line 18: a lot of simplification in this class now that we have an extra AnalyticsQueueManager to handle all queue functions

gradle/tools/detekt/detekt-config.yml
```diff
@@ -119,7 +119,7 @@ complexity:
   LongParameterList:
     active: true
     functionThreshold: 6
-    constructorThreshold: 7
+    constructorThreshold: 8
```
andretortolano, line 122: bumped it up because I couldn't think of any other way to solve this in AnalyticsTracker now that it has 7 dependencies. see what you think about it.

iujames: AnalyticsTracker could perhaps remove the `AppcuesRepository` dependency? And then inject that (and coroutine scope) into the new `AnalyticsQueueManager` - since that is the spot where it actually makes the network call now.  Then you also would not need to pass the `send` closure along.  Would that simplify?

appcues/src/main/java/com/appcues/analytics/AnalyticsQueueManager.kt
iujames: this seems like a reasonable split to me, to break out some of the responsibilities into its own class for the queue.

appcues/src/test/java/com/appcues/analytics/AutoPropertyDecoratorTest.kt
```diff
@@ -165,21 +165,13 @@ class AutoPropertyDecoratorTest {
...
-            with(autoPropertyDecorator.decorateTrack(event)) {
-                // then
-                with(attributes["_identity"] as Map<*, *>) {
-                    println("i is $i: ${get("_sessionRandomizer") as Int}")
-                    sessionSets.add(get("_sessionRandomizer") as Int)
-                }
+        with(autoPropertyDecorator.decorateTrack(event)) {
+            // then
+            with(attributes["_identity"] as Map<*, *>) {
+                assertThat(get("_sessionRandomizer") as Int).isEqualTo(10)
```
iujames, line 172: im not sure if this is really testing the reset during SessionStarted is it? could we verify its 10, then update the mock to return a different value, send another session start, then verify the new value?

iujames (approved):
looks good to me! some minor thoughts to consider possible clean up pass.

appcues/src/test/java/com/appcues/analytics/AutoPropertyDecoratorTest.kt
iujames: I thought i replied on this one but maybe it didnt save.  Anyway - splitting into two Rules seems all good to me! Its originally how I had it actually, just combined them since I was always using both.  But if there is a use case for using individually then it makes sense to make the more granular.

appcues/src/main/java/com/appcues/analytics/AnalyticsQueueProcessor.kt
```diff
@@ -0,0 +1,132 @@
...
+    interface QueueScheduler {
+
+        fun schedule(block: () -> Unit)
+        fun cancel()
+    }
+
+    private val pendingActivity = mutableListOf<ActivityRequest>()
+
+    @VisibleForTesting
+    fun queueForTesting(activity: ActivityRequest) {
```
andretortolano, line 30: this is also helpful for testing, so we can add an activity to the list before we call any method.

appcues/src/main/java/com/appcues/analytics/AnalyticsQueueProcessor.kt
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    class AnalyticsQueueScheduler : QueueScheduler {
+
+        private var debounceTimer: TimerTask? = null
+
+        override fun schedule(block: () -> Unit) {
+            // if null, then schedule
+            if (debounceTimer == null) {
+                debounceTimer = Timer().schedule(delay = 10000L) {
```
iujames, line 118: I am surprised that Detekt did not complain about magic number here? but could pull out to a constant.

iujames (approved):
Sorry, I thought this one was already approved and did not mean to leave it lingering so long.  The most recent updates seem fine to me to improve testability 👍

PR #160, opened 2022-06-08
Support Paused state in the StateMachine and logic to render a new Experience if needed

iujames:
This is a Draft PR to discuss a proposed idea for an issue recently uncovered.

**Issue**: since our AppcuesActivity was set to use `noHistory=true` in commit https://github.com/appcues/appcues-android-sdk/commit/5bedc0096b4779589150b3f1584f23eec4f11c5c (PR #141) - any ExperienceAction that might open up a new Activity would inadvertently kill off a running Experience when launched, even if there was no close action and the expectation was that you could return back to the running Experience.  Example: link to URL - which opens the Chrome Tabs Activity - not good.

**Idea**: revert the `noHistory=true` change and support more robust handling of pause/resume in AppcuesActivity.  The StateMachine can be told to go into a Paused state, remembering the actual state it was in before paused.  Then, if a new Experience render comes in while Paused - we can assume that a new Activity on top has triggered new screens/events to qualify this, and the old Paused Experience can be ended in the background before launching the new one.

This is similar to the override code we put in place to allow an event-triggered Experience to override a screen-triggered experience a little while back.  Due to the nature of how the Android Activity system works, and our single StateMachine architecture - this seems a viable approach to continue to allow a single current Experience, but give priority to whatever Activity is topmost/visible.

Simple way to test: use Charles to trigger a flow with mapLocal on the sample app sign-in screen on launch.  Then, run a deep link like 
```
 adb shell am start -W -a android.intent.action.VIEW -d "appcues-example://events"
```
which will trigger a new Activity to be placed on top, new screen view, and a new Experience to show.  

Prior to this change, if `noHistory=true` was removed the manifest - this second Experience would not show, since another Experience was actively running the state machine, hidden in the background.  However, having `noHistory=true` in the manifest is no good - for reasons already noted above.

### DIFF
--- a/appcues/src/main/AndroidManifest.xml
+++ b/appcues/src/main/AndroidManifest.xml
@@ -8,7 +8,6 @@
         <activity
             android:name=".ui.AppcuesActivity"
             android:exported="false"
-            android:noHistory="true"
             android:theme="@style/Appcues.AppcuesActivityTheme" />
     </application>
 
--- a/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
+++ b/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
@@ -15,6 +15,7 @@ import com.appcues.statemachine.State.BeginningStep
 import com.appcues.statemachine.State.EndingExperience
 import com.appcues.statemachine.State.EndingStep
 import com.appcues.statemachine.State.Idling
+import com.appcues.statemachine.State.Paused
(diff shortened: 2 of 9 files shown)


iujames:
fyi @mmaatttt since this touches on a fairly shared state machine concept, but in a way that I believe is only Android specific with the potential need for a `Paused` state - due to how Android Activities work differently than the iOS ViewController system.

andretortolano (approved):
I liked the solution and good usage of the state machine we have.

PR #161, opened 2022-06-09
Add action for requesting review

iujames:
This stacks on top of #160 , since it requires the capability to launch an Activity on top of a running experience.

Using the [Google Play in-app review lib](https://developer.android.com/guide/playcore/in-app-review/kotlin-java) to allow an experience action that can request a review.  Similar to iOS, it's not really guaranteed to show the review, that is up to the internal Play implementation - but we can request.

A new Activity, the InAppReviewActivity is used to launch this review on top of the current experience, with a callback to return when done and allow the Experience action sequence to continue in natural order.

There is a 3 sec timeout in place to return from the review Activity if the Play library never returns a result when asked for the review task.  The Play library only works on a build distributed from the store, like our internal test app builds. It will not work on a build installed directly from APK or emulator - it would just hang without this timeout, as no callback is invoked in those cases, just silent failure, unfortunately.

### DIFF
--- a/appcues/build.gradle
+++ b/appcues/build.gradle
@@ -94,6 +94,9 @@ dependencies {
     implementation "com.google.accompanist:accompanist-webview:0.24.9-beta"
     // Navigation Animated
     implementation "com.google.accompanist:accompanist-navigation-animation:0.24.9-beta"
+    // Play In-App Review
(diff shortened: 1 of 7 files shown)


andretortolano (approved):
Nice, completable deferred for the win

PR #162, opened 2022-06-10
🐛 Fix bug to properly track session reset on the debugger

andretortolano:
very simple bug, debugger needs a special case to handle SessionReset.

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/DebuggerStatusManager.kt
+++ b/appcues/src/main/java/com/appcues/debugger/DebuggerStatusManager.kt
@@ -34,7 +34,7 @@ internal class DebuggerStatusManager(
 
     private var trackingScreens: Boolean? = null
 
-    private var userIdentified: String? = storage.userId.let { it.ifEmpty { null } }
+    private var userIdentified: String? = storage.userId.ifEmpty { null }
 
     private var experienceName: String? = null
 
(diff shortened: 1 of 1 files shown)


appcues/src/main/java/com/appcues/SessionMonitor.kt
```diff
@@ -48,7 +48,10 @@ internal class SessionMonitor(
     }
 
     fun reset() {
+        // reset storage before calling track so it can use updated storage information
+        storage.reset()
         analyticsTracker.track(AnalyticsEvent.SessionReset, null, false)
+        // set sessionId to null AFTER the track or else it wont track anything as checkSession will return false
```
iujames, line 54: I _think_ this might be incorrect.  If you reset before the SessionReset, then that analytic won't actually track correctly - since no user.  I thought the analytic track had to happen before the values get wiped out.

appcues/src/main/java/com/appcues/SessionMonitor.kt
```diff
@@ -48,7 +48,10 @@ internal class SessionMonitor(
     }
 
     fun reset() {
+        // reset storage before calling track so it can use updated storage information
```
andretortolano, line 51: before this change the SessionReset event was being sent with a userId, so I Think it was wrong but would be nice to doublecheck this as well.

iujames (approved):

PR #166, opened 2022-06-23
🐛 Fix bug to properly display the debugger even when activity is fitsystemwindow false

andretortolano:
Initially I tried to research a way to check if the activity was fitting system window or not, but it seems like a impossible task so I moved away from it and went on a different approach.

 Now we are drawing on the decorView.root which has similar effect as we had before but in addition to this I had to manually add top/bottom margins to the ComposeView used by the debugger or else the debugger could go under status/navigation bar icons.

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/ui/DebuggerFloatingActionButton.kt
+++ b/appcues/src/main/java/com/appcues/debugger/ui/DebuggerFloatingActionButton.kt
@@ -52,69 +52,76 @@ internal fun BoxScope.DebuggerFloatingActionButton(
     debuggerState: MutableDebuggerState,
     debuggerViewModel: DebuggerViewModel,
 ) {
-    val resizeBy = animateFloatAsState(
-        when {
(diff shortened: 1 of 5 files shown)


appcues/src/main/java/com/appcues/debugger/ui/DebuggerFloatingActionButton.kt
```diff
@@ -52,70 +52,77 @@ internal fun BoxScope.DebuggerFloatingActionButton(
...
-                debuggerState.isExpanded.targetState -> 2.dp
-                debuggerState.isVisible.currentState -> 4.dp
-                else -> 0.dp
-            }
-        )
+    val resizeBy = resizeAnimatedState(debuggerState)
 
-        Box(
+    // only draw if IntOffset is not null
+    debuggerState.getFabPositionAsIntOffset()?.let { offset ->
```
andretortolano, line 58: also I'm hiding the whole FAB if IntOffset is not valid, this is to avoid some glitch I found while opening Flows

PR #167, opened 2022-06-28
Add support for an AnalyticsListener for hooking in other integrations

iujames:
Similar to a [change coming on iOS ](https://github.com/appcues/appcues-ios-sdk/pull/214), this provides a mechanism for the host application to listen to all analytics being published, and optionally send them to other destinations on the client side as well.  Particularly of use for flow analytics.  An `isInternal` param on the callback can be used to filter to only those events originating inside the SDK (flow and session events) if desired, in the customer app.


andretortolano (approved):
looks good, so this is actually working in sync with events that the debugger should receive right? I was wondering if was there any way to have Appcues subscribe to the same analyticsFlow and filter out undesired events, but this one is a very slc way of doing it too!

iujames:
Yeah using the `analyticsFlow` would be slick... but I think it would currently add a fair bit of complication as I'd need to pass along more information for the `isInternal` flag, as well as add code on the other side to interpret the type of analytics for the `ActivityRequest` object.  Maybe it could be done if the `analyicsFlow` used some sort of wrapper object around the `ActivityRequest` that provided the additional info.

andretortolano (approved):
looks good to me

PR #173, opened 2022-06-29
Add tests for new AnalyticsListener functionality

iujames:
A few test updates to (A) ensure AnalyticsTracker is publishing the expected tracking data to the analyticsFlow SharedFlow and (B) ensure that Appcues is properly broadcasting that tracking data to any attached AnalyticsListener

### DIFF
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -3,6 +3,7 @@ package com.appcues
 import android.app.Activity
 import android.content.Context
 import android.content.Intent
+import androidx.annotation.VisibleForTesting
 import com.appcues.action.ActionRegistry
 import com.appcues.action.ExperienceAction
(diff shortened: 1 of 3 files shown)


andretortolano (approved):
this is valid test for coverage but what we could do later is actually create an instance of Appcues and register the listener, then we can mock so that some events flow through analyticsFlow while we have listener registered. we can then save every incoming event in a list and check the list size, inspect each element to see if they came in the right order and if they are in fact the event we are expecting

iujames:
@andretortolano I would be interested in talking more about that approach.  That's actually what I tried to do first but got into a bit too much complexity with all the dependencies involved in setting up the scope for Appcues.  But I'd like to cycle back and try it out if you have ideas on how to make that work.  I figured I would at least unit test the individual pieces here and assume that the Kotlin SharedFlow behavior is a given to work and connect them.

PR #175, opened 2022-06-30
Update documentation and prepare to use Dokka

iujames:
end result is some hosted docs on github pages https://appcues.github.io/appcues-android-sdk/

there are no functional code changes here, just some doc cleanup and then reorganization of things to (A) avoid anything leaking in public API that is not needed and (B) keeping our public API surface area minimal in terms of packages:

com.appcues
com.appcues.action
com.appcues.trait

a few things got moved from misc spots into `trait` there for UI code related to trait building, which should be a niche concern at least in short term.

a couple other things got moved into `appcues` root (AnalyticsListener, AnalyticsType) to avoid the one-offs in the analytics package and align with other things in the root like ExperienceListener and AppcuesInterceptor


fastlane/docs.sh
```diff
@@ -0,0 +1,41 @@
...
+repoName="appcues-android-sdk"
+initialBranch=$(git rev-parse --abbrev-ref HEAD)
+# versionFile="./Sources/AppcuesKit/Version.swift"
+
+# # get last line in Version.swift
+# versionLine=$(tail -n 1 $versionFile)
+# # split at the =
+# version=$(cut -d "=" -f2- <<< "$versionLine")
+# # remove quotes and spaces
+# version=$(sed "s/[' \"]//g" <<< "$version")
```
iujames, line 12: planning to take another pass on this with a version release script, coming up.  It will handle packaging up the release to maven, adding the release in github, and generating these docs.  Modeling after what was done on iOS.

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -197,14 +210,16 @@ class Appcues internal constructor(koinScope: Scope) {
...
+     *
+     * @param activity The Activity to launch the debugger over.
      */
     fun debug(activity: Activity) {
         debuggerManager.start(activity)
     }
 
     /**
-     * Signals to Appcues that this instance should stop all on going jobs
+     * Signals to Appcues that this instance should stop all ongoing jobs.
```
andretortolano, line 222: ops.

PR #176, opened 2022-07-01
Add release script for versioning and maven central publish

iujames:
stacks on #175 since it includes usages of the docs script

this is modeled after the ios script, which is modeled after Segment - it provides a single command to package up each of our releases

* validate new version is increasing per semver
* update version number in appcues.properties
* generate changelog from git history
* commit and push version change
* make the github release
* publish to maven central (still requires manually releasing the package on sonatype side)
* update docs on github pages for the release

it requires the signing creds (4 variables) be set in the environment as well for the maven part

### DIFF
--- a/fastlane/semver.sh
+++ b/fastlane/semver.sh
@@ -1,0 +1,368 @@
+#!/bin/sh
+
+# MIT License
+# 
+# Copyright (c) 2020 Ariel Rodriguez
+# 
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


fastlane/semver.sh
```diff
@@ -0,0 +1,368 @@
+#!/bin/sh
```
iujames, line 1: this script is an unchanged copy from the ios repo, which was pulled in from the source noted below - it is just for checking that the version is increasing per semver standards

release.sh
```diff
@@ -0,0 +1,147 @@
+#!/bin/sh
```
iujames, line 1: this script is our actual release stuff - all the things noted in the PR description.

PR #178, opened 2022-07-01
Fix shadow rendering

iujames:
I'm not very knowledgable about this part of the code, but studied how we apply shadows a bit, and studied [this other reference](https://gist.github.com/arthurgonzaga/598267f570e38425fc52f97b30e0619d)

It looked like we were applying the shadow radius in the wrong spot - `drawRoundRect` instead of `setShadowLayer`.  Updating this to align with the gist above more closely gives a more desirable result.  I'm not sure if its perfect yet or not.

Note, [compose 1.2.0-alpha06 also introduced a shadow modifier](https://developer.android.com/reference/kotlin/androidx/compose/ui/draw/package-summary#(androidx.compose.ui.Modifier).shadow(androidx.compose.ui.unit.Dp,androidx.compose.ui.graphics.Shape,kotlin.Boolean,androidx.compose.ui.graphics.Color,androidx.compose.ui.graphics.Color)) directly, but it does not support the same options as our data model, and our custom implementation seems more in line with what our model and iOS side use.

end result - android on the left, ios on the right - more aligned than what is seen in the ticket issue.
Screen Shot 2022-07-01 at 12 48 40 PM

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
+++ b/appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
@@ -153,7 +153,7 @@ internal fun Modifier.coloredShadow(
 ) = drawBehind {
 
     val shadowColor = color.toArgb()
-    val transparent = color.copy(alpha = 0.2f).toArgb()
+    val transparent = color.copy(alpha = 0.0f).toArgb()
 
     drawIntoCanvas {
         val paint = Paint()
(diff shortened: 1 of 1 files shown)


appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
```diff
@@ -153,15 +153,15 @@ internal fun Modifier.coloredShadow(
...
+    val transparent = color.copy(alpha = 0.0f).toArgb()
 
     drawIntoCanvas {
         val paint = Paint()
         val frameworkPaint = paint.asFrameworkPaint()
         frameworkPaint.color = transparent
 
         frameworkPaint.setShadowLayer(
-            4.dp.toPx(),
+            radius.toPx(),
```
andretortolano, line 164: nice catch, I guess this wasnt "THAT" dynamic with a fixed 4.dp value

andretortolano (approved):

PR #179, opened 2022-07-01
Update to Compose 1.2.0-rc03

iujames:
Update:  the workaround in comment 11 here https://issuetracker.google.com/issues/236612358#comment seems to work for us - so I think we can move forward with this until Room 2.4.3 is released for standard Kotlin 1.7.0 compat.

--------------

Marking this one as a draft, as it unfortunately hits a snag.  I was trying to stay current on Compose updates moving toward stable.

Compose 1.2.0-rc03 appears to require the, now stable, compose compiler 1.2.0.  Related - [Compose is now versioning libraries independently](https://android-developers.googleblog.com/2022/06/independent-versioning-of-Jetpack-Compose-libraries.html)

The 1.2.0 compiler requires Kotlin 1.7.0

Updating to Kotlin 1.7.0 breaks the Room code generation though, unfortunately.  Room 2.5.0-alpha02 appears to resolve the issue, but I was hoping to move towards Compose stable, not go backward to another alpha version of another lib 😞 

Looks like this is starting to pop up as a bug though with both [Kotlin](https://youtrack.jetbrains.com/issue/KT-52996/After-updating-kotlin-plugin-in-android-studio-facing-problem-with-Room-Database) and [Google Room](https://issuetracker.google.com/issues/236612358) teams. So we'll see how that plays out.

### DIFF
--- a/appcues/build.gradle
+++ b/appcues/build.gradle
@@ -46,7 +46,7 @@ android {
     }
 
     composeOptions {
-        kotlinCompilerExtensionVersion = "1.2.0-rc02"
+        kotlinCompilerExtensionVersion = "1.2.0"
     }
 
     packagingOptions {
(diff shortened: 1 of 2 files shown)


andretortolano (approved):
lets do it.

PR #183, opened 2022-07-12
🐛 Properly set shadow using cornerRadius for shape and radius for shadow layer

andretortolano:
we werent using properly the radius property for shadow, no I added to the shadow method an extra parameter that takes in consideration the cornerRadius so that we can properly set shadow behind the content its clipped on.

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
+++ b/appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
@@ -133,6 +133,7 @@ internal fun Modifier.styleShadow(style: ComponentStyle?, isDark: Boolean): Modi
         if (style?.shadow != null) Modifier.coloredShadow(
             color = style.shadow.color.getColor(isDark),
             radius = style.shadow.radius.dp,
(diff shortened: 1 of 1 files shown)


appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
```diff
@@ -133,6 +133,7 @@ internal fun Modifier.styleShadow(style: ComponentStyle?, isDark: Boolean): Modi
         if (style?.shadow != null) Modifier.coloredShadow(
             color = style.shadow.color.getColor(isDark),
             radius = style.shadow.radius.dp,
+            cornerRadius = style.cornerRadius.dp,
```
iujames, line 136: double checking - this is always non-null, defaults to 0dp corner radius right?

iujames (approved):
nice fix!

appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
andretortolano: image
yes

PR #186, opened 2022-07-14
✨ Added Box primitive

andretortolano:
* Added Box primitive to match IOS behavior
* Fixed Compose Previews

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/component/AppcuesBoxComponent.kt
+++ b/appcues/src/main/java/com/appcues/ui/component/AppcuesBoxComponent.kt
@@ -1,0 +1,115 @@
+package com.appcues.ui.component
+
+import androidx.compose.foundation.interaction.MutableInteractionSource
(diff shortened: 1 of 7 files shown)


appcues/src/main/java/com/appcues/ui/component/AppcuesBoxComponent.kt
```diff
@@ -0,0 +1,115 @@
...
+        contentAlignment = style.getBoxAlignment(),
+    ) {
+
+        ProvideTextStyle(style.getTextStyle(LocalContext.current, isSystemInDarkTheme())) {
+            items.forEach { it.Compose() }
+        }
+    }
+}
+
+private val items = arrayListOf(
```
iujames, line 53: does this get conditionally compiled in preview only? or can it be set up that way?

iujames (approved):
sorry, this one slipped past me on Friday - looks good!

PR #188, opened 2022-07-19
✨ Added Support for actions after flow is completed

andretortolano:
* I added actionProcessor as the StateMachine dependency.
* Added Tests to cover this new use case of the statemachine

### DIFF
--- a/appcues/src/main/java/com/appcues/action/ActionKoin.kt
+++ b/appcues/src/main/java/com/appcues/action/ActionKoin.kt
@@ -13,7 +13,7 @@ import org.koin.dsl.ScopeDSL
 internal object ActionKoin : KoinScopePlugin {
 
     override fun ScopeDSL.install() {
-        scoped { ActionRegistry(scope = get(), logcues = get()) }
+        scoped { ActionRegistry(scope = get()) }
         scoped { ActionProcessor(scope = get()) }
 
         factory { params ->
--- a/appcues/src/main/java/com/appcues/action/ActionKoin.kt
+++ b/appcues/src/main/java/com/appcues/action/ActionKoin.kt
@@ -14,7 +14,7 @@ internal object ActionKoin : KoinScopePlugin {
 
     override fun ScopeDSL.install() {
         scoped { ActionRegistry(scope = get(), logcues = get()) }
-        scoped { ActionProcessor(scope = get(), appcuesCoroutineScope = get()) }
+        scoped { ActionProcessor(scope = get()) }
 
         factory { params ->
             CloseAction(
@@ -25,8 +25,8 @@ internal object ActionKoin : KoinScopePlugin {
 
         factory { params ->
             LinkAction(
-                scope = this,
                 config = params.getOrNull(),
+                linkOpener = get(),
             )
         }
 
--- a/appcues/src/main/java/com/appcues/action/ActionKoin.kt
+++ b/appcues/src/main/java/com/appcues/action/ActionKoin.kt
@@ -14,7 +14,7 @@ internal object ActionKoin : KoinScopePlugin {
 
     override fun ScopeDSL.install() {
         scoped { ActionRegistry(scope = get(), logcues = get()) }
-        scoped { ActionProcessor(appcues = get(), appcuesCoroutineScope = get()) }
+        scoped { ActionProcessor(scope = get(), appcuesCoroutineScope = get()) }
 
         factory { params ->
             CloseAction(
@@ -25,8 +25,8 @@ internal object ActionKoin : KoinScopePlugin {
 
         factory { params ->
             LinkAction(
+                scope = this,
                 config = params.getOrNull(),
-                linkOpener = get(),
             )
         }
 
--- a/appcues/src/main/java/com/appcues/action/ActionProcessor.kt
+++ b/appcues/src/main/java/com/appcues/action/ActionProcessor.kt
@@ -1,23 +1,27 @@
 package com.appcues.action
 
+import com.appcues.Appcues
 import com.appcues.AppcuesCoroutineScope
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.launch
+import org.koin.core.component.KoinScopeComponent
+import org.koin.core.component.inject
 import org.koin.core.scope.Scope
 
 // Responsible for queueing up any ExperienceActions to run in sequential order,
 // to allow for things like closing current experience + launching a new experience, for example.
-internal class ActionProcessor(
-    private val scope: Scope,
-    private val appcuesCoroutineScope: AppcuesCoroutineScope,
-) {
+internal class ActionProcessor(override val scope: Scope) : KoinScopeComponent {
+
+    // lazy initialization injection to avoid circular dependency
+    private val appcues: Appcues by inject()
+    private val appcuesCoroutineScope: AppcuesCoroutineScope by inject()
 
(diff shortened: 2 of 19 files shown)


appcues/src/main/java/com/appcues/action/ActionProcessor.kt
```diff
@@ -1,14 +1,14 @@
...
 import com.appcues.AppcuesCoroutineScope
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.launch
+import org.koin.core.scope.Scope
 
 // Responsible for queueing up any ExperienceActions to run in sequential order,
 // to allow for things like closing current experience + launching a new experience, for example.
 internal class ActionProcessor(
-    private val appcues: Appcues,
+    private val scope: Scope,
```
andretortolano, line 11: had to change this so we avoid circular dependency.

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -57,12 +58,14 @@ internal class ExperienceLifecycleTracker(
                         trackLifecycleEvent(StepCompleted(it.experience, it.flatStepIndex))
                     }
                     is EndingExperience -> {
-                        if (it.markComplete || it.flatStepIndex == it.experience.flatSteps.count() - 1) {
+                        if (it.isExperienceCompleted()) {
                             // if ending on the last step OR an action requested it be considered complete explicitly,
                             // track the experience_completed event
+                            Log.i("Appcues", "Experience ${it.experience.name} Completed")
```
andretortolano, line 64: logs will be removed.

appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
```diff
@@ -40,6 +43,10 @@ internal class ExperienceMapper(
             priority = priority,
             type = from.type,
             publishedAt = from.publishedAt,
+            completionActions = arrayListOf<ExperienceAction>().apply {
```
andretortolano, line 46: this is where I'm mapping from the remote data and creating the ExperienceActions as a list. (the order here matters)

appcues/src/main/java/com/appcues/statemachine/State.kt
```diff
@@ -24,6 +25,11 @@ internal sealed class State {
         // and it should be set to null if a dismiss is not requested (i.e. moving to next step in same container)
         val dismissAndContinue: (() -> Unit)?,
     ) : State()
-    data class EndingExperience(val experience: Experience, val flatStepIndex: Int, val markComplete: Boolean) : State()
+
+    data class EndingExperience(val experience: Experience, val flatStepIndex: Int, val markComplete: Boolean) : State() {
+        // this defines whether the experience was completed or dismissed
+        fun isExperienceCompleted() = markComplete || flatStepIndex == experience.flatSteps.count() - 1
```
andretortolano, line 31: moved this logic to the data classe so it can be reused in two places.

appcues/src/main/java/com/appcues/statemachine/Transitions.kt
```diff
@@ -144,7 +148,7 @@ internal interface Transitions {
     }
 
     fun EndingExperience.fromEndingExperienceToIdling(action: Reset): Transition {
-        return Transition(Idling)
+        return Transition(Idling, if (isExperienceCompleted()) ProcessActions(experience.completionActions) else null)
```
andretortolano, line 151: if experience is completed we ProcessActions or else we do nothing.

appcues/src/main/java/com/appcues/action/ActionKoin.kt
```diff
@@ -14,7 +14,7 @@ internal object ActionKoin : KoinScopePlugin {
 
     override fun ScopeDSL.install() {
         scoped { ActionRegistry(scope = get(), logcues = get()) }
-        scoped { ActionProcessor(appcues = get(), appcuesCoroutineScope = get()) }
+        scoped { ActionProcessor(scope = get(), appcuesCoroutineScope = get()) }
```
iujames, line 17: this one (and the line above) - if we we are passing the scope in to avoid circular dependencies - maybe we should just not pass any other koin scope dependencies in any longer? Then resolve them as injected props with scope.inject()? It feels strange to have a mix in the same class.

appcues/src/main/java/com/appcues/action/appcues/LaunchExperienceAction.kt
```diff
@@ -9,7 +9,14 @@ internal class LaunchExperienceAction(
     override val config: AppcuesConfigMap,
 ) : ExperienceAction {
 
+    constructor(experienceId: String) : this(
+        hashMapOf<String, Any>().apply {
+            this["experienceID"] = experienceId
+        }
+    )
```
iujames, line 16: nice way to tuck this away 👍

appcues/src/main/java/com/appcues/action/ActionKoin.kt
andretortolano: will do thanks

appcues/src/main/java/com/appcues/action/ActionKoin.kt
```diff
@@ -14,7 +14,7 @@ internal object ActionKoin : KoinScopePlugin {
 
     override fun ScopeDSL.install() {
         scoped { ActionRegistry(scope = get(), logcues = get()) }
-        scoped { ActionProcessor(appcues = get(), appcuesCoroutineScope = get()) }
+        scoped { ActionProcessor(scope = get()) }
```
iujames, line 17: not part of this PR, but the `logcues` thing on the ActionRegistry line above is similar - mix of injection patterns

appcues/src/main/java/com/appcues/action/appcues/LinkAction.kt
```diff
@@ -7,16 +7,25 @@ import com.appcues.data.model.AppcuesConfigMap
...
 
     companion object {
+
         const val TYPE = "@appcues/link"
     }
 
+    constructor(redirectUrl: String, linkOpener: LinkOpener) : this(
+        hashMapOf<String, Any>().apply {
+            this["url"] = redirectUrl
+        },
```
iujames, line 25: nit - is the apply needed here? 
```
hashMapOf<String, Any>("url" to redirectUrl)
```

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -57,7 +57,7 @@ internal class ExperienceLifecycleTracker(
                         trackLifecycleEvent(StepCompleted(it.experience, it.flatStepIndex))
                     }
                     is EndingExperience -> {
-                        if (it.markComplete || it.flatStepIndex == it.experience.flatSteps.count() - 1) {
+                        if (it.isExperienceCompleted()) {
```
iujames, line 60: oh good catch re-using this

iujames (approved):

appcues/src/main/java/com/appcues/action/appcues/LinkAction.kt
andretortolano: oh you are right

appcues/src/main/java/com/appcues/action/ActionKoin.kt
andretortolano: I will fix this as well

PR #189, opened 2022-07-22
✨ Added Support for backgroundImage property for every primitive type

andretortolano:
Far too long PR for what it was intended but I will try to go over parts of the code to explain what changes I made and why.

TL;DL of this is, Coil doesnt expose its core modifier, so to apply the backgroundImage property not only for verticalStack but for other primitives types I had to go over some refactoring and ended up doing some cleanup too.

there was some code duplication between primitives as well, I got this opportunity to make it cleaner too.


appcues/src/main/java/com/appcues/data/mapper/styling/StyleMapperExt.kt
```diff
@@ -0,0 +1,22 @@
+package com.appcues.data.mapper.styling
+
+import com.appcues.data.model.styling.ComponentStyle.ComponentHorizontalAlignment
+import com.appcues.data.model.styling.ComponentStyle.ComponentVerticalAlignment
+
+internal fun String?.toComponentVerticalAlignment(): ComponentVerticalAlignment? {
```
andretortolano, line 6: moved both methods to an Extension file for reusability.

appcues/src/main/java/com/appcues/data/remote/response/styling/StyleSizeResponse.kt
```diff
@@ -3,7 +3,7 @@ package com.appcues.data.remote.response.styling
 import com.squareup.moshi.JsonClass
 
 @JsonClass(generateAdapter = true)
-internal data class SizeResponse(
+internal data class StyleSizeResponse(
```
andretortolano, line 6: noticed class name was not following name convention

appcues/src/main/java/com/appcues/trait/appcues/ModalTrait.kt
```diff
@@ -28,7 +28,7 @@ internal class ModalTrait(
 
     @Composable
     override fun WrapContent(content: @Composable (hasFixedHeight: Boolean, contentPadding: PaddingValues?) -> Unit) {
-        val windowInfo = rememberWindowInfo()
+        val windowInfo = rememberAppcuesWindowInfo()
```
andretortolano, line 31: renamed WindowInfo to AppcuesWindowInfo to ensure we dont confuse it with android System WindowInfo class

andretortolano: also makes it clear its a custom object in this case since both have same concept behind it

appcues/src/main/java/com/appcues/ui/extensions/ImageLoaderExt.kt
```diff
@@ -0,0 +1,68 @@
...
+import coil.decode.SvgDecoder
+import coil.request.ImageRequest
+import coil.size.Scale
+import coil.size.Size
+import com.appcues.data.model.styling.ComponentContentMode
+import com.appcues.data.model.styling.ComponentContentMode.FILL
+import com.appcues.data.model.styling.ComponentContentMode.FIT
+import com.appcues.data.model.styling.ComponentSize
+
+internal fun Context.getImageLoader(): ImageLoader {
```
andretortolano, line 24: added image loader related calls here so we can reuse both for ImagePrimitive and for backgroundImage property.

appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
```diff
@@ -33,7 +33,10 @@ import com.appcues.data.model.styling.ComponentSize
 import com.appcues.data.model.styling.ComponentStyle
 import java.util.UUID
 
-internal fun Modifier.primitiveStyle(
+/**
+ * Used by ExperiencePrimitive Compose
+ */
+internal fun Modifier.outerPrimitiveStyle(
```
andretortolano, line 39: had to break it down into outer and inner modifier, so we can fit in compositions in between (like backgroundImage)

appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
```diff
@@ -70,16 +83,25 @@ private fun Modifier.styleBackground(
     isDark: Boolean,
     defaultColor: Color? = null
 ) = this.then(
+    Modifier
```
andretortolano, line 86: I made this change so its in sync with IOS, instead of picking one of the two we now apply whatever we have in sequence to make sure we can mix those properties if we want to.

also this was supposed to have a 3rd modifier called styleBackgroundImage, but this is where I could not make it work  the way I wanted to, still good change to match IOS behavior.

appcues/src/main/java/com/appcues/ui/extensions/StyleComponentExt.kt
```diff
@@ -176,6 +176,13 @@ internal fun ComponentStyle.getHorizontalAlignment(): Alignment.Horizontal {
 internal fun ComponentStyle?.getBoxAlignment(): Alignment {
     if (this == null) return Alignment.Center
 
+    return getBoxAlignment(horizontalAlignment, verticalAlignment)
+}
+
+internal fun getBoxAlignment(
```
andretortolano, line 182: ComponentStyle.getBoxAlignment used by style.horizontal style.vertical Alignment

we needed a method that could receive both parameters because of backgroundImage.horizontal backgroundImage.verticalAlignment

appcues/src/main/java/com/appcues/ui/primitive/BoxPrimitive.kt
```diff
@@ -17,33 +14,17 @@ import com.appcues.data.model.styling.ComponentColor
...
 import com.appcues.ui.extensions.getBoxAlignment
 import com.appcues.ui.extensions.getTextStyle
-import com.appcues.ui.extensions.primitiveStyle
-import com.appcues.ui.theme.AppcuesPreview
+import com.appcues.ui.theme.AppcuesPreviewPrimitive
 import java.util.UUID
 
 @Composable
-internal fun BoxPrimitive.Compose() {
+internal fun BoxPrimitive.Compose(modifier: Modifier) {
```
andretortolano, line 23: all primitives now receives a modifier it should apply to the parent Compose (in this case is the box on line 24), this ensures that Modifier.innerStylePrimitive is being applied without code duplication.

appcues/src/main/java/com/appcues/ui/primitive/BoxPrimitive.kt
```diff
@@ -17,33 +14,17 @@ import com.appcues.data.model.styling.ComponentColor
...
     Box(
-        modifier = Modifier.primitiveStyle(
```
andretortolano, line 33: this entire block was moved to ExperiencePrimitive.Compose

appcues/src/main/java/com/appcues/ui/primitive/ExperiencePrimitive.kt
```diff
@@ -0,0 +1,100 @@
...
+import com.appcues.ui.extensions.getImageLoader
+import com.appcues.ui.extensions.getImageRequest
+import com.appcues.ui.extensions.innerPrimitiveStyle
+import com.appcues.ui.extensions.outerPrimitiveStyle
+import com.appcues.ui.extensions.toImageAsyncContentScale
+import com.appcues.ui.utils.rememberBlurHashDecoded
+
+@Composable
+internal fun ExperiencePrimitive.Compose() {
+    Box(
```
andretortolano, line 38: every primitive has its down box that applies the outerPrimitiveStyle (getRole for each type too on line 70)

and also calls BackgroundImage on line 53 that will compose the image matchingParentSize below the primitive if backgroundImage is present in style.

appcues/src/main/java/com/appcues/ui/primitive/ExperiencePrimitive.kt
```diff
@@ -0,0 +1,100 @@
...
+                role = getRole()
+            ),
+            isDark = isSystemInDarkTheme()
+        ),
+        contentAlignment = Alignment.Center
+    ) {
+        BackgroundImage(style)
+
+        with(this@Compose) {
+            val modifier = Modifier.innerPrimitiveStyle(this)
```
andretortolano, line 56: get innerPrimitiveStyle to pass as parameter.

appcues/src/main/java/com/appcues/ui/theme/AppcuesTheme.kt
```diff
@@ -50,9 +52,9 @@ private fun getColors(isDark: Boolean): Colors {
  * This is supposed to be used during inspection mode (Preview) only.
  */
 @Composable
-internal fun AppcuesPreview(
+internal fun AppcuesPreviewPrimitive(
```
andretortolano, line 55: changing the Preview to read the primitive as ExperiencePrimitive and call ExperiencePrimitive.Compose in order to apply all modifiers properly.

appcues/src/main/java/com/appcues/ui/utils/BlurHashDecoder.kt
```diff
@@ -17,15 +17,26 @@
...
+import androidx.compose.runtime.remember
 import kotlin.math.cos
 import kotlin.math.pow
 import kotlin.math.withSign
 
-internal object BlurHashDecoder {
+private const val PLACEHOLDER_SIZE_PX = 32
+
+@Composable
+internal fun rememberBlurHashDecoded(blurHash: String?): Bitmap? {
```
andretortolano, line 33: blurHash being reused by both ImagePrimitive and BackgroundImage

appcues/src/main/java/com/appcues/ui/primitive/ExperiencePrimitive.kt
iujames: nice to centralize those outerPrimitiveStyle items!

iujames (approved):
This all looks good - a lot of good cleanup and added flexibility.  

When I see a significant code change like this, especially nearing release timelines, a few thoughts come to mind:

1. is it justified? yes, I believe so - this is solid refactoring to meet the objective of the Hero component
2. is it risky? this feels moderate - most of this appears to be reorganizing code, not change a ton of core logic (some, yes).  It's UI layer nearly exclusively, which also registers lower on risk
3. how can we test it?  This is a spot where it would be great to have more UI testing and snapshot stuff in place - but we have some! and more coming . So I would recommend definitely taking a close look at the primitive tests we do have, and all the previews, and making extra sure that nothing got accidentally looked over.  Then, we'll want Ben to do a solid round of general experience regression testing I'd think.

It still feels a bit strange to me to have the Preview code complied into the resulting release builds.  Would be nice to find a post processing mechanism to cleanly strip that out, proguard or other.

overall - great work! very detailed and thoughtful to get this all working!

andretortolano:
I will trigger a android-example build from this branch before merging into main, and I will test it some more and ask Ben for some help with testing.

I agree that at this point it is risky to make this huge amount of change, it really took some time to make that because I was really trying to avoid any major refactory, but since applying the backgroundImage as a modifier wasnt possible, that was the path I had to take for this one

PR #190, opened 2022-07-25
Add support for component snapshot testing

iujames:
Posting this in draft mode for review, as it needs to rebase after #189 goes in and make changes accordingly.  But these are the basic changes needed to support the first round of snapshot testing work.

1. Added a `SnapshotTest` function in the debug source set only - will not ship with release. This allows for the host app in the verification suite (another repo/project) to pass in some JSON and get back a Composable.  iOS uses the concept of `@testable import` to get to internals in the SDK for rendering components, but this is not available on Android.  This method would let us achieve a similar capability though in debug builds.

2. Update various style properties from `Int` to `Double` to match the spec - involved updating several spots in `@Preview` usages.


appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
```diff
@@ -84,7 +85,7 @@ internal fun Modifier.styleBorder(
     style: ComponentStyle,
     isDark: Boolean
 ) = this.then(
-    if (style.borderWidth != null && style.borderWidth != 0 && style.borderColor != null) {
+    if (style.borderWidth != null && style.borderWidth != 0.0 && style.borderColor != null) {
```
iujames, line 88: thinking we need to introduce some more robust value comparison for floating point here, rather than using `== 0.0` (and others below)

andretortolano: need to see if there is a way to better do this comparison

andretortolano (approved):

appcues/src/main/java/com/appcues/util/DoubleExt.kt
```diff
@@ -0,0 +1,7 @@
+package com.appcues.util
+
+import kotlin.math.abs
+
+private const val TOLERANCE = 0.000001
+internal infix fun Double.eq(other: Double) = abs(this - other) <= TOLERANCE
+internal infix fun Double.ne(other: Double) = abs(this - other) > TOLERANCE
```
iujames, line 7: @andretortolano what do you think of this approach? added these two infix ops for equal and not equal double comparison with tolerance.  Used in `ModifierExt` and `SharedModal` above.

andretortolano: thats cool

andretortolano (approved):
implementing double.eq not qe is a good solution!

appcues/src/main/java/com/appcues/util/DoubleExt.kt
iujames: i think `ne` is not a very clear name - maybe `noteq` ?

PR #191, opened 2022-07-26
🐛 Adding StepGroupPageIndex as step reference for swipe trait

andretortolano:
crazy bug to find at this point in the project.

StepIndex worked for simple flows, but we needed a way to find the correct step index based on the index value from within a step group.


iujames (approved):
very glad there was some testing done on more advanced experiences to find this!

PR #192, opened 2022-07-27
Add support for mocking Coil ImageLoader for testing

iujames:
This is a new approach that I've found workable for loading images from our snapshot testing library - allow an optional override of the Coil `ImageLoader` that gets send down into the composition using a `CompsitionLocalProvider`.  By default, it's null, and the normal logic stays the same.  In tests, we can pass in our own ImageLoader that loads static resources synchronously.  Will post the updates to the test code in the spec repo.

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/primitive/ExperiencePrimitive.kt
+++ b/appcues/src/main/java/com/appcues/ui/primitive/ExperiencePrimitive.kt
@@ -23,6 +23,7 @@ import com.appcues.data.model.ExperiencePrimitive.VerticalStackPrimitive
 import com.appcues.data.model.styling.ComponentStyle
 import com.appcues.ui.LocalAppcuesActionDelegate
 import com.appcues.ui.LocalAppcuesActions
+import com.appcues.ui.LocalImageLoader
 import com.appcues.ui.extensions.PrimitiveGestureProperties
 import com.appcues.ui.extensions.blurHashPlaceholder
 import com.appcues.ui.extensions.getBoxAlignment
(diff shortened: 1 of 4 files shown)


andretortolano (approved):

appcues/src/main/java/com/appcues/ui/primitive/ExperiencePrimitive.kt
```diff
@@ -90,10 +91,11 @@ internal fun BoxScope.BackgroundImage(style: ComponentStyle) {
                 modifier = Modifier.matchParentSize(),
                 model = context.getImageRequest(imageUrl, contentMode),
                 contentDescription = null,
-                imageLoader = context.getImageLoader(),
+                imageLoader = LocalImageLoader.current ?: context.getImageLoader(),
                 placeholder = context.blurHashPlaceholder(decodedBlurHash, intrinsicSize),
                 contentScale = contentMode.toImageAsyncContentScale(),
-                alignment = getBoxAlignment(horizontalAlignment, verticalAlignment)
+                alignment = getBoxAlignment(horizontalAlignment, verticalAlignment),
+                error = context.blurHashPlaceholder(decodedBlurHash, intrinsicSize),
```
iujames, line 98: @andretortolano oh meant to call out - this was a minor fix i noticed in the tests - when an image fails to load, we'll now just continue to show the blurHash, matching iOS. (same thing line 28 in ImagePrimitive below)

andretortolano: nice catch, didnt thought of that

PR #194, opened 2022-07-27
✨ Added Auto refresh on shake for preview flows

andretortolano:
I would love your input in this one.

I basically implemented our version of shake gesture detection and as a plus I added vibration when the gesture is detected.

for vibration to work it needs a permission just like internet, its not invasive but its something we need to declare in the manifest.

give it a shot and let me know if its hard/easy to shake it 🕺


appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
```diff
@@ -94,7 +98,10 @@ internal class AppcuesViewModel(
             val stepIndexInContainer = stepIndexLookup[flatStepIndex]
             // if both are valid ids we return Rendering else null
             if (containerId != null && stepIndexInContainer != null) {
-                Rendering(stepContainers[containerId], stepIndexInContainer)
+                // set the current experience id for later usage on refresh preview
+                currentExperienceId = id
```
iujames, line 102: quick initial scan - maybe we should only set this value if we know it is a preview (`published.not()`) just in case?

iujames (commented):
will check it out on device.  I was actually thinking about haptic feedback when @mmaatttt brought this up a few days back also.  Where I landed there was that I _think_ it is unnecessary since there is already clear visual feedback of an experience dismissing and re-appearing.  But 🤷‍♂️ I'll try it out on the device.

appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
andretortolano: check it again to see if its easier to shake and also see if setting the experienceId in Rendering looks nicer.

iujames: @andretortolano this is feeling pretty good to me - It's still a little bit harder on my Pixel 2 XL than I'm used to on the iphone, have to do sort of a hard 90 degree shake down, but it works. And I'd really like to see what others thing since this may just be a device specific thing - this is certainly an older phone.

andretortolano: okay!

PR #196, opened 2022-07-28
🐛 Properly handle negative values on composition layer

andretortolano:
easier than we thought, the standard PaddingValuesModifier had a guard preventing negative values, I basically copied it and removed that limitation.

in addition to this for paddings if builder for some reason sends us negative value I'm normalizing to 0 to ensure we dont get the previous IllegalArgumentException error.


appcues/src/main/java/com/appcues/ui/utils/AppcuesPaddings.kt
```diff
@@ -0,0 +1,63 @@
...
+import androidx.compose.ui.platform.InspectorInfo
+import androidx.compose.ui.platform.InspectorValueInfo
+import androidx.compose.ui.platform.debugInspectorInfo
+import androidx.compose.ui.unit.Constraints
+import androidx.compose.ui.unit.constrainHeight
+import androidx.compose.ui.unit.constrainWidth
+import androidx.compose.ui.unit.offset
+
+@Stable
+internal fun Modifier.appcuesPaddings(paddingValues: PaddingValues) =
```
iujames, line 19: havent studied details - would it be possible at all to just coerce to zero but still use the default `.padding` for this one - delegate into it? or do we need to reimplement here.

iujames (approved):

appcues/src/main/java/com/appcues/ui/utils/AppcuesPaddings.kt
andretortolano: will look into it

andretortolano: did you get to see the changes I've made?

appcues/src/main/java/com/appcues/ui/extensions/StyleComponentExt.kt
```diff
@@ -27,10 +27,10 @@ internal fun ComponentStyle.getMargins() = PaddingValues(
 )
 
 internal fun ComponentStyle.getPaddings() = PaddingValues(
-    start = paddingLeading.dp,
-    bottom = paddingBottom.dp,
-    top = paddingTop.dp,
-    end = paddingTrailing.dp,
+    start = paddingLeading.dp.coerceAtLeast(0.dp),
```
andretortolano, line 30: this is used to get padding values from component style. on every modifier.

iujames (approved):

PR #197, opened 2022-08-01
Prevent extra page change notifications during programmatic scroll in Pager

iujames:
This additional safeguard appears to fix the issue noted in the ticket, for me.  The root cause was that a programmatic transition from page 3 back to page 1 in a modal was causing a page change event on page 2 along the way - this would then get processed like user interaction, sent to the state machine, and update the view to page 2 erroneously.


andretortolano (approved):
I Approve!

PR #199, opened 2022-08-01
Update to use safe cast in Debugger

iujames:
Discovered during some UI testing that we had this cast to `FragmentActivity` in the `AppcuesDebuggerManager` that may not always be valid - updating this to use `as?` to be safe.

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/AppcuesDebuggerManager.kt
+++ b/appcues/src/main/java/com/appcues/debugger/AppcuesDebuggerManager.kt
@@ -80,7 +80,7 @@ internal class AppcuesDebuggerManager(context: Context, private val koinScope: S
         // control native back press by enabling and disabling on our side
         setDebuggerBackPressCallback(activity)
         // also to make sure that if they change fragment, we will register again after the new fragment is attached
-        (activity as FragmentActivity).supportFragmentManager.addFragmentOnAttachListener { _, _ ->
+        (activity as? FragmentActivity)?.supportFragmentManager?.addFragmentOnAttachListener { _, _ ->
             setDebuggerBackPressCallback(activity)
         }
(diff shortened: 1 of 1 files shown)


appcues/src/main/java/com/appcues/debugger/AppcuesDebuggerManager.kt
```diff
@@ -80,7 +80,7 @@ internal class AppcuesDebuggerManager(context: Context, private val koinScope: S
         // control native back press by enabling and disabling on our side
         setDebuggerBackPressCallback(activity)
         // also to make sure that if they change fragment, we will register again after the new fragment is attached
-        (activity as FragmentActivity).supportFragmentManager.addFragmentOnAttachListener { _, _ ->
+        (activity as? FragmentActivity)?.supportFragmentManager?.addFragmentOnAttachListener { _, _ ->
```
andretortolano, line 83: was there any specific version where this cast didnt work? I wasnt really aware of it. nice

iujames: In the UI test project, the MainActivity derives from ComponentActivity - I believe since its a Compose-only app - so this cast fails.

andretortolano: well possibly the back navigation is not working then.

andretortolano (approved):

appcues/src/main/java/com/appcues/debugger/AppcuesDebuggerManager.kt
iujames: ok, ill merge this to avoid the crash, but look into that.

PR #202, opened 2022-08-02
🐛 Moved all read/put from main to io thread on Storage

andretortolano:
arranged the gettter/setters to be main thread compliancy.

### DIFF
--- a/appcues/src/main/java/com/appcues/Storage.kt
+++ b/appcues/src/main/java/com/appcues/Storage.kt
@@ -2,6 +2,7 @@ package com.appcues
 
 import android.content.Context
 import android.content.SharedPreferences
+import android.os.StrictMode
 import java.util.Date
 import java.util.UUID
 
@@ -9,6 +10,7 @@ internal class Storage(
     context: Context,
     config: AppcuesConfig
 ) {
+
     private enum class Constants(val rawVal: String) {
         DeviceId("appcues.deviceId"),
         UserId("appcues.userId"),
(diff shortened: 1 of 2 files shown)


appcues/src/main/java/com/appcues/Storage.kt
```diff
@@ -21,31 +25,85 @@ internal class Storage(
         context.getSharedPreferences("com.appcues.storage.${config.applicationId}", Context.MODE_PRIVATE)
 
     var deviceId: String
```
andretortolano, line 27: I repeated the same logic for all propertyes, basically created a private _xxxxx: T property we get get() from.

on set we update this value + update it on sharedPreference (following previous rules you had in place for each)

one thing I tried to do was to refactory this "block" of code into a property delegate, but it ended up being a lot more code, more complex than it should be, so this is the just the simple fix.

https://kotlinlang.org/docs/delegated-properties.html

appcues/src/main/java/com/appcues/Storage.kt
```diff
@@ -21,31 +25,85 @@ internal class Storage(
...
-        set(value) = sharedPreferences.edit().putString(Constants.DeviceId.rawVal, value).apply()
+        get() = _deviceId
+        set(value) = updateDeviceId(value)
+
+    private lateinit var _deviceId: String
+
+    private fun updateDeviceId(deviceId: String) {
+        _deviceId = deviceId
+        appcuesCoroutineScope.launch {
+            sharedPreferences.edit().putString(Constants.DeviceId.rawVal, _deviceId).apply()
```
iujames, line 36: from my earlier investigation, the `apply()` didn't actually need a coroutine launch, since its already doing an async save - correct?

appcues/src/main/java/com/appcues/Storage.kt
```diff
@@ -21,31 +25,85 @@ internal class Storage(
...
         }
+    }
 
     init {
-        if (deviceId.isEmpty()) {
-            deviceId = UUID.randomUUID().toString()
+        appcuesCoroutineScope.launch {
+            if (sharedPreferences.getString(Constants.DeviceId.rawVal, String()).isNullOrEmpty()) {
+                _deviceId = UUID.randomUUID().toString()
+            }
```
iujames, line 100: correct me if wrong, but this needs to set `deviceId` instead of `_deviceId` so it actually gets saved?  Otherwise each cold launch would generate a new device ID

iujames (commented):
couple questions, there is a very slim chance of a race condition here causing a `lateinit var` not initialized before accessed it seems, on user ID or device ID.  If something tries to access a value while the SDK is starting up?  Probably not very likely in reality, but I'm thinking of analytics where a session_started event might be firing right as the SDK is starting up and these properties will get accessed.  Any way we can protect for this in usage?

appcues/src/main/java/com/appcues/Storage.kt
andretortolano: its not just that, but this is not actually working, I saw some crashing here. trying to look at it again.

appcues/src/main/java/com/appcues/Storage.kt
```diff
@@ -21,31 +23,79 @@ internal class Storage(
...
+            sharedPreferences.edit().putLong(Constants.LastContentShownAt.rawVal, it.time).apply()
+        } ?: run {
+            sharedPreferences.edit().remove(Constants.LastContentShownAt.rawVal).apply()
         }
+    }
 
     init {
-        if (deviceId.isEmpty()) {
-            deviceId = UUID.randomUUID().toString()
+        val oldPolicy = StrictMode.allowThreadDiskReads()
```
iujames, line 85: if we're going to go the route of allowing for now, what do you think of the approach here to simplify the code a bit and avoid having all the backing fields for properties? https://stackoverflow.com/a/49843629

idea would be to have a helper like 
```
fun <T> allowReads(block: () -> T): T {
    val oldPolicy = StrictMode.allowThreadDiskReads()
    try {
        return block()
    } finally {
        StrictMode.setThreadPolicy(oldPolicy)
    }
}
```

then the old line like 
```
get() = sharedPreferences.getString(Constants.UserId.rawVal, null) ?: ""
```

just becomes
```
get() = allowReads { sharedPreferences.getString(Constants.UserId.rawVal, null) ?: "" }
```

andretortolano: will do

iujames (approved):

PR #203, opened 2022-08-03
Fix analytics reporting during activity pause/resume

iujames:
Issue found in UI testing

If the `AppcuesActivity` goes through pause/resume, it can trigger extraneous `step_seen` analytic events, since the Resume will reset the state machine back to `RenderingStep` and trigger the listener again.

### DIFF
--- a/appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
@@ -84,8 +84,10 @@ internal class StateMachine(
             // update current state
             _state = state
 
-            // emit state change to all listeners via flow
-            _stateFlow.emit(state)
+            if (transition.emitStateChange) {
+                // emit state change to all listeners via flow
+                _stateFlow.emit(state)
+            }
         }
 
         if (sideEffect != null) {
@@ -174,7 +176,7 @@ internal class StateMachine(
 
             // Paused
             state is Paused && action is Resume ->
-                Transition(state.state)
+                Transition(state.state, emitStateChange = false)
 
             state is Paused && action is EndExperience ->
                 Transition(state.state, ContinuationEffect(action))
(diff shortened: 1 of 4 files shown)


appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
```diff
@@ -174,7 +176,7 @@ internal class StateMachine(
 
             // Paused
             state is Paused && action is Resume ->
-                Transition(state.state)
+                Transition(state.state, emitStateChange = false)
```
iujames, line 179: setting this flag `emitStateChange = false` on return from Resume will avoid analytics issues.  I don't see any reason why resets from resume need to be broadcast, as its really just used to reset the internal state.  I didn't go this far (yet) but the `Paused` state transition really seems it doesnt need to be emitted as well.

appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
```diff
@@ -141,6 +141,8 @@ internal class AppcuesViewModel(
     }
 
     fun onPause() {
+        if (uiState.value is Dismissing) return
```
iujames, line 144: another little thing that was hit in UI tests it seemed in some race condition - even though the UI was dismissing, this viewModelScope below still let the Pause action through.  This seems rare, and I don't know how it happened - but we can guard against that pretty easily.

appcues/src/main/java/com/appcues/ui/AppcuesActivity.kt
```diff
@@ -172,6 +173,7 @@ internal class AppcuesActivity : AppCompatActivity() {
                                                 // if we have contentPadding to apply from the WrapContent trait then we apply here
                                                 .then(if (contentPadding != null) Modifier.padding(contentPadding) else Modifier)
                                                 .padding(paddingValues = stepDecoratingPadding.paddingValues.value)
+                                                .testTag("page_$index")
```
iujames, line 176: This is for some upcoming UI test work, so that we can accurately identify elements on a given page inside of a Pager control - came from some good tips I learned here https://akjaw.com/wiriting-ui-tests-for-pager-layouts-jetpack-compose/

andretortolano (approved):
tests are paying off

PR #204, opened 2022-08-03
👌 fine tuning shake motion to make it easier to reproduce

andretortolano:
did a call with ben and we went over the expectations for the shake motion, I did some more work based on that.


iujames (approved):
🕺

PR #207, opened 2022-08-04
Fix image layout handling

iujames:
This should resolve the remaining discrepancies found in UI test cases for various image fit/fill/sizing options.

The core of the issue appears to be the handling of the [Compose `.aspectRatio` Modifier](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/package-summary#(androidx.compose.ui.Modifier).aspectRatio(kotlin.Float,kotlin.Boolean)).  This modifier takes a param for `matchHeightConstraintsFirst` that is fundamentally inflexible compared to how [SwiftUI](https://developer.apple.com/documentation/swiftui/view/aspectratio(_:contentmode:)-771ow) adapts the aspect ratio to do different things based on the content model (FIT vs FILL) - when finding which constraint it will allow to stretch.

After a lot of experimentation, I ended up making my own Modifier by adapting the base Compose aspectRatio to have similar logic to what it appears SwiftUI is doing.  I'll note that detail below, but the rest of that `AppcuesAspectRatioModifier` class is really just carried over from `androidx.compose.foundation.layout.AspectRatioModifier`.

Beyond this, there were some touchups needed
* ensure the aspect ratio modifier is applied AFTER the sizing modifier on the image primitive
* conditionally apply `fillMaxWidth` on images, only when FILL content mode is used
* remove the `fillMaxHeight` used in some places (and height=-1 support) as that does not match the iOS behavior

The end result, our UI test comparisons are now looking good across the board for platform parity
https://user-images.githubusercontent.com/19266448/182942870-f5578c77-c56a-4ed0-9280-e3034dded3b6.mov

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
+++ b/appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
@@ -7,9 +7,6 @@ import androidx.compose.foundation.background
 import androidx.compose.foundation.border
 import androidx.compose.foundation.combinedClickable
 import androidx.compose.foundation.interaction.MutableInteractionSource
-import androidx.compose.foundation.layout.aspectRatio
-import androidx.compose.foundation.layout.fillMaxHeight
-import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
@@ -31,8 +28,10 @@ import com.appcues.action.ExperienceAction
 import com.appcues.data.model.Action
 import com.appcues.data.model.ExperiencePrimitive
 import com.appcues.data.model.ExperiencePrimitive.ImagePrimitive
+import com.appcues.data.model.styling.ComponentContentMode
 import com.appcues.data.model.styling.ComponentSize
 import com.appcues.data.model.styling.ComponentStyle
+import com.appcues.ui.utils.AppcuesAspectRatioModifier
 import com.appcues.ui.utils.margin
 import com.appcues.util.eq
 import com.appcues.util.ne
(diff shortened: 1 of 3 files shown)


appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
```diff
@@ -46,29 +45,26 @@ internal fun Modifier.outerPrimitiveStyle(
...
 )
 
-private fun Modifier.stylePrimitive(experiencePrimitive: ExperiencePrimitive) = this.then(
-    when (experiencePrimitive) {
-        is ImagePrimitive ->
-            Modifier
-                .animateContentSize()
-                .imageAspectRatio(experiencePrimitive.intrinsicSize)
-        else -> Modifier
-    }
```
iujames, line 71: this was only used for the ImagePrimitive and only for the aspectRatio stuff, so moved to new function and changed the callsite.

appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
```diff
@@ -130,31 +126,31 @@ internal fun Modifier.styleBorder(
     }
 )
 
-internal fun Modifier.styleSize(style: ComponentStyle, noSizeFillMax: Boolean = false) = this.then(
+internal fun Modifier.styleSize(style: ComponentStyle, contentMode: ComponentContentMode = ComponentContentMode.FIT) = this.then(
```
iujames, line 129: most things _dont_ stretch to fill size - so default here is FIT, but images can optionally pass in FILL.

appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
```diff
@@ -130,31 +126,31 @@ internal fun Modifier.styleBorder(
...
+        style.height != null -> Modifier.height(style.height.dp).styleDefaultWidth(contentMode)
+        // at the end we fill max width if there is no width/height but the primitive (like image) is set to FILL
+        contentMode == ComponentContentMode.FILL -> Modifier.fillMaxWidth()
+        else -> Modifier
+    }
+)
+
+private fun Modifier.styleDefaultWidth(contentMode: ComponentContentMode) = this.then(
+    when (contentMode) {
+        ComponentContentMode.FILL -> Modifier.fillMaxWidth()
```
iujames, line 153: this adds our full width for the case of only height specified, but a FILL image

appcues/src/main/java/com/appcues/ui/utils/AppcuesAspectRatioModifier.kt
```diff
@@ -0,0 +1,176 @@
...
+
+    // Suppressing these, as this method is taken from the Compose source and adapted slightly to add the
+    // logic based on contentMode.  Attempting to rewrite this differently would be a more confusing exercise at this point.
+    @Suppress("ComplexMethod", "ReturnCount")
+    private fun Constraints.findSize(): IntSize {
+
+        val containerAspectRatio = this.maxWidth.toDouble() / this.maxHeight.toDouble()
+        val matchHeightConstraintsFirst =
+            (contentMode == ComponentContentMode.FIT && containerAspectRatio > aspectRatio) ||
+                (contentMode == ComponentContentMode.FILL && containerAspectRatio < aspectRatio)
```
iujames, line 89: this is really the only thing changed in this class - adds the desired conditional logic to look at the contentModel FIT vs FILL, then look at the aspect ratio desired and compare to the aspect ratio of the container -- it can then decide which constraint set to prefer to give the expected fit vs fill behavior.

mmaatttt:
Just stopping by to say this is incredibly impressive!! 💯

andretortolano (approved):
NICE

appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
andretortolano: yea I did that locally as well. thanks!

PR #208, opened 2022-08-04
♻️  Removing Shot from SDK

andretortolano:
Removing shot from SDK since now we do have screenshot testing on mobile-specs


appcues/src/test/java/com/appcues/action/appcues/TrackEventActionTest.kt
```diff
@@ -19,7 +19,7 @@ internal class TrackEventActionTest : AppcuesScopeTest {
     override val koinTestRule = KoinScopeRule()
 
     @Test
-    fun `tracke event SHOULD have expected type name`() {
+    fun `track event SHOULD have expected type name`() {
```
andretortolano, line 22: found a typo

appcues/build.gradle
```diff
@@ -110,13 +104,9 @@ dependencies {
     implementation "com.google.accompanist:accompanist-navigation-animation:$accompanist_version"
 
     testImplementation "junit:junit:4.13.2"
-    testImplementation "com.squareup.okhttp3:mockwebserver:4.9.3"
     testImplementation "io.mockk:mockk:1.12.2"
     testImplementation "com.google.truth:truth:1.1.3"
+    testImplementation "com.squareup.okhttp3:mockwebserver:4.10.0"
```
iujames, line 109: do we use this at all in this repo? ah right the AppcuesServiceTest.  That is _kinda_ duplicating what the UI experience tests are doing, so perhaps we could remove that and the `content_modal_one.json` file and contentModalOneStubs code in this repo to clean up? No strong opinion on that one.

iujames (approved):

appcues/build.gradle
andretortolano: we can come back to this later, but its an option. I'm not if these tests are helping us in any way, but I must say that I rather have tests in the same repo/module than outside of it. you can easily make changes and run tests  (Unit tests) to ensure everything is still work and all)

iujames: that does actually work as well in the spec repo project! since `appcues` is included as a module via relative path reference

PR #211, opened 2022-08-08
✨ Added support for background-traits

andretortolano:
in order to support background trait some changes had to be made. (I will note it on the code)

### DIFF
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -19,7 +19,7 @@ import com.appcues.debugger.AppcuesDebuggerManager
 import com.appcues.di.AppcuesKoinContext
 import com.appcues.logging.Logcues
 import com.appcues.trait.ExperienceTrait
-import com.appcues.trait.ExperienceTrait.ExperienceTraitLevel
+import com.appcues.trait.ExperienceTraitLevel
 import com.appcues.trait.TraitRegistry
 import com.appcues.ui.ExperienceRenderer
 import kotlinx.coroutines.launch
@@ -183,7 +183,9 @@ class Appcues internal constructor(koinScope: Scope) {
      * @param type Type of the action that is sent by the experience. ex: "my-trait"
      * @param traitFactory Factory (lambda) responsible for creating the ExperienceTrait registered for given [type]
      *                     config is an optional map that will be mapped and used to invoke the custom trait with
-     *                     level defines where that trait was found in the json nodes, varying from EXPERIENCE, GROUP, STEP
+     *                     level informs the trait which level in the experience rendering hierarchy it is being
+     *                     applied - EXPERIENCE, GROUP, or STEP. This can optionally be used by a trait to apply
+     *                     different rendering logic when applied to a group versus a single step, for example.
      * usage:
      * registerTrait("my-trait") { MyCustomExperienceTrait() }
      */
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -19,6 +19,7 @@ import com.appcues.debugger.AppcuesDebuggerManager
 import com.appcues.di.AppcuesKoinContext
 import com.appcues.logging.Logcues
 import com.appcues.trait.ExperienceTrait
+import com.appcues.trait.ExperienceTrait.ExperienceTraitLevel
 import com.appcues.trait.TraitRegistry
 import com.appcues.ui.ExperienceRenderer
 import kotlinx.coroutines.launch
@@ -37,10 +38,10 @@ fun Appcues(
     accountId: String,
     applicationId: String,
     config: (AppcuesConfig.() -> Unit)? = null,
-): Appcues =
+): Appcues = AppcuesKoinContext
     // This creates the Koin Scope and initializes the Appcues instance within, then returns the Appcues instance
     // ready to go with the necessary dependency configuration in its scope.
-    AppcuesKoinContext.createAppcuesScope(context, AppcuesConfig(accountId, applicationId).apply { config?.invoke(this) }).get()
+    .createAppcuesScope(context, AppcuesConfig(accountId, applicationId).apply { config?.invoke(this) }).get()
 
 /**
  * The main entry point for using Appcues functionality in your application - tracking
@@ -181,11 +182,12 @@ class Appcues internal constructor(koinScope: Scope) {
      *
      * @param type Type of the action that is sent by the experience. ex: "my-trait"
      * @param traitFactory Factory (lambda) responsible for creating the ExperienceTrait registered for given [type]
-     *
+     *                     config is an optional map that will be mapped and used to invoke the custom trait with
+     *                     level defines where that trait was found in the json nodes, varying from EXPERIENCE, GROUP, STEP
      * usage:
      * registerTrait("my-trait") { MyCustomExperienceTrait() }
      */
-    fun registerTrait(type: String, traitFactory: (config: Map<String, Any>?) -> ExperienceTrait) {
+    fun registerTrait(type: String, traitFactory: (config: Map<String, Any>?, level: ExperienceTraitLevel) -> ExperienceTrait) {
         traitRegistry.register(type, traitFactory)
     }
 
--- a/appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
+++ b/appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
@@ -9,7 +9,10 @@ internal fun List<ExperienceTrait>.mergeTraits(others: List<ExperienceTrait>, le
     return mutableListOf<ExperienceTrait>().apply {
         addAll(this@mergeTraits)
 
-        others.filter { other -> other.level > level }
+        others
+            // filters out traits that should not be propagated to children nodes
+            .filter { other -> other.level >= level }
+            // then we merge based on its type (prioritize whats on the child node)
             .forEach { other -> if (none { it.type == other.type }) add(other) }
     }
 }
--- a/appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
+++ b/appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
@@ -1,19 +1,21 @@
 package com.appcues.data.mapper
 
 import com.appcues.data.remote.response.action.ActionResponse
-import com.appcues.trait.ExperienceTrait
+import com.appcues.data.remote.response.trait.TraitResponse
 import com.appcues.trait.ExperienceTrait.ExperienceTraitLevel
 import java.util.UUID
 
-internal fun List<ExperienceTrait>.mergeTraits(others: List<ExperienceTrait>, level: ExperienceTraitLevel): List<ExperienceTrait> {
-    return mutableListOf<ExperienceTrait>().apply {
+internal typealias LeveledTraitResponse = Pair<TraitResponse, ExperienceTraitLevel>
+
+internal fun List<LeveledTraitResponse>.mergeTraits(other: List<LeveledTraitResponse>): List<LeveledTraitResponse> {
+    return mutableListOf<LeveledTraitResponse>().apply {
         addAll(this@mergeTraits)
 
-        others
-            // filters out traits that should not be propagated to children nodes
-            .filter { other -> other.level >= level }
-            // then we merge based on its type (prioritize whats on the child node)
-            .forEach { other -> if (none { it.type == other.type }) add(other) }
+        other.forEach { trait ->
+            if (none { it.first.type == trait.first.type }) {
+                add(trait)
+            }
+        }
     }
 }
 
--- a/appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
+++ b/appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
@@ -1,18 +1,16 @@
 package com.appcues.data.mapper
 
 import com.appcues.data.remote.response.action.ActionResponse
-import com.appcues.data.remote.response.trait.TraitResponse
+import com.appcues.trait.ExperienceTrait
+import com.appcues.trait.ExperienceTrait.ExperienceTraitLevel
 import java.util.UUID
 
-internal fun List<TraitResponse>.mergeTraits(other: List<TraitResponse>): List<TraitResponse> {
-    return mutableListOf<TraitResponse>().apply {
+internal fun List<ExperienceTrait>.mergeTraits(others: List<ExperienceTrait>, level: ExperienceTraitLevel): List<ExperienceTrait> {
+    return mutableListOf<ExperienceTrait>().apply {
         addAll(this@mergeTraits)
 
-        other.forEach { trait ->
-            if (none { it.type == trait.type }) {
-                add(trait)
-            }
-        }
+        others.filter { other -> other.level > level }
+            .forEach { other -> if (none { it.type == other.type }) add(other) }
     }
 }
 
--- a/appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
+++ b/appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
@@ -2,7 +2,7 @@ package com.appcues.data.mapper
 
 import com.appcues.data.remote.response.action.ActionResponse
 import com.appcues.data.remote.response.trait.TraitResponse
-import com.appcues.trait.ExperienceTrait.ExperienceTraitLevel
+import com.appcues.trait.ExperienceTraitLevel
 import java.util.UUID
 
 internal typealias LeveledTraitResponse = Pair<TraitResponse, ExperienceTraitLevel>
--- a/appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
+++ b/appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
@@ -20,7 +20,7 @@ import com.appcues.trait.ContainerDecoratingTrait
 import com.appcues.trait.ContentHolderTrait
 import com.appcues.trait.ContentWrappingTrait
 import com.appcues.trait.ExperienceTrait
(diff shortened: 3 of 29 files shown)


appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
```diff
@@ -1,18 +1,16 @@
...
 
 import com.appcues.data.remote.response.action.ActionResponse
-import com.appcues.data.remote.response.trait.TraitResponse
+import com.appcues.trait.ExperienceTrait
+import com.appcues.trait.ExperienceTrait.ExperienceTraitLevel
 import java.util.UUID
 
-internal fun List<TraitResponse>.mergeTraits(other: List<TraitResponse>): List<TraitResponse> {
-    return mutableListOf<TraitResponse>().apply {
+internal fun List<ExperienceTrait>.mergeTraits(others: List<ExperienceTrait>, level: ExperienceTraitLevel): List<ExperienceTrait> {
```
andretortolano, line 8: instead of dealing with TraitResponse, we are now using ExperienceTrait, so we can check its "level" to know if we should propagate that trait to groups, steps

appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
```diff
@@ -51,23 +51,23 @@ internal class ExperienceMapper(
...
 
     private fun List<StepContainerResponse>.mapToStepContainer(
-        superTraits: List<TraitResponse>,
+        superTraits: List<ExperienceTrait>,
         superActions: Map<UUID, List<ActionResponse>>?
     ) = map { stepContainerResponse ->
-        val mergedTraits = stepContainerResponse.traits.mergeTraits(superTraits)
-        val mappedTraits = traitsMapper.map(mergedTraits)
+        val mappedTraits = traitsMapper.map(stepContainerResponse.traits)
+        val mergedTraits = mappedTraits.mergeTraits(superTraits, EXPERIENCE)
```
andretortolano, line 58: here we are merging traits in this node with traits  where level > EXPERIENCE, meaning GROUP, and STEP 

**looking back now I it would be nice to change so its level >= GROUP

appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
```diff
@@ -1,18 +1,16 @@
...
+    return mutableListOf<ExperienceTrait>().apply {
         addAll(this@mergeTraits)
 
-        other.forEach { trait ->
-            if (none { it.type == trait.type }) {
-                add(trait)
-            }
-        }
+        others.filter { other -> other.level > level }
```
iujames, line 12: for my own education - what is this `other.level > level` check adding now? a comment in here about how this merge works would be nice for future reference.

appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
iujames, line 58: goes with my previous question i guess, but trying to understand why `EXPERIENCE` is used here when the traits are at the `GROUP` level for the container here.

appcues/src/main/java/com/appcues/trait/ExperienceTrait.kt
```diff
@@ -2,5 +2,32 @@ package com.appcues.trait
...
     val config: Map<String, Any>?
+
+    /**
+     * [level] defines if this trait should be propagated to a certain point from where it was found in the data modal.
+     *
+     * eg: In case we find an ExperienceTrait with level STEP while reading the experience model,
+     *     we will merge this trait to each group and each step
+     *
+     * eg: In case we find an ExperienceTrait with level GROUP while reading a parent step (group),
+     *     we will leave as is, the only structure that will know this trait will be a stepGroup
```
iujames, line 29: great to have docs added here!

This doesn't exactly match my understanding of how this `level` was designed initially - and maybe I'm wrong and you and @mmaatttt already worked this out.  I thought the `level` was simply to give a hint to the rendering system which container to render the content in (group vs step), rather than implying an other sort of hierarchical merge/propagation logic?

appcues/src/main/java/com/appcues/trait/appcues/BackdropTrait.kt
```diff
@@ -27,6 +28,10 @@ internal class BackdropTrait(
         const val TYPE = "@appcues/backdrop"
     }
 
+    override val type = TYPE
+
+    override val level = ExperienceTraitLevel.GROUP
```
iujames, line 33: And here is where I thought that the TraitMapper would pass in the level that the trait was being applied to, at construction time, rather than hardcoding any particular trait to any particular level.   i.e. perhaps some future use case would allow a backdrop at the Experience level here (no idea what that use case is, just an example) - it would all be seamlessly passed through during the experience mapping time.  

Most Traits would not need it at all, seemingly - only the new background-content.  Maybe it is not even part of the public `ExperienceTrait` interface - just a constructor arg for the one that needs it.

appcues/src/main/java/com/appcues/trait/appcues/DefaultPresentingTrait.kt
```diff
@@ -12,6 +13,15 @@ internal class DefaultPresentingTrait(
...
 ) : PresentingTrait {
 
+    companion object {
+
+        const val TYPE = "@appcues/default-presenting"
+    }
+
+    override val type = TYPE
+
+    override val level = ExperienceTraitLevel.EXPERIENCE
```
iujames, line 23: this default presenting would be at the container (GROUP) level, no?

appcues/src/main/java/com/appcues/ui/AppcuesActivity.kt
```diff
@@ -142,14 +147,14 @@ internal class AppcuesActivity : AppCompatActivity() {
...
+            backdropDecoratingTraits.forEach {
                 with(it) { Backdrop() }
             }
             // create wrapper
             contentWrappingTrait.WrapContent { hasFixedHeight, contentPadding ->
-                Box(
-                    contentAlignment = Alignment.TopCenter
-                ) {
+                Box(contentAlignment = Alignment.TopCenter) {
+                    ApplyUnderlayDecoratingTraits(this)
```
andretortolano, line 156: key change here, now we know if a decorating trait is underlay or overlay,  then we filter and apply those in the correct place. (both for step and container)

appcues/src/main/java/com/appcues/ui/AppcuesActivity.kt
```diff
@@ -162,24 +167,12 @@ internal class AppcuesActivity : AppCompatActivity() {
...
-                                        }
-
-                                        // apply step decorating traits
-                                        traits.forEach { with(it) { Overlay(stepDecoratingPadding) } }
+
+                                        ApplyUnderlayDecoratingTraits(this@Box, stepDecoratingPadding)
+
+                                        ComposeStepContent(index, hasFixedHeight, contentPadding, stepDecoratingPadding)
+
+                                        ApplyOverlayDecoratingTraits(this@Box, stepDecoratingPadding)
```
iujames, line 175: this feels very clean and clear on how the composition is layered now 👍

appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
```diff
@@ -45,28 +44,20 @@ internal fun Modifier.outerPrimitiveStyle(
     gestureProperties: PrimitiveGestureProperties,
     isDark: Boolean,
     defaultBackgroundColor: Color? = null,
+    matchParentBox: BoxScope? = null,
```
andretortolano, line 47: had to add this as a way to inform that we have a parent (box) we would want to matchParentSize to.

appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
```diff
@@ -126,8 +117,13 @@ internal fun Modifier.styleBorder(
...
 )
 
-internal fun Modifier.styleSize(style: ComponentStyle, contentMode: ComponentContentMode = ComponentContentMode.FIT) = this.then(
+internal fun Modifier.styleSize(
+    style: ComponentStyle,
+    matchParentBox: BoxScope? = null,
+    contentMode: ComponentContentMode = ComponentContentMode.FIT,
+) = this.then(
     when {
+        matchParentBox != null -> with(matchParentBox) { Modifier.matchParentSize() }
```
andretortolano, line 126: matchParentBox takes precedent over anything else when its not null.

so if for some reason we would like to build a background where the image is not fullscreen, we would need to put that into a stack and then size it (width/height).

this is in parity with IOS afaik

appcues/src/main/java/com/appcues/ui/primitive/ExperiencePrimitive.kt
```diff
@@ -60,7 +61,7 @@ internal fun ExperiencePrimitive.Compose() {
                 is ButtonPrimitive -> Compose(modifier)
                 is EmbedHtmlPrimitive -> Compose(modifier)
                 is HorizontalStackPrimitive -> Compose(modifier)
-                is ImagePrimitive -> Compose(modifier)
+                is ImagePrimitive -> Compose(modifier, matchParentBox)
```
andretortolano, line 64: since we did this with Image in mind I think its ok to apply this only to image, but maybe we would want to replicate this to other types as well

appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
iujames: sorry I see you commented on this above, ok - still would be nice to have more explanation in the code I think.

appcues/src/main/java/com/appcues/ui/primitive/ImagePrimitive.kt
```diff
@@ -9,18 +11,18 @@ import com.appcues.ui.LocalImageLoader
...
 
 @Composable
-internal fun ImagePrimitive.Compose(modifier: Modifier) {
+internal fun ImagePrimitive.Compose(modifier: Modifier, matchParentBox: BoxScope? = null) {
     val context = LocalContext.current
     val decodedBlurHash = rememberBlurHashDecoded(blurHash = blurHash)
 
     AsyncImage(
-        modifier = modifier.then(Modifier.styleSize(style, contentMode).styleImageAspect(this)),
+        modifier = modifier.applyImageModifier(this, matchParentBox),
```
andretortolano, line 25: image is the only primitive that uses styleSize twice, this is why we are passing the BoxScope as parameter

appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
andretortolano: whatever you see its confusing, make note so I dont forget to add extra explanation in the code.

appcues/src/main/java/com/appcues/trait/ExperienceTrait.kt
andretortolano: @mmaatttt can jump in to see if that makes sense.

appcues/src/main/java/com/appcues/trait/appcues/BackdropTrait.kt
andretortolano: the problem is that I wanted to check for something relatable to ExperienceTrait, so we can use that in a for loop in MapperExt

andretortolano: one thing we could do is convert some interfaces to abstract class, then we can set this one just once in the super class

appcues/src/main/java/com/appcues/trait/appcues/BackgroundContentTrait.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    @OptIn(ExperimentalComposeUiApi::class)
+    @Composable
+    private fun BoxScope.Decorate() {
+        Box(
+            modifier = Modifier
+                .matchParentSize()
+                // merge all children nodes to this one, and mark as invisibleToUser so talkback wont read.
+                // this is the same as old xml: importantForAccessibility = false
+                .semantics(true) { invisibleToUser() }
```
andretortolano, line 59: accessibility guard to ignore everything from this Decorate

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -38,7 +39,7 @@ fun Appcues(
     applicationId: String,
     config: (AppcuesConfig.() -> Unit)? = null,
 ): Appcues =
-    // This creates the Koin Scope and initializes the Appcues instance within, then returns the Appcues instance
+// This creates the Koin Scope and initializes the Appcues instance within, then returns the Appcues instance
```
andretortolano, line 42: auto-formatter.

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -181,11 +182,12 @@ class Appcues internal constructor(koinScope: Scope) {
      *
      * @param type Type of the action that is sent by the experience. ex: "my-trait"
      * @param traitFactory Factory (lambda) responsible for creating the ExperienceTrait registered for given [type]
-     *
+     *                     config is an optional map that will be mapped and used to invoke the custom trait with
```
andretortolano, line 185: I'm not sure how to best explain both properties for the factory, open for suggestions.

appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
```diff
@@ -2,14 +2,17 @@ package com.appcues.data.mapper
 
 import com.appcues.data.remote.response.action.ActionResponse
 import com.appcues.data.remote.response.trait.TraitResponse
+import com.appcues.trait.ExperienceTrait.ExperienceTraitLevel
 import java.util.UUID
 
-internal fun List<TraitResponse>.mergeTraits(other: List<TraitResponse>): List<TraitResponse> {
-    return mutableListOf<TraitResponse>().apply {
+internal typealias LeveledTraitResponse = Pair<TraitResponse, ExperienceTraitLevel>
```
andretortolano, line 8: this was created soh at  merging time (TraitResponse) I already know its level, so that we can use it later when its all merged.

this was not necessary when @iujames did becasue of one change. the mergeTraits was merging ExperienceTraits instead of TraitResponse, and that was working because we had  each ExperienceTrait define a variable called type. (which I removed again)

appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
```diff
@@ -35,10 +37,11 @@ internal class ExperienceMapper(
 ) {
 
     fun map(from: ExperienceResponse, priority: ExperiencePriority = NORMAL): Experience {
+        val experienceTraits = from.traits.map { it to EXPERIENCE }
```
andretortolano, line 40: map to Pair<TraitResponse, ExperienceTraitLevel> before passing it down

appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
```diff
@@ -51,12 +54,13 @@ internal class ExperienceMapper(
     }
 
     private fun List<StepContainerResponse>.mapToStepContainer(
-        superTraits: List<TraitResponse>,
-        superActions: Map<UUID, List<ActionResponse>>?
+        experienceTraits: List<LeveledTraitResponse>,
+        experienceActions: Map<UUID, List<ActionResponse>>?
     ) = map { stepContainerResponse ->
-        val mergedTraits = stepContainerResponse.traits.mergeTraits(superTraits)
+        val containerTraits = stepContainerResponse.traits.map { it to GROUP }
```
andretortolano, line 60: same thing here, but for group traits.

appcues/src/main/java/com/appcues/data/mapper/step/StepMapper.kt
```diff
@@ -19,16 +20,17 @@ internal class StepMapper(
...
     fun map(
         from: StepResponse,
-        superTraits: List<TraitResponse>,
-        superActions: Map<UUID, List<ActionResponse>>,
+        stepContainerTraits: List<LeveledTraitResponse>,
+        stepContainerActions: Map<UUID, List<ActionResponse>>,
     ): Step {
-        val mergedTraits = from.traits.mergeTraits(superTraits)
-        val mergedActions = from.actions.mergeActions(superActions)
+        val stepTraits = from.traits.map { it to STEP }
```
andretortolano, line 26: and same here for Step level traits, then later we map and the mapper knows each trait level without extra parameter.

appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
iujames: seems fine - typically would lean toward just having a simple data class to hold the items (plus allow for future extensibility) over using Pair - but also no issue with using Pair until if or when we actually need more than two things, maybe never.

appcues/src/main/java/com/appcues/trait/ExperienceTrait.kt
```diff
@@ -2,5 +2,13 @@ package com.appcues.trait
 
 interface ExperienceTrait {
 
+    enum class ExperienceTraitLevel {
+        EXPERIENCE, GROUP, STEP
+    }
```
iujames, line 7: maybe un-nest this? or make it `ExperienceTrait.Level` ?  Just thinking over a bit more since its part of the public API now.

iujames (approved):
looks good to me! couple little thoughts, but this is going to be a really nice enhancement to allow the background content!

appcues/src/main/java/com/appcues/data/mapper/MapperExt.kt
andretortolano: thats what came to mind to tie TraitResponse to a Level.

appcues/src/main/java/com/appcues/trait/ExperienceTrait.kt
andretortolano: I will un-nest it

PR #215, opened 2022-08-10
Update wording in Debugger UI for deep link check

iujames:
also fix padding on end of row

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/DebuggerStatusManager.kt
+++ b/appcues/src/main/java/com/appcues/debugger/DebuggerStatusManager.kt
@@ -96,10 +96,11 @@ internal class DebuggerStatusManager(
         updateData()
     }
 
-    fun checkDeepLinkValidation(deepLinkPath: String) {
+    suspend fun checkDeepLinkValidation(deepLinkPath: String) {
         if (deepLinkPath == deepLinkValidationToken) {
             deepLinkConfigured = true
             deepLinkValidationToken = null
+            updateData()
         }
     }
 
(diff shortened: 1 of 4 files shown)


appcues/src/main/java/com/appcues/debugger/DebuggerStatusManager.kt
```diff
@@ -96,10 +96,11 @@ internal class DebuggerStatusManager(
         updateData()
     }
 
-    fun checkDeepLinkValidation(deepLinkPath: String) {
+    suspend fun checkDeepLinkValidation(deepLinkPath: String) {
         if (deepLinkPath == deepLinkValidationToken) {
             deepLinkConfigured = true
             deepLinkValidationToken = null
+            updateData()
```
iujames, line 103: noticed that without this update _sometimes_ the state of the UI would not update to reflect the change appropriately.  I'm not sure how this was missed or working last time...

appcues/src/main/res/values/strings.xml
```diff
@@ -14,14 +14,15 @@
     <string name="appcues_debugger_status_reload_icon_content_description">reload icon</string>
     <string name="appcues_debugger_status_device_title">"%1$s Android %2$s</string>
     <string name="appcues_debugger_status_experience_line1">"Showing group %1$s step %2$s"</string>
-    <string name="appcues_debugger_status_sdk_title">Installed SDK (%1$s)</string>
+    <string name="appcues_debugger_status_sdk_title">Installed SDK %1$s</string>
```
iujames, line 17: removed parens just to match iOS side

appcues/src/main/java/com/appcues/debugger/DebuggerStatusManager.kt
andretortolano: its probably because one of the other items was triggering the update.

andretortolano (approved):

PR #216, opened 2022-08-10
Simplify access to SDK version

iujames:
makes this available on the companion for access from `Appcues.version` directly - don't need an instance. Matches iOS and simplifies things like using in Democues where the instance is not created initially until the account/app input is set.

and remove the build type


andretortolano (approved):
should we remove appcues.verison?

PR #217, opened 2022-08-10
Don't include - in version name if no alpha/beta classifier

iujames:


appcues/publish-maven.gradle
```diff
@@ -19,7 +19,7 @@ private String generateVersionName() {
     String classifier = appcuesProperties.getProperty("VERSION_CLASSIFIER")
     String versionName = "${major}.${minor}.${patch}"
 
-    if (classifier != null) {
+    if (classifier?.trim()) {
```
iujames, line 22: this checks null or empty - we had empty before leading to a `1.0.0-` version name being used in the maven publish

andretortolano (approved):

PR #219, opened 2022-08-24
Add error logging when Coil fails to load images

iujames:
Minor improvement in logging - noticed while working on dependencies in Xamarin binding that Coil can silently fail for various reasons and our logging did not give any indication as to why.

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/AppcuesCompositions.kt
+++ b/appcues/src/main/java/com/appcues/ui/AppcuesCompositions.kt
@@ -35,7 +35,11 @@ internal data class AppcuesPagination(val onPageChanged: (Int) -> Unit)
 
 internal val appcuesPaginationData = mutableStateOf(AppcuesPaginationData(1, 0, 0.0f))
 
-internal val LocalLogcues = staticCompositionLocalOf<Logcues?> { null }
+internal val LocalLogcues = staticCompositionLocalOf<Logcues> { noLocalProvidedFor("LocalLogcues") }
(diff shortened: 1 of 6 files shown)


appcues/src/main/java/com/appcues/ui/AppcuesCompositions.kt
```diff
@@ -34,6 +35,8 @@ internal data class AppcuesPagination(val onPageChanged: (Int) -> Unit)
 
 internal val appcuesPaginationData = mutableStateOf(AppcuesPaginationData(1, 0, 0.0f))
 
+internal val LocalLogcues = staticCompositionLocalOf<Logcues?> { null }
```
andretortolano, line 38: could we maybe change this to be similar to LocalContext. without using "?"

it seems like it does something like this:

```kotlin
val LocalContext = staticCompositionLocalOf<Context> {
    noLocalProvidedFor("LocalContext")
}

private fun noLocalProvidedFor(name: String): Nothing {
    error("CompositionLocal $name not present")
}
```

so basically we have to provide it or we get runtime exception, but since we will always provide it we basically avoid have to deal with ?.let scenarios

andretortolano (approved):

appcues/src/main/java/com/appcues/ui/AppcuesCompositions.kt
iujames: yes! i was looking for a good way to do that thank you

PR #222, opened 2022-08-25
Form input controls - part 1 - data models

iujames:
This ended up being a larger refactor than I anticipated.

The primary purpose of this change was to add the JSON models and domain model mapping for the new `textInput` and `optionSelect` primitives [defined in the spec](https://github.com/appcues/appcues-mobile-experience-spec/blob/main/experience-schema.json).  I discovered that this is the first case where two primitives have the same property (`defaultValue` in this case), which had different data types (String vs List<String>).

This led me to pursue a path or refactoring where we split out what used to all be merged in `StepContentResponse` into individual JSON model classes for each primitive.  This gives a couple advantages: (1) we can easily handle cases where the same prop name might have different types and (2) we can put more structured handling around required/optional fields for each type, and let the JSON deserialization layer in Moshi handle some of that null checking instead of having to put it in our mapper layer.

Moshi provides a `PolymorphicJsonAdapterFactory` for just this situation - deserializing a model where there are a bunch of different "flavors" of a type defined by some key - in our case its the `type` key, like `"type": "stack"` in our primitive models.  [This guide](https://proandroiddev.com/moshi-polymorphic-adapter-is-d25deebbd7c5) is a good overview and explains basically exactly how it's implemented here.

Then, the core work of actually adding the new `TextInputPrimitiveResponse` and `OptionSelectPrimitiveResponse` models, and mapping them to `TextInputPrimitive` and `OptionSelectPrimitive` is pretty straightforward.

Finally, I tried to take a pass at simplifying some of the layers of the mapping stuff to remove the lower level `*Mapper` classes and just use extensions that mapped types directly. I'm not feeling too strongly about this, but just felt like it made that code leaner and easier to add things to, personally.  The usage change is pretty basic, moving from

```
val mapper = StyleMapper()
val style = mapper.map(styleResponse)
```
to
```
val style = styleResponse.mapComponentStyle()
```

The `StepContentMapper` is removed, and instead you can just call `mapPrimitive()` on any `PrimitiveResponse` to invoke a chain of recursive mapper extensions through the content tree.

next step is to start building some UI for these new primitives.


appcues/src/main/java/com/appcues/ui/primitive/ExperiencePrimitive.kt
```diff
@@ -65,12 +67,18 @@ internal fun ExperiencePrimitive.Compose(matchParentBox: BoxScope? = null) {
...
                 is VerticalStackPrimitive -> Compose(modifier)
+                is TextInputPrimitive -> throw NotImplementedError()
+                is OptionSelectPrimitive -> throw NotImplementedError()
             }
         }
     }
 }
 
 private fun ExperiencePrimitive.getRole(): Role {
+    // This is for any tap actions in the experience which currently only
```
andretortolano, line 78: yes, sorry I had no comments here you probably had to do some searching to figure it out whats that about.

andretortolano (approved):
Nice refactory change to mapping layer

PR #223, opened 2022-08-26
🐛 Fix debugger position overlayed on Experiences

andretortolano:
While preparing Ionic samples I noticed that the debugger was not properly being displayed on top of experiences. this is because of a change for a previous bug we worked on, so instead of using fitSystemWindow=true that forces the activity content (decorView) to apply padding we are applying the padding ourselves in root composition, this way we dont mess up with debugger standard behavior to apply padding.

PS: accidentally pushed UI tests updates to main, but here is the commit: https://github.com/appcues/appcues-mobile-experience-spec/commit/c292d23588bbea15fab925102b7b68d5ea98756a


iujames (approved):
Do we have any sort of UI test we could use to simulate the Delivery Hero issue and confirm we don't cause regressions here in the future? Just thinking that would be nice to have, maybe something we could add in that spec repo host app / test runner?

PR #224, opened 2022-08-30
Implement initial UI controls for textInput and optionSelect primitives

iujames:
Spent some time looking at UI options and implemented a basic first version that just uses standard controls for now.  We don't have final requirements for how this is actually supposed to look and the related stying options, so I'm just going to checkpoint this work here for now and go look at the actual persisting/reporting of user input state portion.  We can come back to UI styling when the requirements are known.

| checkbox - vertical | checkbox - horizontal | radio | picker |
| ---- | ----- | ----- | ----|
| checkbox_vert | checkbox_horiz | radio_vert | picker |

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/primitive/OptionSelectPrimitive.kt
+++ b/appcues/src/main/java/com/appcues/ui/primitive/OptionSelectPrimitive.kt
@@ -1,0 +1,139 @@
+package com.appcues.ui.primitive
+
+import androidx.compose.foundation.Image
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.material.DropdownMenu
+import androidx.compose.material.DropdownMenuItem
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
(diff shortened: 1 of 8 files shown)


appcues/src/main/java/com/appcues/ui/primitive/OptionSelectPrimitive.kt
```diff
@@ -0,0 +1,139 @@
...
+                    ) {
+                        options.forEachIndexed { index, optionItem ->
+                            DropdownMenuItem(onClick = {
+                                selectedIndex = index
+                                expanded = false
+                            }) {
+                                optionItem.content.Compose()
+                            }
+                        }
+                    }
```
iujames, line 66: using a `DropdownMenu` as an initial, very basic way to do a Picker.

For reference, a default SwiftUI Picker on iOS looks like:
ios_picker

andretortolano: thats how I imagine we would do on Android so you got that right

andretortolano (approved):
fine checkpoint, I'm not sure how we will style the Text input field "accent" color, right now its probably getting info from MaterialTheme so we could change that to a more neutral color but maybe its something we could have an option to customize from builder.

iujames:
> I'm not sure how we will style the Text input field "accent" color, right now its probably getting info from MaterialTheme so we could change that to a more neutral color but maybe its something we could have an option to customize from builder.

Yep, material theme I'm pretty sure.  There is a good level of control over that using the `colors` property and the `TextFieldDefaults.textFieldColors`, when the time comes.
```
        TextField(
            modifier = ...,
            value = ...,
            colors = TextFieldDefaults.textFieldColors(
                backgroundColor = ...,
                cursorColor = ...,
                disabledLabelColor = ...,
                focusedIndicatorColor = ...,
                unfocusedIndicatorColor = ...
            ),
```

PR #225, opened 2022-08-30
Update Detekt 1.19.0 -> 1.21.0

iujames:
This should also resolve the security check noted here https://github.com/appcues/appcues-android-sdk/runs/8096333392

updating org.yaml:snakeyaml:1.31 (from 1.29)

### DIFF
--- a/.idea/detekt.xml
+++ b/.idea/detekt.xml
@@ -1,5 +1,14 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="DetektPluginSettings">
+    <option name="configurationFilePaths">
+      <set>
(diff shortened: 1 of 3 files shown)


iujames:
argh.  my bad, I thought this would fix the issue but that update is only in the unreleased branch of Detekt so far, not in 1.21 which had the update to snakeyaml:1.30 (not 1.31) https://github.com/detekt/detekt/commit/2866976a40c7802a2ae4670703e8373e66603dee

this change is still prob fine though I think, a step forward and some detekt updates.

iujames:
@andretortolano any objections with moving ahead to latest Detekt here?

andretortolano (approved):
I missed that one, its all cool

PR #229, opened 2022-09-08
Add support for maintaining form state at Step level, initial analytics tracking

iujames:
Adds our form state object `ExperienceStepFormState` to the domain model for `Step`, and then use this value in the CompositionLocal that gets sent into the compose primitives.  This way, we can retain the form state in the domain model as the user moves through steps, and restore previous values if they return to a step.  It also has the data prepared in a way that makes it easy to submit analytics from the Experience and Step model later on.

This change also takes a first pass at formatting the `appcues:v2:step_interaction` event with the `interactionType` and `interactionData` from the initial proposal. This form is still TBD, but this gets some initial structure in place that we can fine tune formatting on as needed later.

There are still some details to polish up, and I'll note inline as needed - but this should at least progress things forward as much as possible at this point, as we continue to refine the requirements for exactly how the form UX should work.

Example of form state persisting through steps in different containers / compositions:
form_state_steps

Example analytics output:
Screen Shot 2022-09-08 at 3 33 41 PM

### DIFF
--- a/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
+++ b/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
@@ -1,7 +1,11 @@
 package com.appcues.analytics
 
+import com.appcues.analytics.ExperienceLifecycleEvent.StepInteraction.InteractionType.FORM_SUBMITTED
 import com.appcues.data.model.Experience
 import com.appcues.statemachine.Error
+import com.appcues.ui.ExperienceStepFormItemState
+import com.appcues.ui.ExperienceStepFormItemState.OptionSelectFormItemState
+import com.appcues.ui.ExperienceStepFormItemState.TextInputFormItemState
 
 internal sealed class ExperienceLifecycleEvent(
     open val experience: Experience,
@@ -16,7 +20,12 @@ internal sealed class ExperienceLifecycleEvent(
     data class StepInteraction(
         override val experience: Experience,
         val stepIndex: Int,
-    ) : ExperienceLifecycleEvent(experience, AnalyticsEvent.ExperienceStepInteraction)
+        val interactionType: InteractionType,
+    ) : ExperienceLifecycleEvent(experience, AnalyticsEvent.ExperienceStepInteraction) {
+        enum class InteractionType {
+            FORM_SUBMITTED
+        }
+    }
 
     data class StepCompleted(
         override val experience: Experience,
@@ -72,6 +81,9 @@ internal sealed class ExperienceLifecycleEvent(
                 this["stepIndex"] = "${experience.groupLookup[it] ?: 0},${experience.stepIndexLookup[it] ?: 0}"
                 this["stepType"] = step.type
             }
+            customProperties?.let {
+                putAll(it)
+            }
             error?.let {
                 this["message"] = it.message
                 this["errorId"] = it.id.toString()
@@ -94,5 +106,43 @@ internal sealed class ExperienceLifecycleEvent(
             is StepError -> stepError
             is ExperienceError -> experienceError
             else -> null
+        }
+
+    private val customProperties: HashMap<String, Any>?
+        get() = when (this) {
+            is StepInteraction -> {
+                flatStepIndex?.let {
+                    val step = experience.flatSteps[it]
+                    hashMapOf(
+                        "interactionType" to interactionType.analyticsName(),
+                        "interactionData" to hashMapOf(
+                            "formResponse" to step.formState.formItems.map { formItem ->
+                                formItem.analyticsProperties()
+                            }
+                        ),
+                    )
+                }
+            }
+            else -> null
+        }
+
+    private fun StepInteraction.InteractionType.analyticsName() =
+        when (this) {
+            FORM_SUBMITTED -> "Form Submitted"
+        }
+
+    private fun ExperienceStepFormItemState.analyticsProperties() =
+        hashMapOf<String, Any>(
+            "fieldId" to id.toString(),
+            "fieldType" to type,
+            "fieldRequired" to isRequired,
+            "label" to label,
+            "value" to formattedValues()
+        )
+
+    private fun ExperienceStepFormItemState.formattedValues() =
+        when (this) {
+            is TextInputFormItemState -> value
+            is OptionSelectFormItemState -> values.joinToString(",") // need actual CSV-ifying
         }
 }
--- a/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
+++ b/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
@@ -7,6 +7,8 @@ import com.appcues.analytics.ExperienceLifecycleEvent.ExperienceError
 import com.appcues.analytics.ExperienceLifecycleEvent.ExperienceStarted
 import com.appcues.analytics.ExperienceLifecycleEvent.StepCompleted
 import com.appcues.analytics.ExperienceLifecycleEvent.StepError
+import com.appcues.analytics.ExperienceLifecycleEvent.StepInteraction
+import com.appcues.analytics.ExperienceLifecycleEvent.StepInteraction.InteractionType.FORM_SUBMITTED
 import com.appcues.analytics.ExperienceLifecycleEvent.StepSeen
 import com.appcues.statemachine.Error
 import com.appcues.statemachine.State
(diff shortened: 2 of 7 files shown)


appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
```diff
@@ -16,7 +20,12 @@ internal sealed class ExperienceLifecycleEvent(
     data class StepInteraction(
         override val experience: Experience,
         val stepIndex: Int,
-    ) : ExperienceLifecycleEvent(experience, AnalyticsEvent.ExperienceStepInteraction)
+        val interactionType: InteractionType,
+    ) : ExperienceLifecycleEvent(experience, AnalyticsEvent.ExperienceStepInteraction) {
+        enum class InteractionType {
+            FORM_SUBMITTED
+        }
```
iujames, line 27: right now this is the only type, but leaving this extensible for the future to try to match the spirit of the event spec

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
```diff
@@ -95,4 +107,42 @@ internal sealed class ExperienceLifecycleEvent(
...
+            "fieldType" to type,
+            "fieldRequired" to isRequired,
+            "label" to label,
+            "value" to formattedValues()
+        )
+
+    private fun ExperienceStepFormItemState.formattedValues() =
+        when (this) {
+            is TextInputFormItemState -> value
+            is OptionSelectFormItemState -> values.joinToString(",") // need actual CSV-ifying
```
iujames, line 146: CSV format is a remaining TODO - want to confirm this is OK with platform still

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -55,6 +57,15 @@ internal class ExperienceLifecycleTracker(
...
                         trackLifecycleEvent(StepCompleted(it.experience, it.flatStepIndex))
+
+                        // TESTING!
+                        // this is a spot where we can output the interaction analytics to validate structure, but still TBD
+                        // on the rules around when these should be submitted related to actions in the UI.  Also, may need
+                        // a way to plug in user profile attributes with the form answers, as web does.
+                        val formState = it.experience.flatSteps[it.flatStepIndex].formState
+                        if (formState.formItems.any() && formState.isFormComplete.value) {
+                            trackLifecycleEvent(StepInteraction(it.experience, it.flatStepIndex, FORM_SUBMITTED))
+                        }
```
iujames, line 68: this block can be removed before merge, but it's helpful for testing the analytics basics

appcues/src/main/java/com/appcues/ui/AppcuesActivity.kt
```diff
@@ -242,7 +242,7 @@ internal class AppcuesActivity : AppCompatActivity() {
                 .padding(paddingValues = stepDecoratingPadding.paddingValues.value)
                 .testTag("page_$index")
         ) {
-            CompositionLocalProvider(LocalExperienceStepFormStateDelegate provides ExperienceStepFormState()) {
+            CompositionLocalProvider(LocalExperienceStepFormStateDelegate provides formState) {
```
iujames, line 245: `formState` here now is directly form the Step

appcues/src/main/java/com/appcues/ui/ExperienceStepFormState.kt
```diff
@@ -1,39 +1,102 @@
...
+        updateFormItem(item)
+    }
+
+    private fun getItem(primitive: TextInputPrimitive): TextInputFormItemState {
+        var item = _formItems[primitive.id] as? TextInputFormItemState
+        if (item != null) return item
+
+        // create new state tracking object
+        item = with(primitive) {
+            TextInputFormItemState(itemIndex++, id, "textInput", label.text, required, defaultValue ?: "")
```
iujames, line 45: this `itemIndex++` is an attempt to keep an ordered set of form inputs so the analytics can be ordered in the same fashion as the JSON structure - it _seems_ to work, but I'm not sure if always guaranteed to do so.

appcues/src/main/java/com/appcues/ui/ExperienceStepFormState.kt
```diff
@@ -1,39 +1,102 @@
 package com.appcues.ui
 
 import androidx.compose.runtime.mutableStateOf
+import com.appcues.data.model.ExperiencePrimitive.OptionSelectPrimitive
+import com.appcues.data.model.ExperiencePrimitive.TextInputPrimitive
+import com.appcues.ui.ExperienceStepFormItemState.OptionSelectFormItemState
+import com.appcues.ui.ExperienceStepFormItemState.TextInputFormItemState
 import java.util.UUID
 
 internal class ExperienceStepFormState {
```
iujames, line 10: this class now maintains all the responsibility for getting and setting form state values, and creating and retaining references to form item state on demand. The primitives are thus simplified a bit and just interact with the get/set value functions.

appcues/src/main/java/com/appcues/ui/ExperienceStepFormState.kt
```diff
@@ -1,39 +1,102 @@
...
             return when (this) {
-                is MultipleTextFormItem -> !isRequired || values.isNotEmpty()
-                is SingleTextFormItem -> !isRequired || value.isNotBlank()
+                is OptionSelectFormItemState -> !isRequired || values.isNotEmpty()
+                is TextInputFormItemState -> !isRequired || value.isNotBlank()
             }
         }
 
-    class SingleTextFormItem(
+    class TextInputFormItemState(
```
iujames, line 85: this was renamed, as was the other option, to more explicitly tie to the primitive types that use them.  I felt like it was better to be a bit more clear about that and the general naming wasn't really serving much purpose.

appcues/src/main/java/com/appcues/ui/primitive/TextInputPrimitive.kt
```diff
@@ -60,6 +54,7 @@ internal fun TextInputPrimitive.Compose(modifier: Modifier) {
             onValueChange = {
                 if (maxLength == null || it.length <= maxLength) {
                     text.value = it
+                    formState.setValue(this@Compose, it)
```
iujames, line 57: these primitives are mostly simplified now, but they do have this one additional requirement of reporting back to the persisted form state on update.

appcues/src/main/java/com/appcues/ui/primitive/TextInputPrimitive.kt
```diff
@@ -42,11 +40,7 @@ import java.util.UUID
 internal fun TextInputPrimitive.Compose(modifier: Modifier) {
 
     val formState = LocalExperienceStepFormStateDelegate.current
-    val text = remember { mutableStateOf(defaultValue ?: "") }
-
-    LaunchedEffect(key1 = text.value) {
-        formState.captureFormItem(this@Compose.id, SingleTextFormItem(label.text, required, text.value))
-    }
+    val text = remember { mutableStateOf(formState.getValue(this)) }
```
iujames, line 43: the value now comes from the form state, so it is restored when you return to this composition from elsewhere, if a value was already set.

appcues/src/main/java/com/appcues/ui/ExperienceStepFormState.kt
andretortolano: I think it should work, compositions happen in the same thread and its linear.

appcues/src/main/java/com/appcues/ui/primitive/OptionSelectPrimitive.kt
```diff
@@ -55,14 +52,11 @@ import com.appcues.ui.extensions.styleBorder
...
-    val selectedValues = remember { mutableStateOf(defaultValue) }
+    val selectedValues = remember { mutableStateOf(formState.getValue(this)) }
 
-    LaunchedEffect(key1 = selectedValues.value) {
-        val formItem = when (selectMode) {
-            MULTIPLE -> MultipleTextFormItem(label.text, required, selectedValues.value)
-            SINGLE -> SingleTextFormItem(label.text, required, selectedValues.value.firstOrNull() ?: "")
-        }
-        formState.captureFormItem(this@Compose.id, formItem)
+    fun setSelectedValues(values: Set<String>) {
```
andretortolano, line 57: I'm not sure if its Ok to declare a function inside a composition

andretortolano (approved):

appcues/src/main/java/com/appcues/ui/primitive/OptionSelectPrimitive.kt
andretortolano: no strong opinion, I was wondering what happens when during the runtime the composition reads a function there, but maybe this is nothing to be concerned about, probably same thing happens with inline functions no that I think abou it

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
iujames: @andretortolano thoughts on using this CSV lib to encode? https://github.com/doyaaaaaken/kotlin-csv

We could also roll our own, but need to make sure we have proper handling of values with commas in them, escape double quotes and newlines.

iujames: ok, will go ahead and progress this one through then cycle back to this.

PR #230, opened 2022-09-09
Ensure selectedContent is used, if available, for selected item(s) in optionSelect

iujames:
Bug found in testing recent work - this is so the custom selection mode (not a typical radio/checkbox) like the `8` option shown here will use the defined selected style.

More generally, the data model allows optional custom selected style in any context, which should now be respected.

Screen Shot 2022-09-09 at 10 09 30 AM

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/primitive/OptionSelectPrimitive.kt
+++ b/appcues/src/main/java/com/appcues/ui/primitive/OptionSelectPrimitive.kt
@@ -151,22 +151,22 @@ private fun List<OptionSelectPrimitive.OptionItem>.ComposeSelections(
         when (controlPosition) {
             LEADING -> Row(verticalAlignment = Alignment.CenterVertically) {
                 selectMode.Compose(isSelected, selectedColor, unselectedColor, accentColor) { updateSelection(it.value, !isSelected) }
-                it.content.Compose()
+                it.contentView(isSelected).Compose()
             }
             TRAILING -> Row(verticalAlignment = Alignment.CenterVertically) {
-                it.content.Compose()
+                it.contentView(isSelected).Compose()
                 selectMode.Compose(isSelected, selectedColor, unselectedColor, accentColor) { updateSelection(it.value, !isSelected) }
             }
             TOP -> Column(horizontalAlignment = Alignment.CenterHorizontally) {
                 selectMode.Compose(isSelected, selectedColor, unselectedColor, accentColor) { updateSelection(it.value, !isSelected) }
-                it.content.Compose()
+                it.contentView(isSelected).Compose()
             }
             BOTTOM -> Column(horizontalAlignment = Alignment.CenterHorizontally) {
-                it.content.Compose()
+                it.contentView(isSelected).Compose()
                 selectMode.Compose(isSelected, selectedColor, unselectedColor, accentColor) { updateSelection(it.value, !isSelected) }
             }
             HIDDEN -> Box(modifier = Modifier.clickable { updateSelection(it.value, !isSelected) }) {
-                it.content.Compose()
+                it.contentView(isSelected).Compose()
             }
         }
     }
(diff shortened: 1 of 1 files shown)


appcues/src/main/java/com/appcues/ui/primitive/OptionSelectPrimitive.kt
```diff
@@ -151,22 +151,22 @@ private fun List<OptionSelectPrimitive.OptionItem>.ComposeSelections(
...
+                it.contentView(isSelected).Compose()
             }
             BOTTOM -> Column(horizontalAlignment = Alignment.CenterHorizontally) {
-                it.content.Compose()
+                it.contentView(isSelected).Compose()
                 selectMode.Compose(isSelected, selectedColor, unselectedColor, accentColor) { updateSelection(it.value, !isSelected) }
             }
             HIDDEN -> Box(modifier = Modifier.clickable { updateSelection(it.value, !isSelected) }) {
-                it.content.Compose()
+                it.contentView(isSelected).Compose()
```
iujames, line 169: these are the normal cases when rendering selection content in a vertical or horizontal list

appcues/src/main/java/com/appcues/ui/primitive/OptionSelectPrimitive.kt
```diff
@@ -216,15 +216,15 @@ private fun List<OptionSelectPrimitive.OptionItem>.ComposePicker(
...
     Box(modifier = Modifier.clickable(onClick = { expanded = true })) {
         // 1. render the selected item as the collapsed state
         Row(modifier = modifier.fillMaxWidth(), verticalAlignment = Alignment.CenterVertically) {
             // we should always either have a selected item, or a placeholder, or this will just be a blank
             // box with a dropdown arrow
-            val selectedValue = selectedValues.firstOrNull()
             if (selectedValue != null) {
                 // if we have a selected value, we assume one of our options will match it, and compose it
-                this@ComposePicker.firstOrNull { it.value == selectedValue }?.content?.Compose()
+                this@ComposePicker.firstOrNull { it.value == selectedValue }?.contentView(true)?.Compose()
```
iujames, line 227: This is the picker style. I'm not sure if this is desired, but seems valid... when rendering the "closed" state of the picker with the selected item visible, if an item has a selected state defined - it is used.

appcues/src/main/java/com/appcues/ui/primitive/OptionSelectPrimitive.kt
```diff
@@ -250,9 +250,16 @@ private fun List<OptionSelectPrimitive.OptionItem>.ComposePicker(
                     expanded = false
                     valueSelectionChanged(setOf(it.value))
                 }) {
-                    it.content.Compose()
+                    it.contentView(selectedValue == it.value).Compose()
```
iujames, line 253: This is the picker expanded menu - if the item in the menu list is the selected item, it is highlighted using its selected style, if exists - again unsure if this is desired for our picker behavior but seems valid.

appcues/src/main/java/com/appcues/ui/primitive/OptionSelectPrimitive.kt
```diff
@@ -250,9 +250,16 @@ private fun List<OptionSelectPrimitive.OptionItem>.ComposePicker(
...
                 }
             }
         }
     }
 }
+
+private fun OptionSelectPrimitive.OptionItem.contentView(isSelected: Boolean) =
```
andretortolano, line 260: this logic already works, a compose way of doing this is to use derivedStateOf to keep the state of "content" based on the selectedState. this way calculation happens only one (I think, we could test this out)

I'm still learning this but from the example I guess we could make it work somewhat like this:

```kotlin

    val contentView by remember(isSelected) {
        derivedStateOf { 
             selectedContent?.let { if(isSelected) it else content } ?: run { content }
         }
    }
```

andretortolano (approved):

appcues/src/main/java/com/appcues/ui/primitive/OptionSelectPrimitive.kt
iujames: Thanks for the tip, does this capture what you had in mind @andretortolano ? https://github.com/appcues/appcues-android-sdk/pull/230/commits/5604cd59025519e01a48d3df8ac3271201bb7c7f

andretortolano: yes!

PR #231, opened 2022-09-09
Fix horizontal alignment of content within Button of larger width

iujames:
This change allows button content to be horizontally aligned leading/trailing, if inside of a button that is not autosized - a fixed or full width button.

Screen Shot 2022-09-09 at 11 56 09 AM

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/primitive/ButtonPrimitive.kt
+++ b/appcues/src/main/java/com/appcues/ui/primitive/ButtonPrimitive.kt
@@ -1,12 +1,11 @@
 package com.appcues.ui.primitive
 
 import androidx.compose.foundation.isSystemInDarkTheme
-import androidx.compose.foundation.layout.Arrangement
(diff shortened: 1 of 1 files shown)


appcues/src/main/java/com/appcues/ui/primitive/ButtonPrimitive.kt
```diff
@@ -28,16 +28,23 @@ internal fun ButtonPrimitive.Compose(modifier: Modifier) {
         color = Color.Transparent,
     ) {
         ProvideTextStyle(style.getTextStyle(LocalContext.current, isSystemInDarkTheme())) {
-            Row(
-                horizontalArrangement = Arrangement.Center,
-                verticalAlignment = Alignment.CenterVertically,
+            Box(
+                contentAlignment = content.style.getBoxAlignment(),
```
andretortolano, line 32: nice catch here. I left center by default but this is in align with convo from last week

andretortolano (approved):

PR #233, opened 2022-09-12
Additional clean up on form controls from test review

iujames:
Some touch up work on the optionSelect and textInput components from recent snapshot test review and related iOS work.

* The optionSelect primitive now uses the `controlPosition` alignment value to derive the alignment of the parent vertical or horizontal stack.  ex: a verticalList layout with controlPosition of trailing will align the rows within the column to trailing edge - allowing the checkboxes or radio buttons to line up consistently.
* The ComposeContent helper is updated to wrap the composition in the AppcuesTheme, as it was discovered that the lack of this could lead to some discrepancy in the snapshot test results versus actual rendered content - this should make the tests more realistic
* The textInput primitive will now respect additional styling settings that were missed before - such as backgroundColor and additional padding (16dp default is fixed in the Compose TextField)
* The textInput primitive will set it's height based on the `numberOfLines` property - so a 3 line textInput will allocate a height to show all three lines, rather than a single line that grows as you type (default Compose behavior)

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/primitive/TextInputPrimitive.kt
+++ b/appcues/src/main/java/com/appcues/ui/primitive/TextInputPrimitive.kt
@@ -2,7 +2,9 @@ package com.appcues.ui.primitive
 
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.defaultMinSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.material.LocalTextStyle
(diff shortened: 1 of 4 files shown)


appcues/src/main/java/com/appcues/ui/primitive/TextInputPrimitive.kt
```diff
@@ -77,14 +106,36 @@ internal fun TextInputPrimitive.Compose(modifier: Modifier) {
...
-                cursorColor = cursorColor?.getColor(isSystemInDarkTheme()) ?: MaterialTheme.colors.primary,
+                cursorColor = cursorColor?.getColor(isDark) ?: MaterialTheme.colors.primary,
                 focusedIndicatorColor = Color.Transparent,
                 unfocusedIndicatorColor = Color.Transparent,
             ),
         )
     }
 }
 
+private fun Modifier.styleInputBoxHeight(textStyle: TextStyle, numberOfLines: Int) = this.then(
```
andretortolano, line 117: nice!

andretortolano (approved):

PR #234, opened 2022-09-19
Survey control updates and analytics

iujames:
This is the next pass at our survey control analytics, cleaning up a few things along the way.

* support adding the question/answer pairs as profile attributes with `_appcuesForm_` prefix, matching web behavior. This occurs in `ExperienceLifecycleTracker.trackFormSubmission`
* support showing the form results interaction data nicely in the debugger, matching iOS.  This led to some more cleanup where I allowed storing the `ExperienceStepFormState` in the custom attributes of the event directly, then use a Moshi adapter to control how the resulting "formResult" JSON mapping occurs on the network request.  Having the full object model in the in-memory attributes makes it simpler to reference this data in the debugger, not having to dig into nested maps - and is very much like the approach Matt took on iOS here https://github.com/appcues/appcues-ios-sdk/pull/242, which I liked.
* small update to the way form state is referenced from the Composables for textInput and optionSelect - allow them to use mutableState objects directly from the form state container, rather than having a duplicate mirrored copy locally. This was mostly inspired by how the view model was referenced in a similar way in examples here https://medium.com/androiddevelopers/effective-state-management-for-textfield-in-compose-d6e5b070fbe5

The form response data can now be seen on the debugger events like this, at the bottom.
Screen Shot 2022-09-19 at 2 21 12 PM

next up:
There is still more to sort out around the rules for if and when form data can be submitted - right now it's just directly tied to step completion events.  Also, will need to add the handling for required fields and error messaging when not all filled out.

### DIFF
--- a/appcues/src/main/java/com/appcues/analytics/AnalyticsTracker.kt
+++ b/appcues/src/main/java/com/appcues/analytics/AnalyticsTracker.kt
@@ -29,13 +29,17 @@ internal class AnalyticsTracker(
         }
     }
 
-    fun identify(properties: Map<String, Any>? = null) {
+    fun identify(properties: Map<String, Any>? = null, interactive: Boolean = true) {
         if (!analyticsPolicy.canIdentify()) return
 
         activityBuilder.identify(properties).let {
             updateAnalyticsFlow(IDENTIFY, false, it)
 
-            analyticsQueueProcessor.flushThenSend(it)
+            if (interactive) {
+                analyticsQueueProcessor.flushThenSend(it)
+            } else {
+                analyticsQueueProcessor.queue(it)
+            }
         }
     }
 
--- a/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
+++ b/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
@@ -1,12 +1,8 @@
 package com.appcues.analytics
 
-import com.appcues.analytics.ExperienceLifecycleEvent.Companion.FORM_RESPONSE_KEY
 import com.appcues.analytics.ExperienceLifecycleEvent.StepInteraction.InteractionType.FORM_SUBMITTED
 import com.appcues.data.model.Experience
 import com.appcues.statemachine.Error
-import com.appcues.ui.ExperienceStepFormItemState
-import com.appcues.ui.ExperienceStepFormItemState.OptionSelectFormItemState
-import com.appcues.ui.ExperienceStepFormItemState.TextInputFormItemState
 import com.appcues.ui.ExperienceStepFormState
 import com.appcues.util.toSlug
 
@@ -17,7 +13,6 @@ internal sealed class ExperienceLifecycleEvent(
     companion object {
         const val INTERACTION_TYPE_KEY = "interactionType"
         const val INTERACTION_DATA_KEY = "interactionData"
-        const val FORM_RESPONSE_KEY = "formResponse"
         const val FORM_SUBMITTED_INTERACTION_TYPE = "Form Submitted"
     }
 
@@ -124,7 +119,9 @@ internal sealed class ExperienceLifecycleEvent(
                     val step = experience.flatSteps[it]
                     hashMapOf(
                         INTERACTION_TYPE_KEY to interactionType.analyticsName(),
-                        INTERACTION_DATA_KEY to step.formState.formattedAsFormResponse(),
+                        INTERACTION_DATA_KEY to when(interactionType) {
+                            FORM_SUBMITTED -> step.formState
+                        },
                     )
                 }
             }
@@ -137,24 +134,6 @@ internal sealed class ExperienceLifecycleEvent(
         }
 }
 
-private fun ExperienceStepFormState.formattedAsFormResponse() = hashMapOf(
-    FORM_RESPONSE_KEY to formItems.map { formItem ->
-        hashMapOf<String, Any>(
-            "fieldId" to formItem.id.toString(),
-            "fieldType" to formItem.type,
-            "fieldRequired" to formItem.isRequired,
-            "label" to formItem.label,
-            "value" to formItem.formattedValues()
-        )
-    }
-)
-
-private fun ExperienceStepFormItemState.formattedValues() =
-    when (this) {
-        is TextInputFormItemState -> text.value
-        is OptionSelectFormItemState -> values.value.joinToString(",") // need actual CSV-ifying
-    }
-
 internal fun ExperienceStepFormState.formattedAsProfileUpdate() = formItems.associate {
-    "_appcuesForm_${it.label.toSlug()}" to it.formattedValues()
+    "_appcuesForm_${it.label.toSlug()}" to it.value
 }
--- a/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
+++ b/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
@@ -6,6 +6,8 @@ import com.appcues.statemachine.Error
 import com.appcues.ui.ExperienceStepFormItemState
 import com.appcues.ui.ExperienceStepFormItemState.OptionSelectFormItemState
 import com.appcues.ui.ExperienceStepFormItemState.TextInputFormItemState
+import com.appcues.ui.ExperienceStepFormState
+import com.appcues.util.toSlug
 
 internal sealed class ExperienceLifecycleEvent(
     open val experience: Experience,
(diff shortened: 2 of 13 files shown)


appcues/src/main/java/com/appcues/analytics/AnalyticsTracker.kt
```diff
@@ -29,13 +29,17 @@ internal class AnalyticsTracker(
...
 
         activityBuilder.identify(properties).let {
             updateAnalyticsFlow(IDENTIFY, false, it)
 
-            analyticsQueueProcessor.flushThenSend(it)
+            if (interactive) {
+                analyticsQueueProcessor.flushThenSend(it)
+            } else {
+                analyticsQueueProcessor.queue(it)
+            }
```
iujames, line 42: we now support non-interactive `identify` calls - so that the new `_appcuesForm_` attributes can be merged in with the flow events and do not need to directly trigger another call to the `/qualify` endpoint.

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
```diff
@@ -114,12 +118,10 @@ internal sealed class ExperienceLifecycleEvent(
...
                     hashMapOf(
-                        "interactionType" to interactionType.analyticsName(),
-                        "interactionData" to hashMapOf(
-                            "formResponse" to step.formState.formItems.map { formItem ->
-                                formItem.analyticsProperties()
-                            }
-                        ),
+                        INTERACTION_TYPE_KEY to interactionType.analyticsName(),
+                        INTERACTION_DATA_KEY to when (interactionType) {
+                            FORM_SUBMITTED -> step.formState
```
iujames, line 123: this is where we now just store the form state object directly in the properties and have a custom serializer adapter in Moshi that controls how the output looks (see new `ExperienceStepFormStateAdapater`)

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
```diff
@@ -128,21 +130,10 @@ internal sealed class ExperienceLifecycleEvent(
...
+}
 
-    private fun ExperienceStepFormItemState.analyticsProperties() =
-        hashMapOf<String, Any>(
-            "fieldId" to id.toString(),
-            "fieldType" to type,
-            "fieldRequired" to isRequired,
-            "label" to label,
-            "value" to formattedValues()
-        )
```
iujames, line 141: this stuff now controlled directly by the new Moshi adapter

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -54,6 +57,7 @@ internal class ExperienceLifecycleTracker(
                         trackLifecycleEvent(StepSeen(it.experience, it.flatStepIndex))
                     }
                     is EndingStep -> {
+                        trackFormSubmission(it.experience, it.flatStepIndex)
```
iujames, line 60: this is where we'll need more conditional logic in the future to determine if and when form results should be submitted - right now its just on every ending step transition.

appcues/src/main/java/com/appcues/data/remote/adapters/ExperienceStepFormStateAdapter.kt
```diff
@@ -0,0 +1,39 @@
...
+                ExperienceStepFormStateRequestItem(
+                    fieldId = it.id,
+                    fieldType = it.type,
+                    fieldRequired = it.isRequired,
+                    label = it.label,
+                    value = it.value,
+                )
+            }
+        )
+    }
```
iujames, line 38: Moshi adapters are quite powerful! allowing us to transform one object `ExperienceStepFormState` into another - the Map<...> result here - as long as the new result type is something that can be directly serialized to JSON.  Since it's using the `*RequestItem` helper class here with a generated adapter - it all just works.

appcues/src/main/java/com/appcues/debugger/DebuggerRecentEventsManager.kt
```diff
@@ -65,8 +67,12 @@ internal class DebuggerRecentEventsManager(
...
                 timestamp = (request.profileUpdate?.get(AutoPropertyDecorator.UPDATED_AT_PROPERTY) as Long?) ?: Date().time,
                 name = request.userId,
-                properties = request.profileUpdate?.toSortedList(),
-                identityProperties = null
+                propertySections = listOf(
+                    DebuggerEventItemPropertySection(
+                        title = contextResources.getString(R.string.appcues_debugger_event_details_properties_title),
+                        properties = request.profileUpdate?.toSortedList(),
+                    )
+                )
```
iujames, line 75: small refactor here - where instead of having `properties` and `identityProperties` we now just have a collection of `propertySections` -- an extensible list that we can add or remove sections from as desired.  This supports adding the new section for the form inputs down on line 123.

appcues/src/main/java/com/appcues/debugger/DebuggerRecentEventsManager.kt
```diff
@@ -178,3 +185,32 @@ internal class DebuggerRecentEventsManager(
...
+
+private fun Map<String, Any>.getFormResponse(): Map<String, Any?> {
+    return (this[ExperienceLifecycleEvent.INTERACTION_DATA_KEY] as? ExperienceStepFormState)?.let {
+        it.formItems.associate { itemState ->
+            itemState.label to itemState.value
+        }
+    } ?: mapOf()
+}
+
+private fun List<Pair<String, Any?>>.sortByPropertyName(): List<Pair<String, Any?>> = sortedBy { it.first }
```
iujames, line 216: these helpers got moved down to the file level and not nested in the class to make Detekt happy - it does not like having too many functions in a class.  These are not really functions of the class but just extension helpers that get used, so moved them out.

appcues/src/main/java/com/appcues/debugger/ui/details/DebuggerEventDetails.kt
```diff
@@ -59,16 +59,11 @@ internal fun DebuggerEventDetails(debuggerEventItem: DebuggerEventItem?, onBackP
...
-
-        if (debuggerEventItem.identityProperties != null && debuggerEventItem.identityProperties.isNotEmpty()) {
-            identityPropertiesTitle()
-
-            properties(debuggerEventItem.identityProperties)
+        debuggerEventItem.propertySections.forEach {
+            if (it.properties != null && it.properties.isNotEmpty()) {
+                propertiesTitle(it.title)
+                properties(it.properties)
+            }
```
iujames, line 66: the debugger UI is now slightly simpler, just looping over the sections provided and using the title provided

appcues/src/main/java/com/appcues/ui/ExperienceStepFormState.kt
```diff
@@ -19,11 +19,11 @@ internal class ExperienceStepFormState {
     val isFormComplete = mutableStateOf(true)
 
     fun getValue(primitive: TextInputPrimitive) =
-        getItem(primitive).value
+        getItem(primitive).text
```
iujames, line 22: the form state now just returns mutableState values that the Composables directly bind to

appcues/src/main/java/com/appcues/ui/primitive/TextInputPrimitive.kt
```diff
@@ -77,10 +74,9 @@ internal fun TextInputPrimitive.Compose(modifier: Modifier) {
 
         // Several styling customization options for TextField noted here https://stackoverflow.com/a/68592613
         TextField(
-            value = text.value,
+            value = formState.getValue(this@Compose).value,
```
iujames, line 77: where we now just use the mutableState from the form state, rather than having a shadow copy in a separate `text` variable like we had before.

appcues/src/main/java/com/appcues/util/StringExt.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import java.util.Locale
+
+fun String.toSlug() = lowercase(Locale.getDefault())
+    .replace("\n", " ")
+    .replace("[^a-z\\d\\s]".toRegex(), " ")
+    .split(" ")
+    .joinToString("-")
+    .replace("-+".toRegex(), "-")
+    .trimEnd('-')
```
iujames, line 11: used when generating the user property name from a question label - matching existing web and ios logic

appcues/src/main/java/com/appcues/debugger/ui/details/DebuggerEventDetails.kt
andretortolano: nice improv to the debugger UI

appcues/src/main/java/com/appcues/util/StringExt.kt
andretortolano: whats the meaning of "[^a-z\\d\\s]"? on line 7? replacing special characters with white space?

andretortolano (approved):

appcues/src/main/java/com/appcues/util/StringExt.kt
iujames: right, this is taken from https://stackoverflow.com/a/57298659 -- thats the idea though, normalize any survey question into a slugified name that is just normal chars or numbers separated by hyphens.

PR #235, opened 2022-09-20
Do not trigger step_completed when dismissing experience early

iujames:
This ended up being an easy fix, so knocking out.  We already knew in the EndExperience action whether we should mark a flow complete or use the default behavior (if on last step, its complete).  This just updates the EndingStep state to also be aware of this flag, and then add similar logic around whether or not to record step_completion analytic.

For now, form submission (step_interaction) is also tied to this same logic, but this will change in the near future as we build the new survey submission action support.

### DIFF
--- a/appcues/src/main/java/com/appcues/statemachine/Transitions.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/Transitions.kt
@@ -71,7 +71,7 @@ internal interface Transitions {
                     // in different groups we want to wait for StartStep action from AppcuesViewModel
                     val response = CompletableDeferred<ResultOf<State, Error>>()
                     Transition(
-                        state = EndingStep(experience, flatStepIndex) {
+                        state = EndingStep(experience, flatStepIndex, true) {
                             coroutineScope.launch {
                                 response.complete(continuation())
                             }
@@ -81,7 +81,7 @@ internal interface Transitions {
                 } else {
                     // in same group we can continue to StartStep internally
                     Transition(
-                        state = EndingStep(experience, flatStepIndex, null),
+                        state = EndingStep(experience, flatStepIndex, true, null),
                         sideEffect = ContinuationEffect(StartStep(action.stepReference)),
                     )
                 }
@@ -101,7 +101,7 @@ internal interface Transitions {
             // this means the AppcuesActivity was destroyed externally (i.e. deep link) and we should
             // immediately transition to EndingExperience - not rely on the UI to do it for us (it's gone)
             Transition(
-                state = EndingStep(experience, flatStepIndex, null),
+                state = EndingStep(experience, flatStepIndex, action.markComplete, null),
                 sideEffect = ContinuationEffect(action)
             )
         } else {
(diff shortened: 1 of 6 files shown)


appcues/src/main/java/com/appcues/statemachine/Transitions.kt
```diff
@@ -71,7 +71,7 @@ internal interface Transitions {
                     // in different groups we want to wait for StartStep action from AppcuesViewModel
                     val response = CompletableDeferred<ResultOf<State, Error>>()
                     Transition(
-                        state = EndingStep(experience, flatStepIndex) {
+                        state = EndingStep(experience, flatStepIndex, true) {
```
iujames, line 74: the case here and on line 84 - we're ending a step because we are getting a StartStep action for a new step - in those cases, we always consider the step we're ending to be completed, as we're moving elsewhere.

appcues/src/main/java/com/appcues/statemachine/Transitions.kt
```diff
@@ -101,7 +101,7 @@ internal interface Transitions {
             // this means the AppcuesActivity was destroyed externally (i.e. deep link) and we should
             // immediately transition to EndingExperience - not rely on the UI to do it for us (it's gone)
             Transition(
-                state = EndingStep(experience, flatStepIndex, null),
+                state = EndingStep(experience, flatStepIndex, action.markComplete, null),
```
iujames, line 104: in this case, and line 115, the step is ending because of the EndExperience action - and this is where we need to actually detect whether its been marked completed, or is on the last step -- the skippable trait will not mark it completed, giving the desired behavior of not triggering step_completed if skipped.

appcues/src/test/java/com/appcues/analytics/ExperienceLifecycleTrackerTest.kt
```diff
@@ -0,0 +1,188 @@
...
+                        it.dismissAndContinue?.invoke()
+                    }
+                    // ignore other state changes
+                    else -> Unit
+                }
+            }
+        }
+
+        coroutineScope.launch {
+            scope.get<ExperienceLifecycleTracker>().start(UnconfinedTestDispatcher())
```
andretortolano, line 150: I got why this was done, there is also another way we could do it. simply by creating an interface and having Koin give different implementations when its running the SDK vs when its testing.

```kotlin
interface AppcuesDispatchers {
   fun main()
   fun io()
}

// for real exe
class DefaultDispatchers: AppcuesDispatchers {
   fun main() = Dispatchers.MAIN
   fun io() = Dispatchers.IO
}

// for tests
class TestDispatchers: AppcuesDispatchers {
  val dispatcher = UnconfinedTestDispatcher()
  fun main() = dispatcher
  fun io() = dispatcher
}
```

if for some reason we see this again we can consider refactoring to this if it looks interesting

andretortolano (approved):

appcues/src/test/java/com/appcues/analytics/ExperienceLifecycleTrackerTest.kt
iujames: sorry i meant to make a note about that and left it as a pending comment but glad you got it anyway.  Yes, that sounds like a good suggestion to keep in mind to use Koin to assist there, thanks!

PR #239, opened 2022-09-29
♻️  Reorganizing compositions for better reusability

andretortolano:
code refactoring that will allow reusability of AppcuesComposition.

* change to SkippableTrait from fillMaxSize to matchParent size so it wont break when used in other presenting types (like tooltips and embeds
* package re-organizing
* ui tests are running successfully after changes


iujames (approved):
looks like a nice organization to me 👍  nice to separate the AppcuesComposition and step/container composition pieces from the AppcuesActivity.

PR #241, opened 2022-10-04
Add the @appcues/submit-form action, related updates

iujames:
This change parallels what was done on the iOS side in https://github.com/appcues/appcues-ios-sdk/pull/247

* new `ExperienceActionQueueTransforming` interface, which `ExperienceAction`s can optionally use to indicate they need to be able to transform the action queue.  This is used by the submit-form action to remove subsequent actions when validation fails - block step progression.
* update `ActionProcessor` to use this new queue transformation capability when processing actions from the Experience. This was updated to allow passing the full List<ExperienceAction> to process on any button tap, rather than each one by one. A small refactor, but supports the need to be able to allow one action to transform the queue before processing.
* new `@appcues/submit-form` action - handles the logic to validate (conditionally) and block later actions if needed. Moves the analytics calls to report the user profile and step_interaction tracking to this action - and out of the lifecycle event handler
* unit tests for the new `ActionProcessor` queue handling logic and the `SubmitFormAction` behaviors


appcues/src/main/java/com/appcues/action/ActionProcessor.kt
```diff
@@ -26,9 +26,21 @@ internal class ActionProcessor(override val scope: Scope) : KoinScopeComponent {
...
 
-    fun process(action: ExperienceAction) {
+    fun process(actions: List<ExperienceAction>) {
         appcuesCoroutineScope.launch {
-            actionQueue.send(action)
+            transformQueue(actions).forEach {
+                actionQueue.send(it)
+            }
+        }
+    }
```
iujames, line 35: this is where we now take in the full list (all of the actions attached to a single button) at once - allows for the new queue transformation to occur.

appcues/src/main/java/com/appcues/action/ActionProcessor.kt
```diff
@@ -26,9 +26,21 @@ internal class ActionProcessor(override val scope: Scope) : KoinScopeComponent {
...
+
+    private fun transformQueue(actions: List<ExperienceAction>) = actions.fold(actions) { currentQueue, action ->
+        val indexInCurrent = currentQueue.indexOfFirst { it === action }
+        val transformingAction = action as? ExperienceActionQueueTransforming
+        if (indexInCurrent != -1 && transformingAction != null) {
+            transformingAction.transformQueue(currentQueue, indexInCurrent, appcues)
+        } else {
+            currentQueue
```
iujames, line 43: learned about `fold` - its much like the `reduce` used in Swift - it walks over each item in the list and allows an accumulator function to run.  In our case, this accumulator is either just returning the queue unchanged (base case) or allowing a transforming action to return a modified queue.  Actions are processed in their original order, and if an item is added during this process, it does not have a chance to transform - same as iOS.

appcues/src/main/java/com/appcues/data/model/ExperienceStepFormState.kt
```diff
@@ -20,58 +20,52 @@ internal class ExperienceStepFormState {
     // this can be used by buttons needing to know if they are enabled or not based on required input
     val isFormComplete = mutableStateOf(true)
 
-    fun getValue(primitive: TextInputPrimitive) =
-        getItem(primitive).text
+    fun register(primitive: ExperiencePrimitive) {
```
iujames, line 23: registration of primitives with the formState object is now explicit.  Previously, it was implicitly done on the first call to get or set the value - a side effect. Making it explicit makes it simpler to test, for one, and feels generally a better approach to me.  It only adds a line atop each of the two primitives that participate in form state (textInput and optionSelect) - and simplifies this class quite a bit.

appcues/src/main/java/com/appcues/statemachine/State.kt
```diff
@@ -37,4 +37,26 @@ internal sealed class State {
...
+
+    val currentStepIndex: Int?
+        get() = when (this) {
+            is Idling -> null
+            is BeginningExperience -> null
+            is BeginningStep -> this.flatStepIndex
+            is EndingExperience -> this.flatStepIndex
+            is EndingStep -> this.flatStepIndex
+            is Paused -> this.state.currentStepIndex
+            is RenderingStep -> this.flatStepIndex
```
iujames, line 60: these are made available now so that the submit-form action can get the current state in the state machine and pull out the form state for the current step and act upon it.

appcues/src/main/java/com/appcues/action/appcues/SubmitFormAction.kt
```diff
@@ -0,0 +1,63 @@
...
+) : ExperienceActionQueueTransforming {
+
+    companion object {
+        const val TYPE = "@appcues/submit-form"
+    }
+
+    private val skipValidation = config.getConfigOrDefault("skipValidation", false)
+
+    // validate form and block future actions if needed
+    override fun transformQueue(queue: List<ExperienceAction>, index: Int, appcues: Appcues): List<ExperienceAction> {
```
andretortolano, line 27: checking to see if I understood it.

if the form has validations and if isFormComplete is false we are dropping all other actions as a way to preventing actions from running when form is not yet submitted?

andretortolano (approved):
Nice addition!

PR #243, opened 2022-10-05
Truncate changelog to 500 chars for Google Play API limit

iujames:
Will need to propagate to all the other Android deploy scripts in other repos.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -26,7 +26,7 @@ platform :android do
       commits_count: 5,
       pretty: "- [%as] %s"
     )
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -45,3 +45,7 @@ platform :android do
     )    
   end
 end
+
+def truncate(string, max)
+  string.length > max ? "#{string[0...max]}" : string
+end
```
iujames, line 51: if you know how to use this https://rdoc.info/gems/fastlane/String:truncate rather than making our own, please let me know!

mmaatttt (approved):

PR #244, opened 2022-10-05
Add UI for error states on survey

iujames:
These are the equivalent updates to the iOS changes here https://github.com/appcues/appcues-ios-sdk/pull/248

* a new `errorLabel` property is sent on the `optionSelect` and `textInput` primitives
* `@appcues/submit-form` action sets a flag on `ExperienceStepFormState` to indicate that errors should be shown (validation failed)
* the `optionSelect` and `textInput` use the `errorLabel` values to conditionally tint the text, draw error border, show error text, or tint radio/ check controls as needed
* also slips in the change to use `\n` as the multi-value delimiter instead of `,` for the csv export

https://user-images.githubusercontent.com/19266448/194133752-c33202cb-0f3c-4b46-949c-a82ec1a1ad87.mov


appcues/src/main/java/com/appcues/ui/extensions/TextPrimitiveExt.kt
```diff
@@ -0,0 +1,16 @@
+package com.appcues.ui.extensions
+
+import com.appcues.data.model.ExperiencePrimitive.TextPrimitive
+
+// helper to optionally adapt this Text to use the foreground color of a different Text, as an error state
+// used in form validation
+internal fun TextPrimitive.checkErrorStyle(showError: Boolean, errorLabel: TextPrimitive?): TextPrimitive {
+    val errorTint = if (showError) errorLabel?.style?.foregroundColor else null
```
andretortolano, line 8: not sure if we could save some computing process by transforming this into a derivedStateOf that produces the TextPrimitive.

andretortolano (approved):

appcues/src/main/java/com/appcues/ui/extensions/TextPrimitiveExt.kt
iujames: I think this update will improve things https://github.com/appcues/appcues-android-sdk/pull/244/commits/0578f3c2f321e73f1054ab05de28665ae30a7290

andretortolano (approved):

PR #245, opened 2022-10-06
Fix step_completed event being improperly triggered on backwards navigation

iujames:
these are the same logical changes as were applied on iOS in https://github.com/appcues/appcues-ios-sdk/pull/249, plus the logic adjustment in https://github.com/appcues/appcues-ios-sdk/pull/250

* when processing the transition from rendering step to ending step, only mark the step complete if we are moving forward in the experience.  moving backward (i.e. previous button) will not trigger step_completed analytics
* remove other logic that was being layered on top of the `markComplete` flag in the state (the `isStepCompleted` and `isExperienceCompleted`, as the state is now the true source of whether it should be marked complete
* refactor to process all `EndExperience` actions through a single funnel in the `ExperienceRenderer.dismissCurrentExperience` function - this allows us to consolidate the logic check where we see if the experience is on the last step when it is being ended - and set the `markComplete` flag accordingly

key cases that this should resolve, examples consider a 3-step carousel flow, skippable
* on step 3 and move back to step 2 - no `step_completed` for step 3
* on step 3 and skip/dismiss - does trigger `step_completed` and `experience_completed`, since last step
* on step 2 and skip/dismiss - no `step_completed` on step 2, and `experience_dismissed`, since not on last step

### DIFF
--- a/appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
@@ -197,7 +197,7 @@ internal class StateMachine(
             // we are sitting on the last step if we get here - check if this is a continue with offset 1 (default continue)
             if (action.stepReference is StepOffset && action.stepReference.offset == 1) {
                 // for a continue -> next, on the last step, treat this the same as a close
-                val endExperience = EndExperience(destroyed = false)
+                val endExperience = EndExperience(markComplete = true, destroyed = false)
                 return state.fromRenderingStepToEndingExperience(endExperience, appcuesCoroutineScope) {
                     handleActionInternal(endExperience)
                 }
@@ -212,7 +212,7 @@ internal class StateMachine(
 
     fun stop() {
         appcuesCoroutineScope.launch {
-            handleAction(EndExperience(destroyed = true))
+            handleAction(EndExperience(markComplete = state.isOnLastStep, destroyed = true))
         }
     }
 }
--- a/appcues/src/main/java/com/appcues/statemachine/Transitions.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/Transitions.kt
@@ -71,7 +71,7 @@ internal interface Transitions {
                     // in different groups we want to wait for StartStep action from AppcuesViewModel
                     val response = CompletableDeferred<ResultOf<State, Error>>()
                     Transition(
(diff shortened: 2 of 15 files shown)


appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
```diff
@@ -197,7 +197,7 @@ internal class StateMachine(
             // we are sitting on the last step if we get here - check if this is a continue with offset 1 (default continue)
             if (action.stepReference is StepOffset && action.stepReference.offset == 1) {
                 // for a continue -> next, on the last step, treat this the same as a close
-                val endExperience = EndExperience(destroyed = false)
+                val endExperience = EndExperience(markComplete = true, destroyed = false)
```
iujames, line 200: in this case, we know we are on the last step as we are handling the continue action on last step to close - so always `true`

appcues/src/main/java/com/appcues/statemachine/Transitions.kt
```diff
@@ -71,7 +71,7 @@ internal interface Transitions {
                     // in different groups we want to wait for StartStep action from AppcuesViewModel
                     val response = CompletableDeferred<ResultOf<State, Error>>()
                     Transition(
-                        state = EndingStep(experience, flatStepIndex, true) {
+                        state = EndingStep(experience, flatStepIndex, nextStepIndex > flatStepIndex) {
```
iujames, line 74: key spots here on 74 and 84 fix the core issue of "moving to a previous step should not mark complete"

appcues/src/test/java/com/appcues/ui/ExperienceRendererTest.kt
```diff
@@ -0,0 +1,52 @@
...
+import io.mockk.mockk
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import org.junit.Test
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class ExperienceRendererTest {
+
+    @Test
+    fun `dismissCurrentExperience SHOULD mark complete WHEN current state is on last step`() = runTest {
```
iujames, line 18: this is the new test that validates the override behavior on last step

andretortolano (approved):

PR #246, opened 2022-10-06
✨ Added Spacer primitive block

andretortolano:
Took me longer than expected to pull this one. the main problem I faced here was testing, because a spacer doesnt actually take space on screen unless we modify it to do so.

some important notes I will note to help understand what I had to do to make Spacer work like IOS works

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/primitive/ExperiencePrimitive.kt
+++ b/appcues/src/main/java/com/appcues/ui/primitive/ExperiencePrimitive.kt
@@ -19,6 +19,7 @@ import com.appcues.data.model.ExperiencePrimitive.EmbedHtmlPrimitive
 import com.appcues.data.model.ExperiencePrimitive.HorizontalStackPrimitive
 import com.appcues.data.model.ExperiencePrimitive.ImagePrimitive
 import com.appcues.data.model.ExperiencePrimitive.OptionSelectPrimitive
+import com.appcues.data.model.ExperiencePrimitive.SpacerPrimitive
 import com.appcues.data.model.ExperiencePrimitive.TextInputPrimitive
 import com.appcues.data.model.ExperiencePrimitive.TextPrimitive
 import com.appcues.data.model.ExperiencePrimitive.VerticalStackPrimitive
@@ -69,6 +70,7 @@ internal fun ExperiencePrimitive.Compose(matchParentBox: BoxScope? = null) {
                 is VerticalStackPrimitive -> Compose(modifier)
                 is TextInputPrimitive -> Compose(modifier)
                 is OptionSelectPrimitive -> Compose(modifier)
+                is SpacerPrimitive -> Compose(modifier)
             }
         }
     }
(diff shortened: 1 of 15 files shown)


appcues/src/main/java/com/appcues/ui/primitive/ExperiencePrimitive.kt
```diff
@@ -89,6 +91,7 @@ private fun ExperiencePrimitive.getRole(): Role {
         is TextPrimitive -> Role.Button
         is TextInputPrimitive -> Role.Button
         is OptionSelectPrimitive -> Role.Button
+        is SpacerPrimitive -> Role.Image
```
andretortolano, line 94: I think its Ok to mark as image? probably wont be used anyways. the only possible usecase where this would be clicable is if we actually make it have background or background image.

appcues/src/main/java/com/appcues/ui/primitive/HorizontalStackPrimitive.kt
```diff
@@ -49,15 +51,22 @@ internal fun HorizontalStackPrimitive.Compose(modifier: Modifier) {
...
     parentVerticalAlignment: Alignment.Vertical,
     content: @Composable BoxScope.() -> Unit
 ) {
     Box(
         modifier = Modifier
-            .align(style.getVerticalAlignment(parentVerticalAlignment))
-            .then(if (distribution == ComponentDistribution.EQUAL) Modifier.weight(1f) else Modifier),
-        contentAlignment = style.getBoxAlignment(),
+            .align(primitive.style.getVerticalAlignment(parentVerticalAlignment))
+            .then(
```
andretortolano, line 61: this is where the magic is happening rather than in SpacerPrimitive composition.

we are checking if the ItemBox is rendering a Spacer as a Row item, if so we check if width is null, and then we apply weight (weight can only be applied to a child of a Row (in a RowScope), this is the reason this needs to be done here

appcues/src/main/java/com/appcues/ui/primitive/SpacerPrimitive.kt
```diff
@@ -0,0 +1,11 @@
+package com.appcues.ui.primitive
+
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import com.appcues.data.model.ExperiencePrimitive.SpacerPrimitive
+
+@Composable
+internal fun SpacerPrimitive.Compose(modifier: Modifier) {
+    Spacer(modifier = modifier)
```
andretortolano, line 10: it could be any Composition, or none really. it doesnt actually make any difference but I think this keeps it at least in pair with other primitive types.

appcues/src/main/java/com/appcues/ui/primitive/HorizontalStackPrimitive.kt
iujames: in the case of a vertical spacer, where some height is defined (if we ever allow that in the builder) - this would also work as it would just span available width (maybe full row) and then apply whatever height is defined as part of the SpacerPrimitive composition itself? Is that correct?

iujames (approved):
a unique one, but if it works, it works! thanks for adding to keep parity across platforms

PR #247, opened 2022-10-10
♻️  Reworked spacer to match IOS behavior

andretortolano:
first time around I thought spacer would take width/height, but instead it seems like it uses "spacing" as a way for determining space it takes depending on which scope its in (row or column) this PR changes so it reflects IOS behavior

### DIFF
--- a/appcues/src/main/java/com/appcues/data/mapper/step/primitives/SpacerPrimitiveMapper.kt
+++ b/appcues/src/main/java/com/appcues/data/mapper/step/primitives/SpacerPrimitiveMapper.kt
@@ -1,10 +1,9 @@
 package com.appcues.data.mapper.step.primitives
 
-import com.appcues.data.mapper.styling.mapComponentStyle
 import com.appcues.data.model.ExperiencePrimitive.SpacerPrimitive
 import com.appcues.data.remote.response.step.primitive.PrimitiveResponse.SpacerPrimitiveResponse
 
 internal fun SpacerPrimitiveResponse.mapSpacerPrimitive() = SpacerPrimitive(
     id = id,
-    style = style.mapComponentStyle(),
+    spacing = spacing,
 )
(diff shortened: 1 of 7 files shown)


appcues/src/main/java/com/appcues/data/mapper/step/primitives/SpacerPrimitiveMapper.kt
```diff
@@ -1,10 +1,9 @@
 package com.appcues.data.mapper.step.primitives
 
-import com.appcues.data.mapper.styling.mapComponentStyle
 import com.appcues.data.model.ExperiencePrimitive.SpacerPrimitive
 import com.appcues.data.remote.response.step.primitive.PrimitiveResponse.SpacerPrimitiveResponse
 
 internal fun SpacerPrimitiveResponse.mapSpacerPrimitive() = SpacerPrimitive(
     id = id,
-    style = style.mapComponentStyle(),
+    spacing = spacing,
```
iujames, line 8: I see this matches iOS now https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Data/Models/Experience/ExperienceComponent.swift#L238

Looks like we should update the spec to include the `spacing` property? https://github.com/appcues/appcues-mobile-experience-spec/blob/main/experience-schema.json#L480-L490

appcues/src/main/java/com/appcues/ui/primitive/HorizontalStackPrimitive.kt
```diff
@@ -60,9 +66,8 @@ private fun RowScope.ItemBox(
             .align(primitive.style.getVerticalAlignment(parentVerticalAlignment))
             .then(
                 if (distribution == ComponentDistribution.EQUAL ||
-                    // for when its a spacer with no width defined,
-                    // we will also fill available space
-                    (primitive is SpacerPrimitive && primitive.style.width == null)
+                    // for when its a spacer with no set spacing, it will fill available space
+                    (primitive is SpacerPrimitive && primitive.spacing.eq(0.0))
```
iujames, line 70: just noting, can use `eq` as infix as well
```
(primitive is SpacerPrimitive && primitive.spacing eq 0.0)
```

iujames (approved):
The `StackScope` makes sense to me - applies a similar sort of behavior to what is built into SwiftUI for expanding across the axis of the containing stack layout 👍

appcues/src/main/java/com/appcues/data/mapper/step/primitives/SpacerPrimitiveMapper.kt
andretortolano: yes

PR #248, opened 2022-10-14
✨ Added step_interaction for primitive interactions

andretortolano:
This change mirrors [IOS change](https://github.com/appcues/appcues-ios-sdk/pull/252) on the Android.

I tried to reuse most of the structure made for step_interaction of type Form Submitted. important changes I will note in the PR.


appcues/src/main/java/com/appcues/action/ActionProcessor.kt
```diff
@@ -27,11 +31,25 @@ internal class ActionProcessor(override val scope: Scope) : KoinScopeComponent {
     }
 
     fun process(actions: List<ExperienceAction>) {
-        appcuesCoroutineScope.launch {
-            transformQueue(actions).forEach {
-                actionQueue.send(it)
-            }
+        transformQueue(actions).send()
```
andretortolano, line 34: the process(actions: List) is used by the stateMachine, when its a internal action processing with no UI triggering that.

appcues/src/main/java/com/appcues/action/ActionProcessor.kt
```diff
@@ -27,11 +31,25 @@ internal class ActionProcessor(override val scope: Scope) : KoinScopeComponent {
...
+    }
+
+    // second process with extra parameters is used by the viewModel to communicate
+    // extra relevant information regarding the user interaction that let to process
+    // all the actions.
+    fun process(actions: List<ExperienceAction>, interactionType: InteractionType, viewDescription: String?) {
```
andretortolano, line 40: new process method signature specifically for the viewModel (matching IOS)

appcues/src/main/java/com/appcues/action/ActionProcessor.kt
```diff
@@ -43,4 +61,30 @@ internal class ActionProcessor(override val scope: Scope) : KoinScopeComponent {
             currentQueue
         }
     }
+
+    private fun getStepInteractionAction(
```
andretortolano, line 65: this b asically looks for the "last" valid MetadataSettingsAction implementation so we know we have a step_interaction.

most of the time there will always be at least one, for instance the ContinueAction, LinkAction are both MetadataSettingsAction implementations, but I set this to return a nullable to safe guard in case there isnt any.

appcues/src/main/java/com/appcues/action/appcues/StepInteractionAction.kt
```diff
@@ -0,0 +1,57 @@
...
+import com.appcues.statemachine.StateMachine
+
+internal class StepInteractionAction(
+    override val config: AppcuesConfigMap,
+    private val interaction: StepInteractionData,
+    private val analyticsTracker: AnalyticsTracker,
+    private val stateMachine: StateMachine,
+) : ExperienceAction {
+
+    class StepInteractionData(
```
andretortolano, line 18: this class represents the data analytics will merge to the step_interaction track event.

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
```diff
@@ -25,9 +31,11 @@ internal sealed class ExperienceLifecycleEvent(
         override val experience: Experience,
         val stepIndex: Int,
         val interactionType: InteractionType,
+        val interactionProperties: HashMap<String, Any> = hashMapOf()
     ) : ExperienceLifecycleEvent(experience, AnalyticsEvent.ExperienceStepInteraction) {
+
         enum class InteractionType {
-            FORM_SUBMITTED
+            FORM_SUBMITTED, BUTTON_TAPPED, BUTTON_LONG_PRESSED
```
andretortolano, line 38: I was initially implementing both in a separate spot but I figured that they are similar to FORM_SUBMITTED and it should reside in here (let me know if thats how you imagined that would work)

appcues/src/main/java/com/appcues/data/model/ExperiencePrimitive.kt
```diff
@@ -21,17 +21,29 @@ internal sealed class ExperiencePrimitive(
     open val style: ComponentStyle,
 ) {
 
+    abstract val textDescription: String?
```
andretortolano, line 24: this is all just matching IOS textDescription [implementation](https://github.com/appcues/appcues-ios-sdk/pull/252/files#diff-f378689d5b13c1f1ba657cc55252ad13715ba708cdec91ee38602b69deafe5bcR16)

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
```diff
@@ -38,7 +38,7 @@ internal fun AppcuesComposition(
             LocalViewModel provides viewModel,
             LocalShakeGestureListener provides shakeGestureListener,
             LocalLogcues provides logcues,
-            LocalAppcuesActionDelegate provides AppcuesActions { viewModel.onActions(it) },
+            LocalAppcuesActionDelegate provides AppcuesActionsDelegate(viewModel),
```
andretortolano, line 41: refactored AppcuesActions into AppcuesActionsDelegate for better readability

appcues/src/main/java/com/appcues/action/ActionProcessor.kt
```diff
@@ -27,11 +31,25 @@ internal class ActionProcessor(override val scope: Scope) : KoinScopeComponent {
...
+    // second process with extra parameters is used by the viewModel to communicate
+    // extra relevant information regarding the user interaction that let to process
+    // all the actions.
+    fun process(actions: List<ExperienceAction>, interactionType: InteractionType, viewDescription: String?) {
+        // find step interaction action if there is any
+        val stepInteraction = arrayListOf<ExperienceAction>().apply {
+            getStepInteractionAction(actions, interactionType, viewDescription)?.also { add(it) }
         }
+        // send it before processing all existing actions
+        stepInteraction.send()
```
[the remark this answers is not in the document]
iujames: this is not how [iOS currently handles it](https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift#L85) but I do recall this being discussed in a thread a while back. I assume this is already on @mmaatttt TODO list, but just noting here.

iujames (approved):
Looks good! couple minor notes but good to progress things as you see fit.  Additional two comments:

1. Can you check how it works in the Debugger? That was nice how the interaction data worked out and showed up there on [iOS](https://github.com/appcues/appcues-ios-sdk/pull/252), I _think_ it should work the same on Android, hopefully no extra work needed.
2. Would be nice to have a unit test or two - specifically - sending a set of actions into the ActionProcessor.process function with an interaction type, and confirming that the correct step_interaction event is auto-generated and set to the AnalyticsTracker (via action queue). Test with/without a MetadataSettingsAction.  Test different StepReference variants.

appcues/src/main/java/com/appcues/action/ActionProcessor.kt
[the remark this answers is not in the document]
andretortolano: yes right. https://appcues.slack.com/archives/C031RLGS4F8/p1665686504398749?thread_ts=1665675972.084679&cid=C031RLGS4F8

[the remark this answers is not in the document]
mmaatttt: https://github.com/appcues/appcues-ios-sdk/pull/256

appcues/src/main/java/com/appcues/statemachine/StepReference.kt
```diff
@@ -27,6 +33,8 @@ internal sealed class StepReference {
 
     data class StepGroupPageIndex(val index: Int) : StepReference() {
 
+        override val destination = "#$index"
```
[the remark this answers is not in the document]
andretortolano: StepGroupPageIndex is not really being used by the ContinueAction, and if we were to get the index from the getIndex method we would need to have a reference to the experience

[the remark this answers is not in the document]
iujames: so this is only ever used by the page swiping gesture right? and thus would never actually show up in a step_interaction analytic for a button?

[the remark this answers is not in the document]
andretortolano: precisely. what should we do here? the only way we would use this is if in the future we change this to add a step_interaction for swipe.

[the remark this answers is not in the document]
iujames: Ok, makes sense it _should_ never matter. I think it would be fairly easy to make it work though just to be safe.

* In AppcuesViewModel, where it has UIState.Rendering - add the flatStepIndex in there as another property it tracks.
* in the same file, line 101 where it creates the Rendering state in toRenderingState() - we already have the flatStepIndex, so just pass it though
* line 116 in onPageChanged - we know we're in Rendering state, and we can grab the flatStepIndex and pass it through to StepGroupPageIndex constructor
* now its available as a supplemental property for analytics usage

appcues/src/test/java/com/appcues/action/ActionProcessorTest.kt
```diff
@@ -72,42 +84,46 @@ class ActionProcessorTest : KoinTest {
...
-        with(actionQueue.listQueue.first() as StepInteractionAction) {
-            assertThat(interaction.properties["destination"]).isEqualTo("c")
-            assertThat(interaction.properties["category"]).isEqualTo("test")
-            assertThat(interaction.properties["text"]).isEqualTo("Button 1")
+        val properties = slot<Map<String, Any>>()
+        verify { analyticsTracker.track("appcues:v2:step_interaction", capture(properties), false, true) }
+        with(properties.captured["interactionData"] as Map<*, *>) {
+            assertThat(this["destination"]).isEqualTo("c")
+            assertThat(this["text"]).isEqualTo("Button 1")
+            assertThat(this["category"]).isEqualTo("test")
```
iujames, line 97: i think you can avoid the `capture` and do
```
verify { analyticsTracker.track("appcues:v2:step_interaction", mapOf("destination" to "c", "text" to "Button 1", "category" to "test"), false, true) }
```
but not 100% sure, and this is fine too!

iujames (approved):

appcues/src/test/java/com/appcues/action/ActionProcessorTest.kt
andretortolano: we cant because the properties map is not just a map of the InteractionData, but all the other attributes related to this track event.

PR #249, opened 2022-10-14
♻️  Sanitizing Date values to Double when publishing analytics

andretortolano:
with this change not only we send the form state over to platform but we can also listen for this event using the SDK listener to grab the entire map of properties

### DIFF
--- a/appcues/src/main/java/com/appcues/AnalyticsPublisher.kt
+++ b/appcues/src/main/java/com/appcues/AnalyticsPublisher.kt
@@ -6,6 +6,7 @@ import com.appcues.AnalyticType.IDENTIFY
 import com.appcues.AnalyticType.SCREEN
 import com.appcues.analytics.ActivityRequestBuilder
 import com.appcues.analytics.TrackingData
+import com.appcues.data.model.ExperienceStepFormState
 import com.appcues.data.remote.request.EventRequest
 import java.util.Date
 
@@ -39,6 +40,7 @@ internal class AnalyticsPublisher(
             val value = it.value
             if (key is String && value != null) {
                 sanitizedMap[key] = when (value) {
+                    is ExperienceStepFormState -> value.toHashMap().sanitize()
                     // convert Date types to Double value
                     is Date -> value.time.toDouble()
                     is Map<*, *> -> value.sanitize()
@@ -56,6 +58,7 @@ internal class AnalyticsPublisher(
         filterNotNull().forEach {
             sanitizedList.add(
                 when (it) {
+                    is ExperienceStepFormState -> it.toHashMap().sanitize()
                     // convert Date types to Double value
                     is Date -> it.time.toDouble()
                     is Map<*, *> -> it.sanitize()
--- a/appcues/src/main/java/com/appcues/AnalyticsPublisher.kt
+++ b/appcues/src/main/java/com/appcues/AnalyticsPublisher.kt
@@ -1,0 +1,69 @@
+package com.appcues
+
+import com.appcues.AnalyticType.EVENT
+import com.appcues.AnalyticType.GROUP
+import com.appcues.AnalyticType.IDENTIFY
+import com.appcues.AnalyticType.SCREEN
+import com.appcues.analytics.ActivityRequestBuilder
+import com.appcues.analytics.TrackingData
+import com.appcues.data.remote.request.EventRequest
+import java.util.Date
+
+internal class AnalyticsPublisher(
+    private val storage: Storage
+) {
+
+    fun publish(listener: AnalyticsListener?, data: TrackingData) {
+        if (listener == null) return
+
+        when (data.type) {
+            EVENT -> data.request.events?.forEach {
+                listener.trackedAnalytic(EVENT, it.name, it.attributes.sanitize(), data.isInternal)
+            }
+            IDENTIFY -> listener.trackedAnalytic(IDENTIFY, storage.userId, data.request.profileUpdate, data.isInternal)
+            GROUP -> listener.trackedAnalytic(GROUP, storage.groupId, data.request.groupUpdate, data.isInternal)
+            SCREEN -> data.request.events?.forEach {
+                listener.trackedAnalytic(SCREEN, it.screenTitle(), it.attributes.sanitize(), data.isInternal)
+            }
+        }
+    }
+
+    private fun EventRequest.screenTitle(): String? =
+        attributes[ActivityRequestBuilder.SCREEN_TITLE_ATTRIBUTE] as? String
+
+    private fun Map<*, *>.sanitize(): MutableMap<String, Any> {
+        val sanitizedMap = mutableMapOf<String, Any>()
+
+        forEach {
+            val key = it.key
+            val value = it.value
+            if (key is String && value != null) {
+                sanitizedMap[key] = when (value) {
+                    // convert Date types to Double value
+                    is Date -> value.time.toDouble()
+                    is Map<*, *> -> value.sanitize()
+                    is List<*> -> value.sanitize()
+                    else -> value
+                }
+            }
+        }
+        return sanitizedMap
+    }
+
+    private fun List<*>.sanitize(): List<*> {
+        val sanitizedList = mutableListOf<Any>()
+
+        filterNotNull().forEach {
+            sanitizedList.add(
+                when (it) {
+                    // convert Date types to Double value
+                    is Date -> it.time.toDouble()
+                    is Map<*, *> -> it.sanitize()
+                    is List<*> -> it.sanitize()
+                    else -> it
+                }
+            )
+        }
+        return sanitizedList
+    }
+}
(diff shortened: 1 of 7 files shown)


appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
```diff
@@ -120,7 +120,7 @@ internal sealed class ExperienceLifecycleEvent(
                     hashMapOf(
                         INTERACTION_TYPE_KEY to interactionType.analyticsName(),
                         INTERACTION_DATA_KEY to when (interactionType) {
-                            FORM_SUBMITTED -> step.formState
+                            FORM_SUBMITTED -> step.formState.toAnalyticsMap()
```
iujames, line 123: so, doing it this way will break the debugger.  See https://github.com/appcues/appcues-android-sdk/blob/main/appcues/src/main/java/com/appcues/debugger/DebuggerRecentEventsManager.kt#L208-L214 where it tries to pull out the `ExperienceStepFormState` from the property map.

That is why it was done that way originally at least. There are likely multiple ways we could handle that.  But this is why on iOS it was done _only_ at the listener level, and not further upstream like this. https://github.com/appcues/appcues-ios-sdk/pull/253/files.  We may also need to handle the Date mapping similarly, as the goal was to move all that logic into the core SDK rather than x-plat wrappers.

side track....
@mmaatttt  I notice now the iOS update maps Date to Int64, but in the wrappers we use Double types
[capacitor android](https://github.com/appcues/appcues-capacitor-plugin/blob/main/android/src/main/java/com/appcues/sdk/capacitor/AppcuesPlugin.kt#L60)
[capacitor ios](https://github.com/appcues/appcues-capacitor-plugin/blob/main/ios/Plugin/AppcuesPlugin.swift#L174)
[react native ios](https://github.com/appcues/appcues-react-native-module/blob/main/ios/AppcuesReactNative.swift)

I don't fully recall - but do those not handle Int64/long and we should use double there as well?

iujames: this was the PR where the form state handling was originally added to simplify that debugger usage, for reference https://github.com/appcues/appcues-android-sdk/pull/234

andretortolano: I did not connect the dots before, but this and the other open PR are related in regards to the debugger. what I propose is to map here and fix the debugger to support more generally hashmap enteries for this Interaction Data instead of casting it to ExperienceStepFormState. this way we can make it work for all types of interactions

iujames: I think it could be done, but it will result in having some messier code dealing with nested map indexing in the debugger.  That was the motivation for handling with typed property objects previously.  This was an intermediate commit where I was doing that https://github.com/appcues/appcues-android-sdk/blob/e6988f14241f4f74f1d6048ec4b9eb4ebe73e9bc/appcues/src/main/java/com/appcues/debugger/DebuggerRecentEventsManager.kt#L169-L182 -- it can work, just a big messier with the map indexing, and I was hoping to avoid that.

mmaatttt: Fixing it to be `Double` in iOS: https://github.com/appcues/appcues-ios-sdk/pull/257

appcues/src/main/java/com/appcues/AnalyticsPublisher.kt
```diff
@@ -0,0 +1,69 @@
...
+    private fun EventRequest.screenTitle(): String? =
+        attributes[ActivityRequestBuilder.SCREEN_TITLE_ATTRIBUTE] as? String
+
+    private fun Map<*, *>.sanitize(): MutableMap<String, Any> {
+        val sanitizedMap = mutableMapOf<String, Any>()
+
+        forEach {
+            val key = it.key
+            val value = it.value
+            if (key is String && value != null) {
```
iujames, line 40: This is where we'll still need something to format the formState as a map first, like https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Data/Analytics/AnalyticsBroadcaster.swift#L50-L54

appcues/src/main/java/com/appcues/AnalyticsPublisher.kt
```diff
@@ -0,0 +1,69 @@
...
+                sanitizedMap[key] = when (value) {
+                    // convert Date types to Double value
+                    is Date -> value.time.toDouble()
```
iujames, line 43: Took a look at AutoPropertyDecorator `_lastContentShownAt` for example - it uses `time` (long) for the value, not an actual Date object.  So the x-plat wrappers right now also look for `Long` in the transformation to `Double`.  Maybe we need to change all usages to `Date` instead of `Long` so that this code works, or handle both `Date` and `Long` here? I don't think it would catch all the usages as currently implemented.

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -277,21 +272,6 @@ class Appcues internal constructor(koinScope: Scope) {
...
-                    listener.trackedAnalytic(EVENT, it.name, it.attributes, data.isInternal)
-                }
-                IDENTIFY -> listener.trackedAnalytic(IDENTIFY, storage.userId, data.request.profileUpdate, data.isInternal)
-                GROUP -> listener.trackedAnalytic(GROUP, storage.groupId, data.request.groupUpdate, data.isInternal)
-                SCREEN -> data.request.events?.forEach {
-                    listener.trackedAnalytic(SCREEN, it.screenTitle(), it.attributes, data.isInternal)
-                }
-            }
-        }
+        analyticsPublisher.publish(analyticsListener, data)
```
iujames, line 275: could move this call up to line 90 and eliminate the need for this function entirely?

andretortolano: it might limit our testing capability if we do so

appcues/src/main/java/com/appcues/AnalyticsPublisher.kt
andretortolano: I think I'm ok doing that too. so changing all Long to Double, and also removing all Date().time from the auto properties

iujames: Ok, yeah. I think we should also update the Moshi DateAdapter we have to use Double instead of Long, for consistency.

The `Long` topic makes me think about iOS... @mmaatttt should we also add an Int64->Double translation on the iOS side so that no Int64s can sneak through to the JS wrappers? thinking custom user props for instance.

mmaatttt: Interesting question. I'm going with "no" because any custom props in the JS wrappers will go from JS->SDK as Doubles and not Int64s, and so it'd be doubles from SDK->JS. I can't imagine how someone'd get an Int64 into the SDK from JS.

iujames: good point.  So @andretortolano I think that means that we don't need any Long handling here, as long as all previous usages are updated to Date, and flow through this logic.

andretortolano: so nothing to do with Long. closing this one

appcues/src/main/java/com/appcues/Appcues.kt
iujames: I don't think it would limit it really. I think we'd just need to update the test to call `analyticsPublisher.publish` instead of `appcues.publishTracking`, and it would effectively test the same logic for the formatting of the analytics listener output.

samples/kotlin-android-app/src/main/java/com/appcues/samples/kotlin/ExampleApplication.kt
```diff
@@ -39,6 +42,13 @@ class ExampleApplication : Application() {
 
         appcues = Appcues(this, BuildConfig.APPCUES_ACCOUNT_ID, BuildConfig.APPCUES_APPLICATION_ID) {
             loggingLevel = if (BuildConfig.DEBUG) LoggingLevel.DEBUG else LoggingLevel.INFO
+            analyticsListener = object : AnalyticsListener {
```
andretortolano, line 45: forgot to clean this up.

appcues/src/main/java/com/appcues/data/remote/adapters/ExperienceStepFormStateAdapter.kt
```diff
@@ -4,36 +4,22 @@ import com.appcues.data.model.ExperienceStepFormState
...
-    val fieldType: String,
-    val fieldRequired: Boolean,
-    val label: String,
-    val value: String,
-)
 
 internal class ExperienceStepFormStateAdapter {
+
+    @JsonClass(generateAdapter = true)
+    internal data class StubExperienceStepFormState(
```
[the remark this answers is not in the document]
iujames: maybe you can just use `String` there?

iujames (approved):
looks great! kudos for working through all the various details here

PR #250, opened 2022-10-18
Add support for experiments

iujames:
similar updates to the [iOS changes here](https://github.com/appcues/appcues-ios-sdk/pull/258)

* decode a top level `experiments` mapping on the `QualifyResponse`
* add support for an optional `experiment_id` on the `ExperienceResponse`
* add mapping for an optional `Experiment` object on the `Experience`, when a matching experiment is found for the `experiement_id`
* modify `ExperienceRenderer` to support experiment logic, if there is a matching experiment for an `Experience` being rendered:
    * if group is `control` - block rendering the experience and move on to next priority, if any
    * if not (group `exposed`) - continue rendering the experience as usual
    * in both cases, trigger an `appcues:experiment_entered` analytic for the user, with the group and experiment ID

note: a bit of work was restructuring `ExperienceRenderer` (and tests) to use lazy props instead of constructor injection with Koin, to avoid a circular dependency noted in code comment.

### DIFF
--- a/appcues/src/main/java/com/appcues/AppcuesKoin.kt
+++ b/appcues/src/main/java/com/appcues/AppcuesKoin.kt
@@ -30,14 +30,7 @@ internal object AppcuesKoin : KoinScopePlugin {
         }
         scoped { AppcuesDebuggerManager(context = get(), koinScope = this) }
         scoped { ContextResources(context = get()) }
-        scoped {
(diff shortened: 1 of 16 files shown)


andretortolano (approved):
I'm concerned about the hard coded "control" check to show or not show the experience in ExperienceRenderer.

PR #252, opened 2022-10-20
Update experiments for array representation in JSON

iujames:
Discovered the JSON representation is an array, not a dictionary as previously thought from the ticket - this fixes up.

### DIFF
--- a/appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
+++ b/appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
@@ -40,7 +40,7 @@ internal class ExperienceMapper(
     fun map(
         from: ExperienceResponse,
         priority: ExperiencePriority = NORMAL,
-        experiments: Map<String, ExperimentResponse> = emptyMap()
+        experiments: List<ExperimentResponse>? = null,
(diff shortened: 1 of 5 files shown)


andretortolano (approved):

appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
```diff
@@ -89,10 +89,10 @@ internal class ExperienceMapper(
         return filterIsInstance<PresentingTrait>().firstOrNull() ?: DefaultPresentingTrait(null, scope, context)
     }
 
-    private fun Map<String, ExperimentResponse>.getExperiment(experimentId: String?) =
+    private fun List<ExperimentResponse>.getExperiment(experimentId: String?) =
         experimentId?.let { id ->
-            this[id]?.let { experimentResponse ->
-                Experiment(id, experimentResponse.group)
+            this.firstOrNull { it.experimentId == id }?.let { experimentResponse ->
+                Experiment(experimentResponse.experimentId, experimentResponse.group)
```
iujames, line 95: I assume there can only be one item in the array with a given `experimentId`, but confirming this.  The previous mapping was more clear in this sense.

PR #253, opened 2022-10-20
Add SDK metrics to capture timings around experience rendering

iujames:
The idea here is to use a static registry of metrics by request ID - in `SdkMetrics`.  This is where other spots can track key points in the qualification/rendering lifecycle
* the moment the analytic comes in to the sdk
* when it is sent out in the network request
* when it returns from the network
* when it starts presentation in the UI

Then, when the `experience_started` analytic is trigged on initial render, we can capture and send the new `_sdkMetrics` node in the attributes, for reporting usage.

A few approaches are used to keep this metrics collection from bloating in size at runtime
* only adds entries for things that might potentially trigger qualification - analytics sent in that are "interactive" / processed immediately (i.e. the customer app screens and events).  Does not capture entries for non-qualifying background/internal events like flow analytics
* remove items when a response comes back with no qualified experiences (common case) - since we know we'll never be rendering any thing there
* does not do any tracking for non-traditional experience rendering: ex. deeplinks, show(id), flow triggering another flow
* clean up other items that do have qualification as soon as the render metrics are sent out
* clean up all items on app backgrounding, as a safety net, since no longer current/relevant apples

One interesting aspect of this was hooking in a `MetricsInterceptor`, in Retrofit, to capture the moment of request/response accurately. To accomplish this, we add a temporary request header for `appcues-request-id` on the /qualify POST request.  Then we read this value and use it to capture the timings, and remove it from the request - so it never actually goes out to the API.

Another interesting aspect was reworking the logic around Compose presentation slightly, such that the `presentationComplete` callback does not get invoked now until we have actually fully completed the presentation, including any animated transition. Previously, this was triggered immediately upon state transition in the view model.  Having it also capture the time around the presentation will give a closer comparison in metrics to how iOS presentation works. A new `LaunchOnShowAnimationCompleted` is added to handle this, and the `isContentVisible` is split out from a new `isBackdropVisible` to ensure that we can always accurately detect when the real content is showing, not just the backdrop (as both use similar AnimatedVisibility approaches.

end result gets us data like this example
Screen Shot 2022-10-20 at 12 03 26 PM

### DIFF
--- a/appcues/src/main/java/com/appcues/analytics/SdkMetrics.kt
+++ b/appcues/src/main/java/com/appcues/analytics/SdkMetrics.kt
@@ -10,6 +10,9 @@ internal data class SdkMetrics(
     private var renderStartAt: Date? = null,
 ) {
     companion object {
+
+        const val METRICS_PROPERTY = "_sdkMetrics"
+
         private val metrics = hashMapOf<UUID, SdkMetrics>()
 
         fun clear() {
@@ -57,7 +60,7 @@ internal data class SdkMetrics(
 
                     remove(it)
 
-                    properties["_sdkMetrics"] = mapOf(
+                    properties[METRICS_PROPERTY] = mapOf(
                         "timeBeforeRequest" to timeBeforeRequest,
                         "timeNetwork" to timeNetwork,
                         "timeProcessingResponse" to timeProcessingResponse,
--- a/appcues/src/main/java/com/appcues/analytics/SdkMetrics.kt
+++ b/appcues/src/main/java/com/appcues/analytics/SdkMetrics.kt
@@ -1,0 +1,64 @@
+package com.appcues.analytics
+
+import java.util.Date
+import java.util.UUID
+
+internal data class SdkMetrics(
+    private var trackedAt: Date? = null,
+    private var requestedAt: Date? = null,
+    private var respondedAt: Date? = null,
+) {
+    companion object {
+        private val metrics = hashMapOf<UUID, SdkMetrics>()
+
+        fun clear() {
+            metrics.clear()
+        }
+
+        fun tracked(id: UUID, time: Date?) {
+            time?.let { metrics.getOrPut(id) { SdkMetrics() }.trackedAt = time }
+        }
+
+        fun requested(id: String?, time: Date = Date()) {
+            id?.let { metrics[UUID.fromString(it)]?.requestedAt = time }
+        }
+
+        fun responded(id: String?, time: Date = Date()) {
+            id?.let { metrics[UUID.fromString(it)]?.respondedAt = time }
+        }
+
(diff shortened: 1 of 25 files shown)


iujames:
I need to update the Debugger representation of this data still...

iujames:
> I need to update the Debugger representation of this data still, will update this PR.

Updated:
Screen Shot 2022-10-20 at 2 30 03 PM

appcues/src/main/java/com/appcues/analytics/SdkMetrics.kt
```diff
@@ -0,0 +1,76 @@
+package com.appcues.analytics
+
+import java.util.Date
+import java.util.UUID
+
+internal data class SdkMetrics(
```
andretortolano, line 6: I'm not sure if it would be cleaner if SdkMetric was an internal object, and we had a data called QualificationMetrics with these 4 properties.

andretortolano: that way if in the future we decide to track other things we could create other data classes, and another set of functions to support that metrics

appcues/src/main/java/com/appcues/data/remote/retrofit/RetrofitWrapper.kt
```diff
@@ -48,3 +52,19 @@ internal class RetrofitWrapper(
         }
     }
 }
+
+internal class MetricsInterceptor : Interceptor {
+
+    override fun intercept(chain: Chain): Response {
```
andretortolano, line 58: NICE!

appcues/src/main/java/com/appcues/debugger/DebuggerRecentEventsManager.kt
```diff
@@ -224,6 +241,7 @@ private fun Map<String, Any>.getInteractionData(): Map<String, Any?> {
     this[ExperienceLifecycleEvent.INTERACTION_DATA_KEY].let {
         if (it is ExperienceStepFormState || it == null) return mapOf()
 
+        @Suppress("UNCHECKED_CAST")
```
andretortolano, line 244: I meant to add that in my next PR, thanks

andretortolano (approved):
complex but clear change to make this work. wondering if we could make so its easier to add another types of metrics in the future (but we can always come back later and refactor when needed if you want to keep IOS and Android matching for now)

appcues/src/main/java/com/appcues/analytics/SdkMetrics.kt
iujames: ok, yeah I think that would be a pretty straightforward refactor like you said - if/when we have the need to expand things. I'm inclined to just keep the single use case for now since that's all I'm currently aware of needing.

PR #258, opened 2022-10-25
♻️ Updated Size finder logic to match specs

andretortolano:
Updating the AspecRatio reader logic to fit specs for Side-by-side content.

Before this change aspectRatio was calculated based only on the target primitive, now by knowning [stackScope](https://github.com/appcues/appcues-android-sdk/pull/258/files#diff-ba5fc2a50d116a563e61531b0fc5fb60f4d2538b904a422b789db5cf004a4d56R45) and by each primitive reporting its own size [here](https://github.com/appcues/appcues-android-sdk/pull/258/files#diff-cf90f453a88a858b5ab8c9546f22d74473fff88d732c9875114c33bf846569faR88) we can better define how Constraint.findSize()  works to mirror Web builder and IOS renderings.

updating tests with this change in [this PR](https://github.com/appcues/appcues-mobile-experience-spec/pull/106).

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/composables/CompositionLocals.kt
+++ b/appcues/src/main/java/com/appcues/ui/composables/CompositionLocals.kt
@@ -1,6 +1,8 @@
 package com.appcues.ui.composables
 
+import androidx.compose.runtime.State
 import androidx.compose.runtime.compositionLocalOf
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.staticCompositionLocalOf
 import coil.ImageLoader
 import com.appcues.action.ExperienceAction
@@ -10,7 +12,7 @@ import com.appcues.data.model.ExperienceStepFormState
 import com.appcues.logging.Logcues
 import com.appcues.ui.AppcuesViewModel
 import com.appcues.ui.ShakeGestureListener
-import com.appcues.ui.composables.StackScope.COLUMN
+import com.appcues.ui.composables.StackScope.ColumnStackScope
 import java.util.UUID
 
 // used to register callback for all Actions triggered from primitives
@@ -18,11 +20,16 @@ internal val LocalAppcuesActionDelegate = staticCompositionLocalOf<AppcuesAction
     noLocalProvidedFor("LocalAppcuesActionDelegate")
 }
 
-// class to make it easier to understand the bridge between the
-// LocalAppcuesActionDelegate and the AppcuesViewModel
-internal class AppcuesActionsDelegate(private val viewModel: AppcuesViewModel) {
+// interface to facilitate understanding the bridge
+// between LocalAppcuesActionDelegate and the AppcuesViewModel
+internal interface AppcuesActionsDelegate {
 
-    fun onActions(actions: List<ExperienceAction>, interactionType: InteractionType, viewDescription: String?) {
+    fun onActions(actions: List<ExperienceAction>, interactionType: InteractionType, viewDescription: String?)
+}
+
+internal class DefaultAppcuesActionsDelegate(private val viewModel: AppcuesViewModel) : AppcuesActionsDelegate {
+
+    override fun onActions(actions: List<ExperienceAction>, interactionType: InteractionType, viewDescription: String?) {
         viewModel.onActions(actions, interactionType, viewDescription)
     }
 }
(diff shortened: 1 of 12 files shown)


appcues/src/main/java/com/appcues/ui/composables/CompositionLocals.kt
```diff
@@ -49,10 +56,26 @@ internal val LocalLogcues = staticCompositionLocalOf<Logcues> { noLocalProvidedF
...
 internal val LocalExperienceStepFormStateDelegate = compositionLocalOf { ExperienceStepFormState() }
 
-internal val LocalStackScope = compositionLocalOf { COLUMN }
+internal val LocalStackScope = compositionLocalOf<StackScope> { ColumnStackScope(null, 0) }
+
+internal sealed class StackScope(private val childrenCount: Int) {
+
+    private var childrenSizes = hashMapOf<UUID, Int>()
+    private val _greaterHeight = mutableStateOf(0)
+    val greaterHeight: State<Int> = _greaterHeight
```
iujames, line 65: should this be `greatestSize`? meaning height or width depending on the type of scope? Should it be a Double?

appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
```diff
@@ -66,9 +70,12 @@ internal fun Modifier.outerPrimitiveStyle(
  * eg. ButtonPrimitive, ImagePrimitive
  */
 internal fun Modifier.innerPrimitiveStyle(component: ExperiencePrimitive) = this.then(
-    with(component) {
-        Modifier.padding(style.getPaddings())
-    }
+    when (component) {
+        // image primitive needs to be clipped to bound to avoid
+        // content drawing overflow (side-by-side primitives)
+        is ImagePrimitive -> Modifier.clipToBounds()
```
iujames, line 76: does this affect the use case of a negative margin at all? where content could lay over the top of the item above it, for example? If not, any harm in applying `clipToBounds()` to all primitive types vs. the Image special case here?

appcues/src/main/java/com/appcues/ui/utils/AppcuesAspectRatioModifier.kt
```diff
@@ -78,88 +94,72 @@ internal class AppcuesAspectRatioModifier(
...
 
-    private fun Constraints.tryMinWidth(enforceConstraints: Boolean = true): IntSize {
-        val minWidth = this.minWidth
-        val height = (minWidth / aspectRatio).roundToInt()
-        if (height > 0) {
-            val size = IntSize(minWidth, height)
-            if (!enforceConstraints || isSatisfiedBy(size)) {
-                return size
-            }
+    @Suppress("unused")
```
iujames, line 142: is this an unused private fun?

iujames (approved):
I don't think I really understand the root of the issue and the solution here, but if it tests out, it works. It would be nice to have a little more explanation about what the change is actually doing, if possible.

Are there related shapshot test suite updates? Ideally this one would be paired with a full test run in the snapshot tests to demonstrate the change.

appcues/src/main/java/com/appcues/ui/utils/AppcuesAspectRatioModifier.kt
andretortolano: its just the Constraint. that is unused

appcues/src/main/java/com/appcues/ui/composables/CompositionLocals.kt
andretortolano: it should be greatestSize, but its a Int, its actually pixel from a IntSize object

andretortolano:
I'm not sure how the best way to link this to a test run, but this change is made so all tests pass and are matching IOS/Builder implementation as per spec library.

appcues/src/main/java/com/appcues/ui/utils/AppcuesAspectRatioModifier.kt
iujames: oh, so should the extension function be on something else? granted, not a big deal here it seems.

iujames:
> I'm not sure how the best way to link this to a test run, but this change is made so all tests pass and are matching IOS/Builder implementation as per spec library.

I think the best way is to make a PR in the spec library with those updates and then just mention with a link back and forth - to link the change.

appcues/src/main/java/com/appcues/ui/extensions/ModifierExt.kt
andretortolano: Ithought of leaving it to image since we can save this extra clipping to any other unless we see its necessary. (the border stuff is coming up and might be good to revisit this if we need this applying for other types too

iujames (approved):
It's complex, but looks like everything is testing out great, so let's ship it! Nice work chasing down the details here

PR #260, opened 2022-10-28
✨ Added handling of virtual keyboard for textInput

andretortolano:
Created some logic around the imeAction (Done for the last textInput else Next) and the behavior for those different actions.

Tested on side-by-side content and also accessibility (talkback) on.

here is an example of how this is changing the actual behavior to:

https://user-images.githubusercontent.com/5244805/198672253-66c9c690-851d-4fb9-996f-67f702ecbe48.mp4

### DIFF
--- a/appcues/src/main/java/com/appcues/data/model/ExperienceStepFormState.kt
+++ b/appcues/src/main/java/com/appcues/data/model/ExperienceStepFormState.kt
@@ -18,6 +18,9 @@ internal class ExperienceStepFormState {
     val formItems: Collection<ExperienceStepFormItemState>
         get() = _formItems.values.sortedBy { it.index }
 
+    val lastTextInputItem: ExperienceStepFormItemState?
+        get() = _formItems.values.lastOrNull { it.isTextFocusable }
+
     // this can be used by buttons needing to know if they are enabled or not based on required input
     val isFormComplete: Boolean
         get() = !_formItems.values.any { !it.isComplete }
@@ -82,6 +85,17 @@ internal sealed class ExperienceStepFormItemState(
     val attributeName: String?,
 ) {
 
+    companion object {
+
+        private const val ITEM_TYPE_TEXT_INPUT = "textInput"
+        private const val ITEM_TYPE_OPTION_SELECT = "optionSelect"
+    }
+
+    val isTextFocusable: Boolean
+        get() {
+            return type == ITEM_TYPE_TEXT_INPUT
+        }
+
     val isComplete: Boolean
         get() {
             return when (this) {
@@ -122,7 +136,7 @@ internal sealed class ExperienceStepFormItemState(
     ) : ExperienceStepFormItemState(
         index = index,
         id = primitive.id,
-        type = "textInput",
+        type = ITEM_TYPE_TEXT_INPUT,
         label = primitive.label.text,
         isRequired = primitive.required,
         attributeName = primitive.attributeName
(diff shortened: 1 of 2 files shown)


appcues/src/main/java/com/appcues/data/model/ExperienceStepFormState.kt
```diff
@@ -18,6 +18,9 @@ internal class ExperienceStepFormState {
     val formItems: Collection<ExperienceStepFormItemState>
         get() = _formItems.values.sortedBy { it.index }
 
+    val lastTextInputItem: ExperienceStepFormItemState?
```
andretortolano, line 21: the lastTextInputItem is useful to know whether we should use NEXT or DONE as imeAction [here](https://github.com/appcues/appcues-android-sdk/pull/260/files#diff-09087bb0e71de3f379426edc210622bd24fbcdeff8d12de15e0f27d3eeda986bR162)

appcues/src/main/java/com/appcues/data/model/ExperienceStepFormState.kt
```diff
@@ -82,6 +85,17 @@ internal sealed class ExperienceStepFormItemState(
...
+    companion object {
+
+        private const val ITEM_TYPE_TEXT_INPUT = "textInput"
+        private const val ITEM_TYPE_OPTION_SELECT = "optionSelect"
+    }
+
+    val isTextFocusable: Boolean
+        get() {
+            return type == ITEM_TYPE_TEXT_INPUT
+        }
```
iujames, line 97: Could just make this prop part of the `ExperienceStepFormItemState` constructor and require implementations to pass it in? so `TextInputFormItemState` would pass true.

Just thinking of the future if/when we add another type, and might forget to come back here and update this other line of code.

appcues/src/main/java/com/appcues/data/model/ExperienceStepFormState.kt
```diff
@@ -18,6 +18,9 @@ internal class ExperienceStepFormState {
     val formItems: Collection<ExperienceStepFormItemState>
         get() = _formItems.values.sortedBy { it.index }
 
+    val lastTextInputItem: ExperienceStepFormItemState?
+        get() = _formItems.values.lastOrNull { it.isTextFocusable }
```
iujames, line 22: you might want to use `formItems` (the index sorted list) rather than `_formItems.values` (potentially not sorted)? to be safe

appcues/src/main/java/com/appcues/ui/primitive/TextInputPrimitive.kt
```diff
@@ -129,6 +145,50 @@ internal fun TextInputPrimitive.Compose(modifier: Modifier) {
...
+        cursorColor = cursorColor?.getColor(isDark) ?: MaterialTheme.colors.primary,
+        focusedIndicatorColor = Color.Transparent,
+        unfocusedIndicatorColor = Color.Transparent,
+    )
+
+@Composable
+private fun TextInputPrimitive.getKeyboardOptions(isLastInputItem: State<Boolean>) =
+    KeyboardOptions(
+        keyboardType = mapKeyboardType(dataType),
+        imeAction = if (isLastInputItem.value) ImeAction.Done else ImeAction.Next,
```
iujames, line 162: cool!

iujames (approved):
video looks good! couple small ideas to possibly improve code details.

PR #261, opened 2022-10-28
Support auto downscaling on single character text - avoid emoji ellipsis

iujames:
Lots of experimentation and [discussion in Slack](https://appcues.slack.com/archives/C031RLGS4F8/p1666965887297349) but this is basically the minimal approach to handle the single line / single char case that currently ellipsizes, to address the issue in the ticket.

Important note: for this fix to work, a bugfix in Compose 1.3.0 ellipsis handling has to be included - so this requires upgrading to Compose 1.3.0 (from 1.2.1) - and associated Kotlin 1.7.20 (from 1.7.10), and updating the targetSdk to 33. Android 13 (33) has been out a few months now, but this could impose an update on our customers which they had not yet made.

An example to illustrate - uses 4 side-by-side text elements, all of which have too large of a font size to fit as normally expected.

* "Hello World" - 30pt
* "X" - 150pt
* "Somereallylongwordhere" - 30pt
* "👋🏼" - 200pt  <-- the primary issue in ticket

| After | Before |
| --- | --- |
|  Screen Shot 2022-10-28 at 4 06 38 PM  |  Screen Shot 2022-10-28 at 1 04 34 PM  |

I looked into a broader fix for trying to scale large text down and improve line break behavior a bit, but so far there does not appear to be a way to do this.


iujames:
@andretortolano fyi discovered I should have also updated Accompanist to the [corresponding 0.27.0 version](https://github.com/google/accompanist/releases), to match Compose 1.3.0 - will push that update as well.

andretortolano:
sure, I noticed there are some dependencies not up to date in the project like Material Design, maybe we could update those too? or in another PR, up to you

PR #266, opened 2022-11-04
🐛 Fix bug with virtualkeyboard on scrollable content

andretortolano:
working around a previous issue from delivery-hero (customer) we now have a appcues_activity_layout.xml containing the container for our composition, then we can set fitsSystemWindow to true without introducing unwanted behavior from activity customization. updated UI [tests PR](https://github.com/appcues/appcues-mobile-experience-spec/pull/114)

### DIFF
--- a/appcues/build.gradle
+++ b/appcues/build.gradle
@@ -52,6 +52,7 @@ android {
 
     buildFeatures {
         compose true
+        viewBinding true
     }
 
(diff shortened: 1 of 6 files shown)


iujames (approved):
Looks like a nice simplification and solution for the keyboard handling 👍

PR #267, opened 2022-11-10
Fix embed video expansion to full screen

iujames:
To support a fullscreen button on an embedded video inside of a WebView, we need to implement a custom `AccompanistWebChromeClient` and override the `onShowCustomView` and `onHideCustomView` functions. These calls get invoked to pass a `View` that should be shown/hidden respectively, when the user toggles that button. The general concept for hooking into the Accompanist WebView chrome client was modeled after ideas in https://github.com/google/accompanist/pull/1113. Some of the other implementation details were inspired by https://github.com/akhgupta/WebviewVideo/blob/master/src/com/example/webview/MyActivity.java#L99.

This implementation shows the new custom view by using an empty, hidden `FrameLayout` in the root activity layout, which is then used as a custom view container, conditionally, for this purpose. A CompositionLocal is used to pass this custom chrome client down through the composition, since it needs to be implemented at the host Activity layer for access to the top level layout references.

The `android:configChanges` is an approach to enable rotation to landscape for the fullscreen video, without restarting the AppcuesActivity, and thus stopping the video.

https://user-images.githubusercontent.com/19266448/201178607-7a8ac48d-9baf-4447-9ae3-28335938fcfa.mov

### DIFF
--- a/appcues/src/main/AndroidManifest.xml
+++ b/appcues/src/main/AndroidManifest.xml
@@ -11,6 +11,7 @@
             android:name=".ui.AppcuesActivity"
             android:exported="false"
             android:windowSoftInputMode="adjustResize"
+            android:configChanges="orientation|screenSize|screenLayout"
             android:theme="@style/Appcues.AppcuesActivityTheme" />
     </application>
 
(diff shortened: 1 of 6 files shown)


andretortolano (approved):
Thanks for looking into that, good change to support fullscreen webview videos

PR #268, opened 2022-11-10
✨ Added proguard rules on for consumers of the library

andretortolano:
when minifying the library we need proguard rules to keep files in the data remote package so it can be used to generate the deserializers

### DIFF
--- a/appcues/consumer-rules.pro
+++ b/appcues/consumer-rules.pro
@@ -1,0 +1,1 @@
+-keep class com.appcues.data.remote.response.** { *; }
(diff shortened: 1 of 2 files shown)


iujames:
linking to [sc-44253]

iujames (approved):
nice!

appcues/consumer-rules.pro
```diff
@@ -0,0 +1,5 @@
+-keepclassmembers class com.appcues.data.remote.response.** {
+  <init>(...);
+  <fields>;
+  <methods>;
+}
```
iujames, line 5: maybe can simplify to 
```
-keep class com.appcues.data.remote.response.** { *; }
```
to keep everything in the JSON model package

andretortolano: I will try, thanks

PR #271, opened 2022-11-17
Add support for screen navigation prior to presenting new step container

iujames:
NOTE: targeting new `release/1.2 branch`

Adds a guided walkthrough capability, analogous to iOS work done here https://github.com/appcues/appcues-ios-sdk/pull/283, and supporting the model definition here https://github.com/appcues/appcues-mobile-experience-spec/pull/118.

## Core updates to integrate the pre-step-container-present action concept
* Update to support `"navigate"` as an Action trigger in the JSON
* Update to properly capture Action instances at the `StepContainer` level, in addition to the `Step` level. The previous `mergeActions` behavior would not be appropriate there, as it would result in the container level action actually getting executed N times for N steps.
* Removal of Actions at the `ExperienceResponse` level - this is currently an open TBD/TODO, and unclear when we would support an Action at this level ([same as iOS](https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Data/Models/Experience.swift#L59))
* Updates to `ActionProcessor` - to support the immediate processing of a list of actions, in sequence, and being able to wait on the completion (suspend function). This code path _does not_ flow through the global actionQueue like other interactive actions - comments in the code to clarify.
* When transitioning to BeginningStep - in all cases except the first BeginExperience transition - we now look for any "navigate" actions on the container, and allow those to be passed through into the `PresentContainerEffect` (in Transitions.kt)
* In StateMachine.kt, when processing the `PresentContainerEffect`, we now execute any pre-step actions sequentially (via update in `ActionProcessor` above) and wait on completion (suspend function)

## Support for more app control over link behavior
* Add concept of optional NavigationHandler that can be attached to Appcues instance - this would allow delegating the navigation to Uri to the host application, rather than just triggering a new Intent directly (still the default behavior). Optional for more fine-grained control over the completion of navigation, if an app needs it.
* Update to `LinkAction` - if a navigationHandler is supplied on the Appcues instance, delegate to that (the host app) instead of handling directly, for external or scheme links.

https://user-images.githubusercontent.com/19266448/202529939-59ccd203-405a-42d3-b4ff-7eae240eff84.mov


appcues/src/main/java/com/appcues/action/ActionProcessor.kt
```diff
@@ -68,25 +75,21 @@ internal class ActionProcessor(override val scope: Scope) : KoinScopeComponent {
...
+        return scope.get {
+            parametersOf(
+                StepInteractionData(
+                    interactionType = interactionType,
+                    viewDescription = viewDescription,
+                    category = primaryAction?.category ?: "",
+                    destination = primaryAction?.destination ?: "",
+                )
+            )
+        }
```
iujames, line 93: simplified to always return the `StepInteractionAction`, even if no `MetadataSettingsAction` is found (it should always be...) - with default values in category and destination. Done for [consistency with how iOS is reporting](https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift#L79).

samples/kotlin-android-app/src/main/java/com/appcues/samples/kotlin/ExampleApplication.kt
```diff
@@ -39,6 +42,21 @@ class ExampleApplication : Application() {
...
+                override suspend fun navigateTo(uri: Uri): Boolean {
+                    Intent(Intent.ACTION_VIEW).apply {
+                        data = uri
+                        flags = Intent.FLAG_ACTIVITY_NEW_TASK
+                    }.also {
+                        startActivity(it)
+                    }
+                    return true // navigation successful
+                }
+            }
```
iujames, line 59: seems like this would be generally less necessary on Android than on iOS, where the AndroidManifest intent-filter system is more prescriptive about routing paths to Activities. On iOS, it tends to be more custom logic and chaining of ViewController presentations with animations and completion - which the host app needs to inform the SDK about completion.

I'm imagining perhaps a scenario on Android where a more sophisticated deep linking system is actually doing various Fragment reconfigurations - or even a Compose app recomposing itself in a more advanced way. I want to provide some safety hook to allow the host app to control navigation and inform the SDK when done, so we can safely present the next step _after_ navigation completes.

appcues/src/main/java/com/appcues/NavigationHandler.kt
```diff
@@ -0,0 +1,19 @@
...
+
+    /**
+     * Navigates to the given uri, suspending until navigation has completed. The navigation path
+     * value in `uri` may be a scheme link (deep link) inside of the current application, or an
+     * external web or application link.
+     *
+     * @param uri the uri for the destination to navigate.
+     * @return true if the navigation completed successfully, false if not.
+     */
+    suspend fun navigateTo(uri: Uri): Boolean
```
andretortolano, line 18: Uri is probably the way to go but we could also go for String and have they setup any pattern they want for their screen navigation, but I'm ok using Uri too

appcues/src/main/java/com/appcues/data/remote/response/experience/ExperienceResponse.kt
```diff
@@ -11,7 +10,7 @@ internal data class ExperienceResponse(
     val id: UUID,
     val name: String,
     val theme: String?,
-    val actions: Map<UUID, List<ActionResponse>>?,
+    // val actions: Map<UUID, List<ActionResponse>>?, // not currently using concept of Experience level actions
```
andretortolano, line 13: lets just remove it

appcues/src/main/java/com/appcues/statemachine/Transitions.kt
```diff
@@ -169,14 +170,16 @@ private fun transitionsToBeginningStep(
 ) = if (experience.areStepsFromDifferentGroup(currentStepIndex, nextStepIndex)) {
     // given that steps are from different container, we now get step container index to present
     experience.groupLookup[nextStepIndex]?.let { stepContainerIndex ->
+        val stepGroup = experience.stepContainers[stepContainerIndex]
+        val actions = stepGroup.actions[stepGroup.id]?.filter { it.on == NAVIGATE }?.map { it.experienceAction } ?: emptyList()
```
andretortolano, line 174: to make it more readable we could make this an extension function that has a clear name like stepGroup.getNavigationActions(stepContainerIndex)

andretortolano (approved):

PR #272, opened 2022-11-18
♻️  Update backdrop trait behavior inside composition to enable keyhoole trait

andretortolano:
Before this change, backroptraits were rendering one on top of the other, as totally different compositions, now its all in the same end composition, in a way that you can (if desired) alter the composition coming from the ApplyBackdropTrait chain. an example of this is found on KeyholeTrait, which is not accepting any styling options for now but works as a proof of concept and will be updated later.

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/BackdropDecoratingTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/BackdropDecoratingTrait.kt
@@ -6,5 +6,5 @@ import androidx.compose.runtime.Composable
 interface BackdropDecoratingTrait : ExperienceTrait {
 
     @Composable
-    fun BoxScope.Backdrop()
+    fun BoxScope.BackdropDecorate(content: @Composable BoxScope.() -> Unit)
 }
--- a/appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
+++ b/appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
@@ -10,6 +10,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.platform.LocalDensity
 import com.appcues.logging.Logcues
+import com.appcues.trait.BackdropDecoratingTrait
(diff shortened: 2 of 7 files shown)


appcues/src/main/java/com/appcues/trait/BackdropDecoratingTrait.kt
```diff
@@ -6,5 +6,5 @@ import androidx.compose.runtime.Composable
 interface BackdropDecoratingTrait : ExperienceTrait {
 
     @Composable
-    fun BoxScope.Backdrop()
+    fun BoxScope.BackdropDecorate(content: @Composable BoxScope.() -> Unit)
```
iujames, line 9: no issue at all as its necessary for our project, but this might mark our official break to 2.0 (first public API change)

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
```diff
@@ -133,3 +133,12 @@ private fun BoxScope.ComposeLastRenderingState(state: Rendering) {
...
     }
 }
+
+@Composable
+private fun BoxScope.ApplyBackgroundDecoratingTraits(list: List<BackdropDecoratingTrait>) {
+    // get last trait if its not null compose it and drop last calling it again recursively
+    val item = list.lastOrNull()
+    if (item != null) {
+        with(item) { BackdropDecorate { ApplyBackgroundDecoratingTraits(list.dropLast(1)) } }
+    }
```
iujames, line 143: oh clever ok, so going in reverse like this makes sure that the first trait is actually at the deepest level inside this composition tree right?

iujames (approved):

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
andretortolano: yes this basically means that if:

traits {
  backdrop
  keyhole
}

keyhole is the parent of the composition here. so we can draw the backdrop INSIDE the keyhole trait, giving us the ability to crop everything in there.

PR #273, opened 2022-11-21
♻️  Update PresentingTrait to throw AppcuesTraitException when present() is not possible

andretortolano:
state machine now expects a possible AppcuesTraitException coming from the present() call method, also when it gets it it updates the state to Idling as no experience will be displayed, on top of that @appcues/modal is now a presentingTrait to pair with IOS in preparation to add a tooltip trait coming soon.


appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
```diff
@@ -102,11 +104,25 @@ internal class StateMachine(
                     Failure(sideEffect.error)
                 }
                 is PresentContainerEffect -> {
-                    // kick off UI
-                    sideEffect.experience.stepContainers[sideEffect.containerIndex].presentingTrait.present()
-
-                    // wait on the RenderingStep state to flow in from the UI, return that result
-                    sideEffect.completion.await()
+                    try {
```
andretortolano, line 107: this is the important piece.

try {} around this sideEffect and if we catch AppcuesTraitException we update state to Idling to stop the auto transitions from one state to another to continue. (not sure if thats the right way to do this)

returns the Failure and also notify the errorFlow which ends up being collected by analytics

let me know if I'm missing anything here. there is a unit test for this part

appcues/src/main/java/com/appcues/trait/AppcuesTraitException.kt
```diff
@@ -0,0 +1,3 @@
+package com.appcues.trait
+
+class AppcuesTraitException(message: String) : Exception("AppcuesTraitException: $message")
```
iujames, line 3: wondering if it should be the handler's job to append any prefix onto the message?

appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
```diff
@@ -102,11 +104,25 @@ internal class StateMachine(
...
+                        // kick off UI
+                        sideEffect.experience.stepContainers[sideEffect.containerIndex].presentingTrait.present()
+                        // wait on the RenderingStep state to flow in from the UI, return that result
+                        sideEffect.completion.await()
+                    } catch (exception: AppcuesTraitException) {
+                        // force state machine to move to idling when we get this exception
+                        _state = Idling
```
iujames, line 114: this part we need to think about a bit. I think it needs to not just set to Idling, but actually do a proper transition through the ending of the experience, for analytics purposes. We may need to call back into `handleAction` (like the ContinuationEffect does above) - with an `EndExperience` action, or maybe `Reset`? I need to understand the expected analytics behavior when say, step 3 of 6 fails to present, and the experience is killed.

iujames: on iOS, this type of fatal error gets processed here https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine.swift#L264, which does end up setting the machine right back to idling https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine.swift#L237 -- so maybe this _is_ just fine.

@mmaatttt double checking, do you recall if there was any prior discussion here on errors? Is it expected that no experience_dismissed or any other step events would be triggered, if an experience has a fatal error? thanks

mmaatttt: Yeah, I don't recall how extensively we talked about it, but to me, the idea of a fatal error is you get that and nothing else.

appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
```diff
@@ -87,10 +85,12 @@ internal class ExperienceMapper(
         return filterIsInstance<ContentHolderTrait>().firstOrNull() ?: DefaultContentHolderTrait(null)
     }
 
-    private fun List<ExperienceTrait>.getExperiencePresentingTraitOrDefault(): PresentingTrait {
-        return filterIsInstance<PresentingTrait>().firstOrNull() ?: DefaultPresentingTrait(null, scope, context)
+    private fun List<ExperienceTrait>.getExperiencePresentingTraitOrThrow(): PresentingTrait {
+        return filterIsInstance<PresentingTrait>().firstOrNull() ?: throw AppcuesPresentingTraitNotFound()
     }
 
+    private class AppcuesPresentingTraitNotFound : Exception("Presenting capability trait required")
```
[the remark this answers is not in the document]
andretortolano: I dont think so because this is an Exception during mapper, doesnt actually need to be a traitException, we got some others similar to this one like in [ActionRegestry](https://github.com/appcues/appcues-android-sdk/blob/ca958b7a5379c4b67777b65cd5877392151d9d73/appcues/src/main/java/com/appcues/action/ActionRegistry.kt#L49)

appcues/src/main/java/com/appcues/trait/AppcuesTraitException.kt
andretortolano: yeah thinking back I will just clear it out.

appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
andretortolano: can we leave as is and iterate back if we need to ?

iujames: yes, sorry just wanted to think that one through, but as long as we are consistent, and I didn't see anything in the [event model docs](https://www.notion.so/appcues/Experience-event-model-a05021ffec6c40f091ac01b5e9f94bc4) stating otherwise - then this seems good

appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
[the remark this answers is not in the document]
iujames: ok, makes sense.

iujames (approved):

PR #278, opened 2022-12-01
Add initial support for lossy decoding of Experiences in a qualification response

iujames:
Note: targeting `release/1.2`

There are likely a few ways we could deal with this, open to other ideas!

Experimented a bit with Moshi options around this and landed on a custom adapter factory approach.  The new custom adapter factory is the `LossyExperienceResponseAdapterFactory`.

Approach: 
* Wrap the prior `ExperienceResponse` as one of two derived types of a new `LossyExperienceResponse` sealed class.
* The new adapter factory applies a `LossyExperienceResponseAdapter` when a `List<LossyExperienceResponse>` needs to be deserialized
* When decoding a `List<LossyExperienceResponse>`, loop over each item and try to decode the standard `ExperienceResponse` as usual with the existing generated adapter, normally this works. 
* If not, try to decode a minimal amount of information into a `FailedExperienceResponse` object using a generated adapter. This object would have enough information (i.e. the experience ID) to eventually report flow issues, and also capture error message details to help diagnose the issue
* If all else fails - skip the item
* The `QualificationResponse` then has `List<LossyExperienceResponse>` instead of `List<ExperienceResponse>` - for now, we simply filter it down to the valid `ExperienceResponse` types within and map and process them as normal - minimal changes elsewhere.

In the future in [sc-45196] we could explore actually handling the failures and reporting flow issues in a more meaningful way.

Can test this by just taking one of our normal experience responses and duplicating the experience a few times and going in and removing some required JSON values - for instance, remove the required `type` from a `trait` object - making some of them invalid.  The rest should still deserialize, and will render the top priority item available.

copying @mmaatttt here for thoughts on general approach and whether it aligns with what he had in mind in the original idea.


appcues/src/main/java/com/appcues/data/AppcuesRepository.kt
```diff
@@ -148,13 +150,21 @@ internal class AppcuesRepository(
...
                 logcues.info("qualify request failed, reason: $it")
                 when (it) {
-                    is NetworkError -> successful = false
+                    is NetworkError -> {
+                        when (it.throwable) {
+                            // don't retry on JSON parse failures, server responded, it was just unexpected structure
+                            is JsonDataException -> Unit
+                            else -> successful = false
+                        }
+                    }
```
iujames, line 167: not directly related to this PR feature, but an issue i noticed - we should not retry _every_ `NetworkError`. I noticed that this code path is hit when a valid server response comes back, but we just cannot deserialize it - the `JsonDataException`.  This type of exception should not trigger a re-POST to the server, as the server has already received and processed this request.

appcues/src/main/java/com/appcues/data/AppcuesRepository.kt
```diff
@@ -148,13 +150,21 @@ internal class AppcuesRepository(
 
             qualifyResult.doIfSuccess { response ->
                 val priority: ExperiencePriority = if (response.qualificationReason == "screen_view") LOW else NORMAL
-                experiences += response.experiences.map { experienceMapper.map(it, priority, response.experiments, activity.requestId) }
+                experiences += response.experiences.filterIsInstance<ExperienceResponse>().map {
```
iujames, line 153: `filterIsInstance<ExperienceResponse>` here is where we are just throwing out the experiences that failed to deserialize at this point, during mapping to domain objects.

appcues/src/main/java/com/appcues/data/MoshiConfiguration.kt
```diff
@@ -55,7 +56,8 @@ internal object MoshiConfiguration {
         .add(SerializeNull.JSON_ADAPTER_FACTORY)
         .add(StepContainerAdapter())
         .add(ExperienceStepFormStateAdapter())
-        .add(KotlinJsonAdapterFactory())
+        .add(LossyExperienceResponseAdapter())
+        .addLast(KotlinJsonAdapterFactory())
```
iujames, line 60: changed to `addLast` per documentation note https://github.com/square/moshi#reflection

andretortolano (approved):
reasonable change to me, its well organized the success and failures in json deserialization

iujames:
@andretortolano fyi my last commit makes a slight edit here, generally same approach - but found a way to just completely remove invalid items, rather than needing any `UnknownExperienceResponse` object that had no value.

appcues/src/main/java/com/appcues/data/remote/response/step/StepResponse.kt
```diff
@@ -12,5 +12,5 @@ internal data class StepResponse(
     val content: PrimitiveResponse,
     val traits: List<TraitResponse>,
     val actions: Map<UUID, List<ActionResponse>>,
-    val type: String?,
+    val type: String,
```
iujames, line 15: @andretortolano a head's up that my latest commit adds a unit test for this deserialization logic, and made this minor change to make `type` required here (and in the `StepOrContainerResponse` line 21 above). This is to match iOS and the spec https://github.com/appcues/appcues-mobile-experience-spec/blob/main/experience-schema.json#L64 for consistency.

PR #279, opened 2022-12-01
✨ Added MetadataSettingTrait and implemented it on BackdropTrait

andretortolano:
A coulpe of changes to support MetadataSettingTrait on Android.

* backdropDecoratingTraits and containerDecoratingTraits are no longer part of StepContainer, they belong to each individual step now as in the future we want to be able to change from different values between traits based on the step (implemented here on BackdropTrait)

* Added metadataSettingsTraits to Step in [here](https://github.com/appcues/appcues-android-sdk/pull/279/files#diff-8885f04f0ee7a5cdadf8cf1f93006f25760eed18aec58567b59ed00eca6229cdR15).

* BackdropTrait sets the backgroundColor value on metadata [here](https://github.com/appcues/appcues-android-sdk/pull/279/files#diff-d85c23a152449d98fbf7b381a7ac22447c652cfbe7bd3f100488858e180c5751R42) and this allows us to get that value back during [composition](https://github.com/appcues/appcues-android-sdk/pull/279/files#diff-d85c23a152449d98fbf7b381a7ac22447c652cfbe7bd3f100488858e180c5751R51)

* Changes to AppcuesComposition noted [here](https://github.com/appcues/appcues-android-sdk/pull/279/files#diff-81aaa1294988d87059097807f6f935893e82bade68cc49995ed8a2de3d5b0eb6R83).

UI tests run with success. no changes applied to spec repo.


appcues/src/main/java/com/appcues/data/mapper/step/StepMapper.kt
```diff
@@ -26,11 +29,16 @@ internal class StepMapper(
         val stepTraits = from.traits.map { it to STEP }
         val mergedTraits = stepTraits.mergeTraits(stepContainerTraits)
         val mergedActions = from.actions.mergeActions(stepContainerActions)
+        val mappedTraits = traitsMapper.map(mergedTraits)
 
         return Step(
             id = from.id,
             content = from.content.mapPrimitive(),
-            stepDecoratingTraits = traitsMapper.map(mergedTraits).filterIsInstance(StepDecoratingTrait::class.java),
+            // TODO only one of each type, eg: @appcues/backdrop @appcues/skippable etc
```
andretortolano, line 37: as part of future task we will add some logic here maybe? 

we need to add a check somewhere, maybe in the mappers or in the moshi adapter to scan traits and look for traits of same type in the same "level"

appcues/src/main/java/com/appcues/trait/MetadataSettingTrait.kt
```diff
@@ -0,0 +1,9 @@
+package com.appcues.trait
+
+interface MetadataSettingTrait : ExperienceTrait {
+
+    /**
+     * returns a map of shared values
+     */
+    fun produceMetadata(): Map<String, Any?>
```
andretortolano, line 8: suggestion for better naming?

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
```diff
@@ -91,44 +95,63 @@ private fun BoxScope.ComposeLastRenderingState(state: Rendering) {
...
-                                    ApplyUnderlayStepTraits(this@Box, stepDecoratingPadding)
-
-                                    ComposeStepContent(index, hasFixedHeight, contentPadding, stepDecoratingPadding)
-
-                                    ApplyOverlayStepTraits(this@Box, stepDecoratingPadding)
+        val backdropDecoratingTraits = remember(state.position) { mutableStateOf(steps[state.position].backdropDecoratingTraits) }
+        val containerDecoratingTraits = remember(state.position) { mutableStateOf(steps[state.position].containerDecoratingTraits) }
+        val metadataSettingTraits = remember(state.position) { mutableStateOf(steps[state.position].metadataSettingTraits) }
+        val previousStepMetaData = remember { mutableStateOf(AppcuesStepMetadata()) }
+        val stepMetadata = remember(metadataSettingTraits.value) {
```
andretortolano, line 102: will try to refactor this into a simpler method, but its basically observing changes from metadataSettingsTraits and when that happens its iterating over all those traits and producing a new set of stepMetadata, used by LocalAppcuesStepMetadata below

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
```diff
@@ -91,44 +95,63 @@ private fun BoxScope.ComposeLastRenderingState(state: Rendering) {
...
+                    putAll(it.produceMetadata())
+                }
+            }
+
+            mutableStateOf(AppcuesStepMetadata(previous = previousStepMetaData.value.actual, actual = actual)).also {
+                previousStepMetaData.value = it.value
+            }
+        }
+
+        CompositionLocalProvider(LocalAppcuesStepMetadata provides stepMetadata.value) {
```
andretortolano, line 115: All traits, steps can access the most recent stepMetadata value, that contains the actual and possibly a previous value of that same key

appcues/src/main/java/com/appcues/trait/MetadataSettingTrait.kt
iujames: I think this was discussed in Slack some, and maybe still an open thread. This is where the spec talks about the trait being able to apply/remove, and then Matt noted the other ability for things to listen for changes and coordinate somehow https://github.com/appcues/appcues-mobile-experience-spec/pull/119/files#r1029729395 . That part is not all clear to me yet.

edit: i think i follow the observation part now with the remembers

appcues/src/main/java/com/appcues/trait/appcues/BackdropTrait.kt
```diff
@@ -11,28 +12,57 @@ import androidx.compose.foundation.layout.BoxScope
...
+        val previousColor = remember(metadata) {
+            (metadata.previous[METADATA_BACKGROUND_COLOR] as ComponentColor?).getColor(isDark) ?: Color.Transparent
+        }
+
+        val actualColor = remember(metadata) {
+            (metadata.actual[METADATA_BACKGROUND_COLOR] as ComponentColor?).getColor(isDark) ?: Color.Transparent
+        }
+
+        // whenever metadata changes, set this to false
+        val animateToActual = remember(metadata) { mutableStateOf(false) }
```
iujames, line 59: ok so I think I'm following, this is where the `LocalAppcuesStepMetadata` gets "observed" - via the remembers

iujames (approved):
Looking good - seems like a really slick way to handle the metadata observation using the remember system that Compose already provides - nice!

appcues/src/main/java/com/appcues/trait/MetadataSettingTrait.kt
andretortolano: right now its working like this:

whenever the step change, we change the metadata based on new metadataSettingTraits we have for that step.

when that changes happens, because of the remember it will re-compose its children.

this is working differently from IOS, I'm trying to figure out if I can make that more similar, having dificulties there, if you want to hop in we can chat about it

PR #291, opened 2022-12-21
✨ Added TooltipTrait - first pass

andretortolano:
Added @appcues/tooltip. trait to the list. there is a lot of things to define, like what will be the styling options (so far we are only applying the background related styles). also the tooltip pointer is a triangle manually drawn, I tried spending some time trying to make that better with no success for now. 

I think next steps is to work on other styling properties for the tooltip while we wait for more definitions on behavior.

https://user-images.githubusercontent.com/5244805/210830091-e89dedd5-3b84-41c5-be4d-452b6bbdaeb3.mov

### DIFF
--- a/appcues/build.gradle
+++ b/appcues/build.gradle
@@ -71,6 +71,7 @@ dependencies {
     implementation "androidx.core:core-ktx:1.9.0"
     implementation 'androidx.activity:activity-compose:1.6.1'
     implementation "androidx.browser:browser:1.4.0"
+    implementation "androidx.startup:startup-runtime:1.1.1"
     implementation "androidx.navigation:navigation-compose:2.5.3"
     implementation "androidx.compose.ui:ui:$compose_version"
     implementation "androidx.compose.material:material:$compose_version"
--- a/appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
@@ -1,0 +1,166 @@
+package com.appcues.trait.appcues
+
+import androidx.compose.animation.core.animateDpAsState
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.requiredSizeIn
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.GenericShape
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.draw.rotate
+import androidx.compose.ui.geometry.Rect
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.layout.onSizeChanged
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.IntSize
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.max
+import com.appcues.data.model.AppcuesConfigMap
+import com.appcues.trait.AppcuesTraitAnimatedVisibility
+import com.appcues.trait.ContentWrappingTrait
+import com.appcues.trait.PresentingTrait
+import com.appcues.ui.AppcuesOverlayViewManager
+import com.appcues.ui.composables.LocalAppcuesStepMetadata
+import com.appcues.ui.composables.rememberAppcuesContentVisibility
+import com.appcues.ui.modal.dialogEnterTransition
+import com.appcues.ui.modal.dialogExitTransition
+import com.appcues.ui.utils.AppcuesWindowInfo
+import com.appcues.ui.utils.rememberAppcuesWindowInfo
+import org.koin.core.scope.Scope
+
+internal class TooltipTrait(
+    override val config: AppcuesConfigMap,
+    val scope: Scope,
+) : PresentingTrait, ContentWrappingTrait {
+
+    companion object {
+
+        const val TYPE = "@appcues/tooltip"
+
+        private val SCREEN_PADDING = 16.dp
+        private val MAX_WIDTH_DP = 350.dp
+        private val MAX_HEIGHT_DP = 200.dp
+    }
+
+    @Composable
+    override fun WrapContent(content: @Composable (hasFixedHeight: Boolean, contentPadding: PaddingValues?) -> Unit) {
+        val windowInfo = rememberAppcuesWindowInfo()
+        val metadata = LocalAppcuesStepMetadata.current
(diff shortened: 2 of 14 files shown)


appcues/build.gradle
```diff
@@ -71,6 +71,7 @@ dependencies {
     implementation "androidx.core:core-ktx:1.9.0"
     implementation 'androidx.activity:activity-compose:1.6.1'
     implementation "androidx.browser:browser:1.4.0"
+    implementation "androidx.startup:startup-runtime:1.1.1"
```
andretortolano, line 74: this is again needed so we can have our provider started to register the activity lifecycle monitor at every cold start of the app.

appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
```diff
@@ -0,0 +1,164 @@
...
+
+        const val TYPE = "@appcues/tooltip"
+
+        private val SCREEN_PADDING = 16.dp
+        private val MAX_WIDTH_DP = 350.dp
+        private val MAX_HEIGHT_DP = 200.dp
+    }
+
+    private enum class PointerPosition {
+        TOP, BOTTOM, NONE
```
andretortolano, line 60: none is when a "rect" is not present so we know we should position the tooltip at the bottom (for now) and also that we should not compose the pointer at all

appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    private val style = config.getConfigStyle("style")
+
+    override fun present() {
+        AppcuesOverlayViewManager(scope = scope).addView()
```
andretortolano, line 66: AppcuesOverlayViewManager is what inflates the overlay view where our "AppcuesComposition" resides

appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
```diff
@@ -0,0 +1,164 @@
...
+                                .clip(triangleShape)
+                                .styleBackground(style, isSystemInDarkTheme())
+                        )
+                    }
+                }
+            }
+        }
+    }
+
+    private val triangleShape = GenericShape { size, _ ->
```
andretortolano, line 134: this is the bit that draws the triangle. the best I could do for now

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
```diff
@@ -0,0 +1,93 @@
...
+import com.appcues.databinding.AppcuesOverlayLayoutBinding
+import com.appcues.logging.Logcues
+import com.appcues.monitor.AppcuesActivityMonitor
+import com.appcues.ui.composables.AppcuesComposition
+import com.appcues.ui.primitive.EmbedChromeClient
+import com.appcues.util.getNavigationBarHeight
+import com.appcues.util.getStatusBarHeight
+import org.koin.core.scope.Scope
+
+class AppcuesOverlayViewManager(private val scope: Scope) : DefaultLifecycleObserver {
```
andretortolano, line 21: just like the DebuggerManager, it controls the lifecycle of the entire view in a way

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
```diff
@@ -0,0 +1,93 @@
...
+                    it.removeView(this)
+                }
+            }
+        }
+    }
+
+    private fun Activity.getParentView(): ViewGroup {
+        // if there is any difference in API levels we can handle it here
+        return window.decorView.rootView as ViewGroup
+    }
```
iujames, line 92: I once noticed [this issue](https://app.shortcut.com/appcues/story/44252/android-flutter-review-flutterfragmentactivity-requirement-for-debugger) with Flutter, related to the debugger only, and some instance where it was not compatible with the usage of decorView it seemed. That one does not seem critical, but for this new work, just want to make sure we do whatever we can to make compatible across a wide range of potential host app view configurations.

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    override fun onPause(owner: LifecycleOwner) {
+        super.onPause(owner)
+        viewModel?.onPause()
+        shakeGestureListener?.stop()
+    }
+
+    fun addView() {
+        AppcuesActivityMonitor.activity?.run {
```
iujames, line 42: so now that we have the Activity reference always available, are there other parts of the code we can simplify (in later passes is fine) - like the Debugger, I assume? We could update the public API in Appcues.kt to no longer require passing the Activity reference in on the `debug()` call?

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
iujames: Did a first pass scan but wanted to step back and check my understanding on overall plan.

So this is moving toward inflating views directly over the customer view, inside of the customer app Activity, correct? Rather than the overlay Activity approach we had used with modals.

Are we thinking we'll be able to consolidate and share the approach across all pattern types moving forward? In particular, we've got the new layout .xml file now that is (i think) a copy of the old activity layout.xml. Then, we've got the current logic in the AppcuesActivity and more in the AppcuesViewModel - will we be able to share this VM in some fashion? AppcuesComposition now has the core of all the UI stuff in a nicely modularized spot, so assuming that would get shared.

Is it even technically possible to do tooltips in the overlay fashion like modals? or are we forced into the view inflating in the customer Activity to be able to effectively handle they element targeting coming up, keyholes, etc?

pardon all my questions, just trying to step back and understand the approach here as we lay out the new stuff - and partly jogging my memory from many days doing other stuff. We can have a little zoom on it maybe

andretortolano: it is reasonable to question this at this point. I figured it would be safer to not change modals for now. until we have tooltips all working so we can make tests against one and the other to see if we lose anything by going the view inflation approach instead of the activity.

right now we are sharing the Composition and the view model. I created a new layout file just because of the ID basically. but it should be easy to make it all work later using only this approach

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
andretortolano: correct, we can make that call change if we move foward with this PR

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
andretortolano: I remember this one, and yes it would be a possible issue here probably.

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
iujames: ok, mostly just wanted to understand thought process and future plans - totally agree on doing things incrementally and maybe cycling back to modals after things are progressed here.

I have no issue with continuing to progress things here as you build out the new concepts. It sounds like we have a good handle on the core re-use potential with the major parts there, so I'm happy. The more we can make things implement content rendering "the same" - the better, IMO.

iujames (approved):

PR #292, opened 2023-01-03
Update optionSelect and textInput components to handle label horizontalAlignment

iujames:
note: targeting `release/1.2`

similar to the update on iOS https://github.com/appcues/appcues-ios-sdk/pull/306

This updates the `textInput` and `optionSelect` components to allow their `label` and `errorLabel` primitives within to apply their own `style.horizontalAlignment` properties, and position the text as desired, relative to the survey options/controls.

This allows for the common case of having a `leading` aligned label/error, and `center` aligned ratings content

Screenshot 2023-01-03 at 2 25 50 PM

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/extensions/StyleComponentExt.kt
+++ b/appcues/src/main/java/com/appcues/ui/extensions/StyleComponentExt.kt
@@ -164,12 +164,12 @@ internal fun ComponentStyle.getVerticalAlignment(default: Alignment.Vertical): A
     }
 }
 
-internal fun ComponentStyle.getHorizontalAlignment(): Alignment.Horizontal {
+internal fun ComponentStyle.getHorizontalAlignment(default: Alignment.Horizontal): Alignment.Horizontal {
     return when (horizontalAlignment) {
         ComponentHorizontalAlignment.LEADING -> Alignment.Start
         ComponentHorizontalAlignment.CENTER -> Alignment.CenterHorizontally
         ComponentHorizontalAlignment.TRAILING -> Alignment.End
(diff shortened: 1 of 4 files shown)


appcues/src/main/java/com/appcues/ui/extensions/StyleComponentExt.kt
```diff
@@ -164,12 +164,12 @@ internal fun ComponentStyle.getVerticalAlignment(default: Alignment.Vertical): A
     }
 }
 
-internal fun ComponentStyle.getHorizontalAlignment(): Alignment.Horizontal {
+internal fun ComponentStyle.getHorizontalAlignment(default: Alignment.Horizontal): Alignment.Horizontal {
```
iujames, line 167: changed this because I wanted to be able to pass in `Start` as the default, rather than `CenterHorizontally`, for the label components (which also matches iOS default)

appcues/src/main/java/com/appcues/ui/primitive/TextInputPrimitive.kt
```diff
@@ -130,18 +133,16 @@ internal fun TextInputPrimitive.Compose(modifier: Modifier) {
             shape = RoundedCornerShape(8.dp),
             maxLines = numberOfLines,
             singleLine = numberOfLines == 1,
-            placeholder = placeholder?.let {
-                {
-                    it.Compose()
-                }
-            },
+            placeholder = placeholder?.let { { it.Compose() } },
```
iujames, line 136: this is just playing games to get within 60 lines in the function for Detekt 😅

andretortolano (approved):

PR #294, opened 2023-01-18
♻️ Update Tooltip Trait 

andretortolano:
This pass on tooltip trait comes with some changes.

* AppcuesActivityMonitor and AppcuesOverlay changed to support "subscribing" to the existing activity, I changed to this approach as a way to  re-add our view in case the activity gets destroyed and restored for any reason (orientation changes, layout inspector, etc)
* BackdropKeyHoleTrait now applies, a graphicLayer with 0.99F Alpha to add a new layer that enables BlendMondes to work as [expected](https://stackoverflow.com/questions/73313025/jetpack-compose-canvas-blendmode-not-working-as-expected) 
* Various changes to TooltipTrait files to add support for better positioning the pointer (at the top or bottom of the content). Plus adding in all animation to the animation trait.
* Minor change to AppcuesComposition to add an Spacer fillMaxSize to intercept all events (for now at least) and block user of unwanted interaction.

https://user-images.githubusercontent.com/5244805/220458781-6b57da89-ef36-4545-ae51-de4c29a2289f.mov

Remaining work:

* implement distanceFromTarget option
* implement preferredPosition (top/bottom)
* implement pointer handling for edge positions
* Adjust safe content area for tooltips
* Change Debugger to ensure its always the view on top

### DIFF
--- a/appcues/src/main/java/com/appcues/data/mapper/step/StepMapper.kt
+++ b/appcues/src/main/java/com/appcues/data/mapper/step/StepMapper.kt
@@ -38,7 +38,6 @@ internal class StepMapper(
         return Step(
             id = from.id,
             content = responseContent.mapPrimitive(),
-            // TODO only one of each type, eg: @appcues/backdrop @appcues/skippable etc
             stepDecoratingTraits = mappedTraits.filterIsInstance(StepDecoratingTrait::class.java),
             backdropDecoratingTraits = mappedTraits.filterIsInstance<BackdropDecoratingTrait>(),
             containerDecoratingTraits = mappedTraits.filterIsInstance<ContainerDecoratingTrait>(),
--- a/appcues/src/main/java/com/appcues/monitor/AppcuesActivityMonitor.kt
+++ b/appcues/src/main/java/com/appcues/monitor/AppcuesActivityMonitor.kt
@@ -3,7 +3,6 @@ package com.appcues.monitor
 import android.app.Activity
 import android.app.Application
 import android.os.Bundle
-import androidx.appcompat.app.AppCompatActivity
 import com.appcues.ui.AppcuesActivity
 import java.lang.ref.WeakReference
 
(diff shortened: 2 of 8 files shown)


appcues/src/main/java/com/appcues/data/mapper/step/StepMapper.kt
```diff
@@ -38,7 +38,6 @@ internal class StepMapper(
         return Step(
             id = from.id,
             content = responseContent.mapPrimitive(),
-            // TODO only one of each type, eg: @appcues/backdrop @appcues/skippable etc
```
iujames, line 41: right, we have this handled now in TraitResponseAdapterFactory so I think all good 👍

appcues/src/main/java/com/appcues/monitor/AppcuesActivityMonitor.kt
```diff
@@ -3,23 +3,44 @@ package com.appcues.monitor
...
+    }
+
     fun initialize(application: Application) = apply {
         application.registerActivityLifecycleCallbacks(this)
     }
 
     override fun onActivityResumed(activity: Activity) {
-        if (activity !is AppcuesActivity) {
-            this.activityWeakReference = WeakReference(activity)
+        if (activity !is AppcuesActivity && activity is AppCompatActivity) {
```
iujames, line 37: does the change to `AppCompatActivity` (here and in `AppcuesOverlayViewManager`) introduce any integration limitations with customer apps? Is it no longer possible to use `Activity` here?

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
```diff
@@ -158,8 +164,16 @@ private fun BoxScope.ComposeLastRenderingState(state: Rendering) {
...
 
+@OptIn(ExperimentalComposeUiApi::class)
 @Composable
 private fun BoxScope.ApplyBackgroundDecoratingTraits(list: List<BackdropDecoratingTrait>) {
+    // adds a layer behind all Backdrop decorating traits to prevent touch events to
+    Spacer(
+        modifier = Modifier
+            .fillMaxSize()
+            .pointerInteropFilter { false }
+    )
```
iujames, line 175: wondering if this needs to be pulled out to the caller? this method recursively calls itself on line 180 below, so this would actually add N Spacers each time through the recursion between background decorations, if multiple exist?

iujames (commented):
couple initial questions, no objections here, but I'd like to study TooltipTrait a bit more as there is so much going on and I don't understand it all yet honestly.

appcues/src/main/java/com/appcues/monitor/AppcuesActivityMonitor.kt
andretortolano: good catch, I will change so its similar to what we have for the Debugger

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
andretortolano: totally

iujames (approved):

PR #295, opened 2023-01-18
✨ Added verticalAlignment to modal dialog styles

andretortolano:
`targeting release/1.3.0`

Added verticalAlignment handling for dialogs

image
image


iujames (approved):
Nice and simple. My only question is whether the top/bottom margin are satisfactory for Sam, or how they compare to iOS. Might be worth running these screenshots by him and double checking that.

PR #296, opened 2023-01-20
Add deep link to launch screen capture mode in debugger

iujames:
First phase of work for screen capture tool - support a deep link path to
```
appcues-{app_id}://sdk/capture_screen
```

This will open up the Debugger in a screen capture flavor - changing the FAB icon and click action. Most of this work was handling a new concept of `DebugMode` that now allows the core Debugger code to handle multiple different types of behaviors. Currently two - `Debugger` and `ScreenCapture`. The floating events display will only show in `Debugger` mode.

When new deep links are processed, it will be able to gracefully transition between the two - i.e. if I have the debugger open or expanded, and I launch a link for screen capture, it will close the debugger and put me into capture mode.

The FAB click in screen capture mode does nothing for now, that will be the follow on work - implement the layout and screenshot capture and then confirmation dialog.

https://user-images.githubusercontent.com/19266448/213786748-579f0130-03c8-4589-a901-b2b37512b8d5.mov


appcues/src/main/java/com/appcues/DeepLinkHandler.kt
```diff
@@ -18,36 +20,45 @@ internal class DeepLinkHandler(
...
+            }
+            segments.any() && segments[0] == "debugger" -> {
+                val deepLinkPath = if (segments.count() > 1) segments[1] else null
+                debuggerManager.start(activity, Debugger(deepLinkPath))
+                true
             }
+            segments.any() && segments[0] == "capture_screen" -> {
+                debuggerManager.start(activity, ScreenCapture)
+                true
+            }
```
iujames, line 60: This is the new part in this file. The rest is just some restructuring to split things apart and resolve a Detekt warning about too much function complexity (15 branches) in the previous function when this new path was added.

appcues/src/main/java/com/appcues/debugger/AppcuesDebuggerManager.kt
```diff
@@ -44,14 +45,16 @@ internal class AppcuesDebuggerManager(context: Context, private val koinScope: S
...
+        val changingMode = this.mode != mode
         this.currentActivity = activity
+        this.mode = mode
         coroutineScope.coroutineContext.cancelChildren()
         // it is possible to re-enter start without a stop (deepLinks) - in which case we continue to
         // use the VM we already have - else, make new one here
         val viewModel = debuggerViewModel ?: DebuggerViewModel(koinScope)
         debuggerViewModel = viewModel // and save reference
-        viewModel.onStart(deepLinkPath)
+        viewModel.onStart(mode, changingMode)
```
iujames, line 57: Maybe a better way to handle, but at some level we need to know if we are `changingMode` (i.e. going from `Debugger` to `ScreenCapture` or vice-versa).  This is so we can force reset the mode down in the composition where it controls the FAB display.

appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
```diff
@@ -38,9 +40,9 @@ internal class DebuggerViewModel(
 
     sealed class UIState {
         object Creating : UIState()
-        object Idle : UIState()
+        data class Idle(val mode: DebugMode) : UIState()
         data class Dragging(val dragAmount: Offset) : UIState()
-        data class Expanded(val deepLinkPath: String? = null) : UIState()
+        data class Expanded(val mode: DebugMode) : UIState()
```
iujames, line 45: Imagining that the new capture confirmation dialog may extend from `Expanded` when the mode is `ScreenCapture`, but I'm not totally sure yet - might just be a modal popup on top, haven't gone down that path too far yet.

appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
```diff
@@ -107,36 +109,59 @@ internal class DebuggerViewModel(
...
+        this.mode = mode
+
+        when (mode) {
+            is Debugger -> {
+                val deepLinkPath = mode.deepLinkPath
+                if (deepLinkPath.isNullOrEmpty()) {
+                    // if no path is given, and currently expanded, go back to Idle
+                    if (_uiState.value is Expanded ||
+                        // OR if showing FAB but the mode changed, reset to new Idle mode
+                        (_uiState.value is Idle && reset)) {
```
iujames, line 122: this is a spot where that `reset` is used - going from one type of Idle to another.

appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
```diff
@@ -107,36 +109,59 @@ internal class DebuggerViewModel(
...
-            // the deep link might be from the debugger checking that the configuration is correct
-            // pass along to allow it to check
-            viewModelScope.launch {
-                debuggerStatusManager.checkDeepLinkValidation(deepLinkPath)
+            is ScreenCapture -> {
+                // if in debugger mode, expanded, go to new Idle state
+                if (_uiState.value is Expanded ||
+                    // OR if in Idle (FAB) mode but mode changed, reset to new Idle mode
+                    (_uiState.value is Idle && reset)) {
+                    _uiState.value = Idle(mode)
```
iujames, line 148: and here as well.

I could have made an entirely new state for `ScreenCapture` for the FAB, but it seemed like that would have a lot more cascading changes and this would keep it more isolated.

appcues/src/main/java/com/appcues/debugger/ui/DebuggerFloatingActionButton.kt
```diff
@@ -54,6 +58,9 @@ internal fun BoxScope.DebuggerFloatingActionButton(
 ) {
     val resizeBy = resizeAnimatedState(debuggerState)
 
+    val iconResId = remember(debuggerState.debugMode) { derivedStateOf { debuggerState.debugMode.value.fabIconResourceId() } }
+    val clickLabelResId = remember(debuggerState.debugMode) { derivedStateOf { debuggerState.debugMode.value.fabClickLabelResourceId() } }
```
iujames, line 62: these two `remember`s are the key thing for switching styling based on mode

appcues/src/main/java/com/appcues/debugger/ui/DebuggerFloatingActionEvents.kt
```diff
@@ -74,7 +75,9 @@ internal fun BoxScope.DebuggerFloatingActionEvents(
                 .offset { eventsProperties.positionOffset },
             contentAlignment = Alignment.TopEnd
         ) {
-            val shouldDisplayEvents = debuggerState.isDragging.targetState.not() && debuggerState.isExpanded.targetState.not()
+            val shouldDisplayEvents = debuggerState.debugMode.value is Debugger &&
+                debuggerState.isDragging.targetState.not() &&
+                debuggerState.isExpanded.targetState.not()
```
iujames, line 80: only display floating events in `Debugger` mode

appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
```diff
@@ -151,30 +176,37 @@ internal class DebuggerViewModel(
...
     fun onFabClick() {
-        if (_uiState.value == Idle) {
-            _uiState.value = Expanded()
-        } else if (_uiState.value is Expanded) {
-            _uiState.value = Idle
+        when (val state = _uiState.value) {
+            is Idle -> {
+                when (state.mode) {
+                    is Debugger -> _uiState.value = Expanded(state.mode)
+                    is ScreenCapture -> Unit
```
iujames, line 191: tap action is nothing right now - but this is where we'll carry on and add the capture logic and then the confirmation dialog.

appcues/src/main/java/com/appcues/debugger/DebugMode.kt
```diff
@@ -0,0 +1,6 @@
+package com.appcues.debugger
+
+internal sealed class DebugMode {
+    data class Debugger(val deepLinkPath: String?) : DebugMode()
+    object ScreenCapture : DebugMode()
```
andretortolano, line 5: are we thinking that when we have a "token" from the deep link we will be passing that here?

appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
andretortolano: sure thats fine, I think it would be very slick if we reuse all the animation work we have and just change the content inside the expanded area instead of a totally different dialog popping up, good to keep options open if we want to propose a new UI

appcues/src/main/java/com/appcues/debugger/ui/DebuggerFloatingActionEvents.kt
andretortolano: will there be a way to exit capture mode without dismissing the FAB?

iujames: Sam indicated that the only way to exit is dragging the FAB to the dismiss zone (or deep link back into debugger mode)

appcues/src/main/java/com/appcues/debugger/DebugMode.kt
iujames: I _think_ so. There is a separate ticket for the API integration once that's ready, this is focused on the UI mechanics mostly. But good point to think about there.

andretortolano (approved):

PR #297, opened 2023-01-20
✨ Added spans property to TextPrimitive

andretortolano:
`targeting release/1.3.0`

adds spans as a TextPrimitive property and handles it so whenever we have spans we basically ignore the content sent in "text" and override it with the combined values of all spans.

also when we have only text we map to a single span item so its straight forward way to handle all texts (as span elements).

image

### DIFF
--- a/appcues/src/main/java/com/appcues/data/mapper/step/primitives/TextPrimitiveMapper.kt
+++ b/appcues/src/main/java/com/appcues/data/mapper/step/primitives/TextPrimitiveMapper.kt
@@ -13,7 +13,7 @@ internal fun TextPrimitiveResponse.mapTextPrimitive(): TextPrimitive {
     val processedSpans = when {
         // when text is not null and spans is null or empty we use the raw text as a single span item
         text != null && spans.isNullOrEmpty() -> text.toTextSpanPrimitiveList(style)
-        // else when span is not null we nao to Primitive
+        // else when span is not null we map to Primitive
         spans != null -> spans.toTextSpanPrimitive()
         // else an empty list
         else -> arrayListOf()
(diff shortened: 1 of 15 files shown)


appcues/src/main/java/com/appcues/data/model/ExperiencePrimitive.kt
```diff
@@ -31,11 +31,17 @@ internal sealed class ExperiencePrimitive(
         override val id: UUID,
         override val style: ComponentStyle = ComponentStyle(),
         val text: String,
+        val spans: List<TextSpanPrimitive>,
```
iujames, line 34: can we get rid of `text` entirely at this point?

appcues/src/main/java/com/appcues/data/mapper/step/primitives/TextPrimitiveMapper.kt
```diff
@@ -1,11 +1,41 @@
...
+    val processedSpans = when {
+        // when text is not null and spans is null or empty we use the raw text as a single span item
+        text != null && spans.isNullOrEmpty() -> text.toTextSpanPrimitiveList(style)
+        // else when span is not null we nao to Primitive
+        spans != null -> spans.toTextSpanPrimitive()
+        // else an empty list
+        else -> arrayListOf()
+    }
+
+    // check if this TextPrimitive is valid (contains text in form of spans)
```
iujames, line 22: suggest adding a unit test for this text/span mapping logic

iujames (approved):
looks good! couple additional thoughts below

appcues/src/main/java/com/appcues/data/model/ExperiencePrimitive.kt
andretortolano: it is possible, but its still useful to have either "text" the way it is, or replace  it entirely and use the contentDescription when we need the full text

iujames: is there anywhere that we still use this `text` value after mapping to spans?

iujames (approved):
looks great!

PR #298, opened 2023-02-02
Update handling of sticky content

iujames:
targeting `release/1.3.0`

related to spec change in https://github.com/appcues/appcues-mobile-experience-spec/pull/132

A `Step` now has optional `topStickyContent` and `bottomStickyContent` extracted from the main `content` during the mapping phase. If any stacks with `sticky` are found, they are grouped in order in either top or bottom and wrapped in a vertical stack.

This sticky content then gets composed on top of the step composition (overlay like before) after any other overlay step decoration traits are applied.  The previous `StickyContentTrait` is now removed completely.

https://user-images.githubusercontent.com/19266448/216430878-06b193db-2214-4766-8c17-90cb7a0b3487.mov


appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
```diff
@@ -107,7 +107,10 @@ private fun BoxScope.ComposeLastRenderingState(state: Rendering) {
                         currentPage = state.position,
                         composePage = { index ->
                             with(steps[index]) {
-                                CompositionLocalProvider(LocalAppcuesActions provides actions) {
+                                CompositionLocalProvider(
+                                    LocalAppcuesActions provides actions,
+                                    LocalExperienceStepFormStateDelegate provides formState
```
iujames, line 112: I moved the composition local for the form state here, instead of only within the `ComposeStepContent`. The reason is that it is possible for sticky content to contain form components - really any step decorating traits applied in this spot _might_ have a reason to need access to the form state. I think it was previously too local in the ComposeStepContent.

andretortolano: good call

andretortolano (commented):
This looks clean!

andretortolano (approved):

iujames:
note to self: add tests for step mapper logic

iujames:
Last commit here adds some tests that should cover all the branches of StepMapper now.

Screenshot 2023-02-06 at 9 38 47 AM

PR #299, opened 2023-02-03
Add initial layout capture model and populate from current Activity

iujames:
The only thing using this right now is the FAB tap action from the deep link to /sdk/capture_screen. This matches the model used on the iOS side and defined in the spec for platform. Initially just filling out the Android View hierarchy and device metadata and printing to the log for review.

Next steps will be to handle the screenshot portion, the confirmation dialog, and eventually the integration with the API.

Also want to take another pass and look into the Compose hierarchy capture capabilities, specifically researching how it is done in https://github.com/google/talkback.

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
+++ b/appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
@@ -1,8 +1,10 @@
 package com.appcues.debugger
 
+import android.app.Activity
 import androidx.compose.ui.geometry.Offset
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
+import com.appcues.AppcuesConfig
 import com.appcues.analytics.AnalyticsTracker
 import com.appcues.debugger.DebugMode.Debugger
 import com.appcues.debugger.DebugMode.ScreenCapture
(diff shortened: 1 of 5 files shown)


appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
```diff
@@ -238,3 +252,26 @@ internal class DebuggerViewModel(
...
+            appId = config.applicationId,
+            displayName = displayName,
+            screenshotImageUrl = null,
+            layout = layout,
+            metadata = contextResources.generateCaptureMetadata(),
+            timestamp = timestamp,
+        )
+
+        // TESTING!!
+        capture.prettyPrint()
```
iujames, line 275: this is where we'll eventually trigger the confirmation and then api data flow in some fashion.

thinking possibly this stuff lifts up into `AppcuesDebuggerManager` in some form, rather than in the VM file here.

appcues/src/main/java/com/appcues/debugger/screencapture/ViewExt.kt
```diff
@@ -0,0 +1,77 @@
...
+import com.appcues.R
+import com.appcues.monitor.AppcuesActivityMonitor
+import com.appcues.ui.ElementSelector
+import java.util.Date
+
+internal fun View.screenCaptureDisplayName(timestamp: Date): String {
+    var name = ""
+    val activity = AppcuesActivityMonitor.activity
+    if (activity != null) {
+        name = activity.javaClass.simpleName
```
iujames, line 17: I suppose we could try to find the current Fragment to make this suggested initial name better. I have not looked into this, just used the Activity to start with something simple.

appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
```diff
@@ -188,7 +202,7 @@ internal class DebuggerViewModel(
             is Idle -> {
                 when (state.mode) {
                     is Debugger -> _uiState.value = Expanded(state.mode)
-                    is ScreenCapture -> Unit
+                    is ScreenCapture -> AppcuesActivityMonitor.activity?.captureScreen(config, contextResources)
```
iujames, line 205: note, relying on the new `AppuesActivityMonitor` at this point. If this bubbled up to the `AppcuesDebuggerManager` we'd already have a `currentActivity` reference there.

appcues/src/main/java/com/appcues/debugger/screencapture/ViewExt.kt
andretortolano: you may try to cast to FragmentActivity, look for supportFragment Manager maybe and see if you can gather more information, but I would not rely on that either.

andretortolano (approved):
Looking good, lets iterate over it

PR #300, opened 2023-02-03
♻️  Adjustments to keyhole trait pos-testing

andretortolano:
Another pass on keyhole trait, this one after testing found some inconsistencies between Android and IOS.

important keynotes.

1. enforcing the order of BackdropDecoratingTrait considering a new property called "priority", that is internally defined on trait instance. The reasoning behind is to free the builder from the responsibility of keeping traits in the order that Android needs to make cutout/background work
2. initial pass on blurRadius (will need more work) that will be done right after, but right now if blurRadius is defined, at least the size of the Circle is increased. 

Animation is working, there was a inconsistency between sdks where animation duration

### DIFF
--- a/appcues/src/main/java/com/appcues/data/mapper/trait/TraitsMapper.kt
+++ b/appcues/src/main/java/com/appcues/data/mapper/trait/TraitsMapper.kt
@@ -1,6 +1,7 @@
 package com.appcues.data.mapper.trait
 
 import com.appcues.data.mapper.LeveledTraitResponse
+import com.appcues.trait.BackdropDecoratingTrait
 import com.appcues.trait.ExperienceTrait
 import com.appcues.trait.TraitRegistry
 
@@ -15,6 +16,11 @@ internal class TraitsMapper(
                     add(factory.invoke(it.first.config, it.second))
                 }
             }
+
+            // sort BackdropDecoratingTrait items so that the one with greater priority is at the bottom of the list
+            // during composition ApplyBackgroundDecoratingTraits applies traits in reversed order. this way the trait
+            // with greater is applied first (useful for enforcing order between backdrop and keyhole trait)
+            sortBy { if (it is BackdropDecoratingTrait) it.priority else -1 }
         }
     }
 }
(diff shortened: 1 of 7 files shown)


appcues/src/main/java/com/appcues/trait/BackdropDecoratingTrait.kt
```diff
@@ -5,6 +5,14 @@ import androidx.compose.runtime.Composable
 
 interface BackdropDecoratingTrait : ExperienceTrait {
 
+    companion object {
+        const val BACKDROP_KEYHOLE_PRIORITY = 99
+        const val BACKDROP_BACKGROUND_PRIORITY = 89
+        const val BACKDROP_SKIPPABLE_PRIORITY = 59
+    }
+
+    val priority: Int
```
iujames, line 14: this works, but still wondering if we can do anything to hide this from the public API at all, since Traits are a public extensibility point. For the priority levels, can they be more semantic? like "HIGH", "DEFAULT", "LOW"?  Alternatively, if we just have the one special case - the idea was to just recognize the `BackdropKeyholeTrait` internally as special case and not expose this priority publicly until we found we needed to? I can see how a custom trait might need it as well though and perhaps some public `BackdropClippingTrait` interface could be used to allow those to execute before other backdrop decorations?

iujames (approved):
no issue with merging to keep progress going, I understand why we need some special treatment here. Just still pondering what option is best for the portion touching the public trait API.

appcues/src/main/java/com/appcues/trait/BackdropDecoratingTrait.kt
andretortolano: I could simplify and have this special case for the keyhole for now until we see another scenario we might need priority or another trait there.

appcues/src/main/java/com/appcues/data/mapper/trait/TraitsMapper.kt
```diff
@@ -15,6 +16,9 @@ internal class TraitsMapper(
                     add(factory.invoke(it.first.config, it.second))
                 }
             }
+
+            // ensure BackdropKeyholeTrait is applied first
+            sortBy { it is BackdropKeyholeTrait }
```
andretortolano, line 21: @iujames what do you think? its the easiest way to go about it until we actually need to change our public API

iujames: yeah, I'd be in favor of keeping this an internal special case to start, until we discover we need something more general.

PR #302, opened 2023-02-08
✨ Added property blurRadius for keyhole trait

andretortolano:
Added support for blurRadius to keyhole trait.

Refactored some of backdrop trait too since both use similar code, wanted both traits to look as close as possible.
The current approach is move advanced now using BlenderModes instead of clipping the shape and adding the radius on top of it. 

https://user-images.githubusercontent.com/5244805/217609963-b33a39ab-2c05-499d-beda-7b2fc7704faf.mov


iujames (approved):
looks nice! totally doesn't matter for this PR, but for the demo video - it might help avoid any confusion with others on the team if we update the "background should be green" and "background should be yellow" to whatever is actually is set.

PR #305, opened 2023-02-13
✨ Include Source and Javadoc with library .aar

andretortolano:
`targeting release 1.3.0`
With this change customers can access the source and javadoc through the Android Studio.


iujames (approved):
nice update!

PR #307, opened 2023-02-15
Add support for Authorization header on API requests

iujames:
targeting `release/1.3.1` - unsure exact timing on when we need this, but sooner rather than later.

The code changes here are not actually that large, but a lot of test updates and new tests, as well as some Room DB schema files. I'll note the important parts inline.

In short, you can identify with this optional new property, supplying a token, which then gets added to the Authorization header on any requests from this user.
```
appcues.identify(userID, mapOf("appcues:user_id_signature" to base64token))
```

### DIFF
--- a/appcues/src/main/java/com/appcues/data/local/model/ActivityStorage.kt
+++ b/appcues/src/main/java/com/appcues/data/local/model/ActivityStorage.kt
@@ -11,5 +11,6 @@ internal data class ActivityStorage(
     val accountId: String,
     val userId: String,
     val data: String,
+    val userSignature: String?,
     val created: Date = Date(),
 )
--- a/appcues/src/main/java/com/appcues/data/local/room/AppcuesDatabase.kt
+++ b/appcues/src/main/java/com/appcues/data/local/room/AppcuesDatabase.kt
@@ -1,11 +1,18 @@
 package com.appcues.data.local.room
 
+import androidx.room.AutoMigration
 import androidx.room.Database
 import androidx.room.RoomDatabase
 import androidx.room.TypeConverters
 import com.appcues.data.local.model.ActivityStorage
 
-@Database(entities = [ActivityStorage::class], version = 1, exportSchema = false)
+@Database(
+    entities = [ActivityStorage::class],
+    version = 2,
+    autoMigrations = [
+        AutoMigration(from = 1, to = 2)
+    ]
+)
 @TypeConverters(DateConverter::class)
 internal abstract class AppcuesDatabase : RoomDatabase() {
     abstract fun activityStorageDao(): ActivityStorageDao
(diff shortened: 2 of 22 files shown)


appcues/src/main/java/com/appcues/data/local/model/ActivityStorage.kt
```diff
@@ -11,5 +11,6 @@ internal data class ActivityStorage(
     val accountId: String,
     val userId: String,
     val data: String,
+    val userSignature: String?,
```
iujames, line 14: This changed the Room DB schema, adding a column...

appcues/src/main/java/com/appcues/data/local/room/AppcuesDatabase.kt
```diff
@@ -1,11 +1,18 @@
...
 import com.appcues.data.local.model.ActivityStorage
 
-@Database(entities = [ActivityStorage::class], version = 1, exportSchema = false)
+@Database(
+    entities = [ActivityStorage::class],
+    version = 2,
+    autoMigrations = [
+        AutoMigration(from = 1, to = 2)
+    ]
+)
```
iujames, line 15: thus the new `version` here and the `autoMigration` to handle it (newer Room feature as documented [here](https://developer.android.com/training/data-storage/room/migrating-db-versions))

removing `exportSchema = false` is required, which generates the new files in the repo seen here in 
```
appcues/schemas/com.appcues.data.local.room.AppcuesDatabase/1.json
appcues/schemas/com.appcues.data.local.room.AppcuesDatabase/2.json
```
for db versions 1 and 2, and providing the ability to do the diff and migrate this new column.

appcues/src/main/java/com/appcues/data/remote/retrofit/AppcuesService.kt
```diff
@@ -18,13 +18,15 @@ internal interface AppcuesService {
...
         @Path("user") user: String,
+        @Header("Authorization") authorization: String?,
         @Body activity: RequestBody
     ): ActivityResponse
 
     @POST("v1/accounts/{account}/users/{user}/qualify")
     suspend fun qualify(
         @Path("account") account: String,
         @Path("user") user: String,
+        @Header("Authorization") authorization: String?,
```
iujames, line 29: this is the key spot where the new headers get sent through retrofit

appcues/src/main/java/com/appcues/data/remote/retrofit/RetrofitAppcuesRemoteSource.kt
```diff
@@ -23,38 +23,55 @@ internal class RetrofitAppcuesRemoteSource(
...
+        userSignature: String?,
+        requestId: UUID,
+        activityJson: String,
+    ): ResultOf<QualifyResponse, RemoteError> =
         request {
             appcuesService.qualify(
                 account = accountId,
                 user = userId,
                 requestId = requestId,
+                authorization = userSignature?.let { "Bearer $it" },
```
iujames, line 74: these are the key spots where the `Bearer {token}` gets passed through, when set.

andretortolano (approved):

PR #308, opened 2023-02-15
Use a ConcurrentHashMap for SdkMetrics

iujames:
This is the one-liner change discussed when an issue arose on the iOS side with SdkMetrics and concurrent access https://github.com/appcues/appcues-ios-sdk/pull/336, [discussed here](https://appcues.slack.com/archives/C031RLGS4F8/p1676042652884329?thread_ts=1675968353.962369&cid=C031RLGS4F8)

changes
```kotlin
private val metrics = hashMapOf<UUID, SdkMetrics>()
```

to
```kotlin
private val metrics = ConcurrentHashMap<UUID, SdkMetrics>()
```

However, I was never able to see any real issue here in testing. This includes a unit test to exercise concurrent access in the same fashion that made iOS crash in some cases. This seems like perhaps a safe change without a perceived downside, but unclear if it is really needed. Going ahead and posting the work here for review/discussion.

### DIFF
--- a/appcues/src/main/java/com/appcues/analytics/SdkMetrics.kt
+++ b/appcues/src/main/java/com/appcues/analytics/SdkMetrics.kt
@@ -2,6 +2,7 @@ package com.appcues.analytics
 
 import java.util.Date
 import java.util.UUID
+import java.util.concurrent.ConcurrentHashMap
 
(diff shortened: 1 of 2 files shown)


andretortolano (approved):
preventive patch is always good

PR #309, opened 2023-02-17
♻️  Update skippable trait config options

andretortolano:
Updating skippable trait to support new options defined [here](https://github.com/appcues/appcues-mobile-experience-spec/blob/main/trait-schemas/appcues-skippable-schema.json).

There is a minor update to the current layout for modals too. should not be too different but rather than using a xml file I'm using the same new logic to draw the previous style of skippable button (using drawBehind). this might fail some ui tests because the resource is different. the size is also adjusted to have the touchable area of 48x48.dp on default and 32x32.dp on minimal. 

image Default on light background

image Default on dark background

image Minimal on light background

image Minimal on dark background

image Hidden

Thanks to matt for finding the right combination of colors and BlendMode to make the minimal button appearance to work


appcues/src/main/java/com/appcues/trait/appcues/SkippableTrait.kt
```diff
@@ -34,46 +49,125 @@ internal class SkippableTrait(
...
     override fun BoxScope.BackdropDecorate(content: @Composable BoxScope.() -> Unit) {
-        Spacer(
-            modifier = Modifier
-                .matchParentSize()
-                // add click listener but without any ripple effect, should this happen?
-                .pointerInput(Unit) {
-                    detectTapGestures {
-                        appcuesCoroutineScope.launch {
-                            experienceRenderer.dismissCurrentExperience(markComplete = false, destroyed = false)
+        if (ignoreBackdropTap.not()) {
```
andretortolano, line 106: skip backdrop decorating if ignoreBackdropTap is true

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -2,18 +2,23 @@
   <profile version="1.0">
     <option name="myName" value="Project Default" />
     <inspection_tool class="PreviewAnnotationInFunctionWithParameters" enabled="true" level="ERROR" enabled_by_default="true">
+      <option name="composableFile" value="true" />
```
iujames, line 5: seems like this file should be getting ignored 🤔 https://github.com/appcues/appcues-android-sdk/blob/main/.gitignore#L22

iujames (approved):
slick! nice work

.idea/inspectionProfiles/Project_Default.xml
andretortolano: weird looking into it.

andretortolano: it looks like gitIgnore ignores newly added files, but it keep tracks of existing ones, I will try to push a deletion of this one so it doesnt bother us again.

PR #335, opened 2023-02-17
Do not mark experience completed if dismissed on the final step

iujames:
target `release/1.3.1`

Requested change for parity with web, and to make our Studio analytics dashboards more comprehendible for customers. This is the same change as was applied on iOS here https://github.com/appcues/appcues-ios-sdk/pull/345.


andretortolano (approved):
👌

PR #337, opened 2023-02-21
Remove the properties parameter from anonymous function

iujames:
This is a `main` / 2.0 change (breaking) - we'll no longer support `properties` passed in on `anonymous()` user calls. This is to match the web SDK and comply with data security expectations.

### DIFF
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -18,6 +18,7 @@ import com.appcues.trait.TraitRegistry
 import com.appcues.ui.ExperienceRenderer
 import kotlinx.coroutines.launch
 import org.koin.core.scope.Scope
+import kotlin.DeprecationLevel.ERROR
 
 /**
  * Construct and return an instance of the Appcues SDK.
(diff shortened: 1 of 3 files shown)


andretortolano (approved):
matching web is the way to go then?

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -132,14 +133,22 @@ class Appcues internal constructor(koinScope: Scope) {
...
+    }
+
+    /**
+     * This function has been removed. Calling the anonymous function with a properties parameter
+     * is no longer supported. A call to `anonymous()` with no parameters should be used instead.
+     */
+    @Deprecated("properties are no longer supported for anonymous users.", level = ERROR)
+    @Suppress("UnusedPrivateMember")
+    fun anonymous(properties: Map<String, Any>?) {
+        // removed
```
iujames, line 151: @andretortolano does this match your understanding for how to give a deprecation error with more info to any existing callers? Did [something similar](https://github.com/appcues/appcues-ios-sdk/pull/349/files#diff-864ffdf7e8b525ca234734f7b71d722506418595a7c147f2597998eadee3a246) on iOS side for this change.

andretortolano: I guess we could keep the level as warning since it wont break any code if they still use this one. you could make so there is a link to the correct function with ReplaceWith

@Deprecated("properties are no longer supported for anonymous users.", replaceWith = ReplaceWith("anonymous()"))

https://dev.to/mreichelt/the-hidden-kotlin-gem-you-didn-t-think-you-ll-love-deprecations-with-replacewith-3blo

andretortolano: it seems like with that, Android Studio will suggest the change

iujames: Yeah, I think the idea is that we are explicitly removing functionality here, so the more forceful error is appropriate, versus a deprecation warning or internally dropping the args, which might lead to more subtle unexpected behavior where it is still "allowed" but will stop functioning as it was expected before.  @mmaatttt any thoughts?

mmaatttt: Yeah it should be an compile error not a warning

PR #338, opened 2023-02-22
♻️  Ensure Tooltip view is added just after content view

andretortolano:
very easy fix in the end but I took the time to explore different options, in the end enforcing an index was the way to go for our new tooltip view.

I also tried to reposition in other places our views but it doesnt work as well. and also its probably inconsistent between android versions/devices the structure of the decorView, the only thing we know for sure is that there is a decorView, somewhere in there we have android.R.id.content which contains the content (excluding old actionbar). so one of the takes I had was to get the its parent by calling

( findViewById(android.R.id.content).parent as ViewGroup)

and adding the view into this one. the problem is that here on my emulator this one is actually an ActionBarOverlayView which is internal to the android and position views similarly to linear layout, so even if we set the LayoutParams to Match_parent it wont take space since none is available.

good to keep in mind that whenever we add the debugger view we are just adding it, and now with this change we force this overlayed view to be put at the same index as the debugger view makes so that debugger view moves on the list and remains on top of it.

image


iujames (approved):
nice, seems like this should work well to keep the debugger on top

PR #339, opened 2023-02-23
✨ Add support for property contentDistanceFromTarget

andretortolano:
properly applying contentDistanceFromTarget (from TargetRectangleTrait) on TooltipTraits

* moving some related code around for better reusability 

Most of the changes are basically moving and deleting duplicated code, but will note in comments important parts

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/appcues/StepAnimationTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/StepAnimationTrait.kt
@@ -17,20 +17,27 @@ internal class StepAnimationTrait(
 
         const val TYPE = "@appcues/step-transition-animation"
 
-        const val METADATA_ANIMATION_DURATION = "animationDuration"
-        const val METADATA_ANIMATION_EASING = "animationEasing"
-        const val DEFAULT_ANIMATION = 300
+        const val STEP_TRANSITION_ANIMATION_METADATA = "stepTransitionAnimation"
+        const val DEFAULT_ANIMATION_DURATION = 300
     }
 
-    override fun produceMetadata(): Map<String, Any?> {
-        return hashMapOf(
-            METADATA_ANIMATION_DURATION to config.getConfigInt("duration"),
-            METADATA_ANIMATION_EASING to config.getConfig<String>("easing").toEasing()
-        )
-    }
+    data class StepTransitionAnimationInfo(
+        val duration: Int,
+        val easing: StepAnimationEasing,
+    )
 
     enum class StepAnimationEasing {
         LINEAR, EASE_IN, EASE_OUT, EASE_IN_OUT
+    }
+
+    private val duration = config.getConfigInt("duration") ?: DEFAULT_ANIMATION_DURATION
+    private val easing = config.getConfig<String>("easing").toEasing()
+
+    override fun produceMetadata(): Map<String, Any?> {
+        return StepTransitionAnimationInfo(
+            duration = duration,
+            easing = easing,
+        ).let { hashMapOf(STEP_TRANSITION_ANIMATION_METADATA to it) }
     }
 
     private fun String?.toEasing(): StepAnimationEasing {
--- a/appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
@@ -1,13 +1,7 @@
 package com.appcues.trait.appcues
 
-import androidx.compose.animation.core.EaseIn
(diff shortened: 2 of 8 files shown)


appcues/src/main/java/com/appcues/trait/appcues/StepAnimationTrait.kt
```diff
@@ -17,22 +17,29 @@ internal class StepAnimationTrait(
...
+        const val DEFAULT_ANIMATION_DURATION = 300
     }
 
-    override fun produceMetadata(): Map<String, Any?> {
-        return hashMapOf(
-            METADATA_ANIMATION_DURATION to config.getConfigInt("duration"),
-            METADATA_ANIMATION_EASING to config.getConfig<String>("easing").toEasing()
-        )
-    }
+    data class StepTransitionAnimationInfo(
```
andretortolano, line 24: Transformed duration and easing into this object to maintain it more organized.

appcues/src/main/java/com/appcues/trait/appcues/TargetElementTrait.kt
```diff
@@ -15,7 +15,7 @@ internal class TargetElementTrait(
 
         const val TYPE = "@appcues/target-element"
 
-        const val METADATA_TARGET_RECT = "targetRectangle"
+        const val METADATA_TARGET_RECT = "targetElement"
```
andretortolano, line 18: most for understanding, this is not being used right now but I noticed it was referring to targetRectangle, when we get to targetElement we might want to have both in different keys, for now will leave as this to avoid any confusion

appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
```diff
@@ -106,20 +96,23 @@ internal class TooltipTrait(
...
         val tooltipSettings =
             remember(targetRect, containerDimens.value) {
                 getTooltipSettings(
-                    density,
-                    getPointerPosition(windowInfo, targetRect),
-                    pointerBaseDp,
-                    pointerLengthDp,
+                    density = density,
+                    position = getPointerPosition(windowInfo, targetRect),
+                    distance = distance,
```
andretortolano, line 113: added distance as part of tooltipSettings

appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
```diff
@@ -197,9 +190,9 @@ internal class TooltipTrait(
 
         val tooltipPaddingTop = animateDpAsState(
             targetValue = when (pointerSettings.pointerPosition) {
-                TOP, TOP_START, TOP_END -> max((targetRect?.bottom?.dp ?: 0.dp) - SCREEN_VERTICAL_PADDING, 0.dp)
+                TOP, TOP_START, TOP_END -> max((targetRect?.bottom?.dp ?: 0.dp) - SCREEN_VERTICAL_PADDING + pointerSettings.distance, 0.dp)
```
andretortolano, line 193: applying here when its positioned to the top

appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
```diff
@@ -197,9 +190,9 @@ internal class TooltipTrait(
...
                 BOTTOM, BOTTOM_START, BOTTOM_END -> max(
-                    (targetRect?.top?.dp ?: 0.dp) - containerDimens.heightDp - SCREEN_VERTICAL_PADDING, 0.dp
+                    (targetRect?.top?.dp ?: 0.dp) - containerDimens.heightDp - SCREEN_VERTICAL_PADDING - pointerSettings.distance, 0.dp
```
andretortolano, line 195: applying here when its positioned to the bottom

appcues/src/main/java/com/appcues/trait/appcues/TargetElementTrait.kt
iujames: I think we may always want "targetRectangle"? shared between anything that consumes the target-element and the target-rectangle trait. Both of those will resolve to an underlying rectangle that other things (like keyhole) will use. I don't think we'll have a separate "targetElement" metdata - just it's underlying rectangle.

iujames (approved):

appcues/src/main/java/com/appcues/trait/appcues/TargetElementTrait.kt
andretortolano: makes sense, unless we want to somehow combine both to make something very flexible. will keep an eye for that when we get to that.

PR #340, opened 2023-02-23
Add screen capture confirmation dialog with current activity screenshot

iujames:
First round of the screen capture UI work. When the capture button is pressed, it will create a Bitmap of the current screen (minus our Debugger view) and then launch this new Expanded mode of the debugger. 

When the UIState is Expanded for the ScreenCapture mode, it will show the confirmation dialog using the Capture model. When a valid name is input and the "Send to builder" button is pressed, it will call back to the VM to continue processing - which is where this PR leaves off. Follow on work will handle the new API calls for saving the image and screen.

This work required some refactoring of how the debugger composition works, I'll try to comment on those inline.

Screenshot 2023-02-23 at 3 45 56 PM


appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
```diff
@@ -185,25 +187,14 @@ internal class DebuggerViewModel(
...
-    }
-
-    fun onDragEnd() {
-        _uiState.value = Idle(mode)
-    }
-
-    fun onDismiss() {
-        _uiState.value = Dismissing
+    fun transition(state: UIState) {
+        _uiState.value = state
```
iujames, line 191: Detekt was not happy about how many functions this class had. A lot of them were simply pass-through functions to update the `_uiState`. So this new `transition` function replaces some of those and allows updated callsites to just initiate that transition as needed -- for example `onDismiss` below.

appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
```diff
@@ -212,16 +203,19 @@ internal class DebuggerViewModel(
         }
     }
 
-    fun onBackPress() {
+    fun closeExpandedView() {
```
iujames, line 206: there were also multiple functions doing the same thing here - returning from `Expanded` state and going to `Idle` - so now they share this one function.

appcues/src/main/java/com/appcues/debugger/screencapture/Capture.kt
```diff
@@ -21,6 +23,11 @@ internal data class Capture(
     val metadata: Metadata,
     val timestamp: Date,
 ) {
+    // this is a lateinit var instead of a constructor arg so we can ignore it in JSON
+    // serialization and a default value is not needed for the Moshi generated adapter
+    @Json(ignore = true)
+    lateinit var screenshot: Bitmap
```
iujames, line 29: later, this will be used to upload the image data, get back the image URL and then pass along the link in the screen API.

appcues/src/main/java/com/appcues/debugger/ui/CaptureConfirmDialog.kt
```diff
@@ -0,0 +1,220 @@
...
+import androidx.compose.ui.unit.sp
+import androidx.compose.ui.window.Dialog
+import coil.compose.rememberAsyncImagePainter
+import com.appcues.R.string
+import com.appcues.debugger.DebuggerViewModel
+import com.appcues.debugger.screencapture.Capture
+import com.appcues.ui.theme.AppcuesColors
+
+@Composable
+internal fun CaptureConfirmDialog(capture: Capture, debuggerState: MutableDebuggerState, debuggerViewModel: DebuggerViewModel) {
```
iujames, line 54: This file is all the new UI for this `Dialog` composition - open to any and all improvements! I'm still learning a lot about Compose.

appcues/src/main/java/com/appcues/debugger/ui/DebuggerComposition.kt
```diff
@@ -70,6 +71,15 @@ internal fun DebuggerComposition(viewModel: DebuggerViewModel, onDismiss: () ->
             debuggerState = debuggerState,
             debuggerViewModel = viewModel,
         )
+
+        val capture = debuggerState.screenCapture.value
+        if (capture != null) {
```
iujames, line 76: is this the best way to conditionally show this dialog based on state?

appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
```diff
@@ -247,31 +241,42 @@ internal class DebuggerViewModel(
...
-    layout?.let {
-        val capture = Capture(
-            appId = config.applicationId,
-            displayName = displayName,
-            screenshotImageUrl = null,
-            layout = layout,
-            metadata = contextResources.generateCaptureMetadata(),
-            timestamp = timestamp,
-        )
+): Capture? =
```
andretortolano, line 251: this function could be on a separate file like "ScreenCaptor"  and it could be a dependency of DebuggerViewModel, what do you think?

appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
```diff
@@ -247,31 +241,42 @@ internal class DebuggerViewModel(
...
+    window.decorView.rootView.let {
+        // hide the debugger view for screen capture, if present
+        val debuggerView = it.findViewById<View>(R.id.appcues_debugger_view)
+        debuggerView?.let { view ->
```
andretortolano, line 255: we could in the same class have methods like prepare() and restore() for things that we need to do before and after

appcues/src/main/java/com/appcues/debugger/screencapture/ViewExt.kt
```diff
@@ -75,3 +77,11 @@ internal fun View.asCaptureView(): Capture.View? {
         children = children,
     )
 }
+
+internal fun View.screenshot() =
```
andretortolano, line 81: we could move these specific methods inside that new class that would only work on capturing the screen.

appcues/src/main/java/com/appcues/debugger/ui/CaptureConfirmDialog.kt
```diff
@@ -0,0 +1,220 @@
...
+                .clickable(
+                    onClick = { debuggerViewModel.closeExpandedView() },
+                    role = Role.Button,
+                    interactionSource = remember { MutableInteractionSource() },
+                    indication = rememberRipple(bounded = false, 24.dp),
+                    onClickLabel = stringResource(id = string.appcues_screen_capture_dismiss)
+                )
+                .drawBehind {
+                    xShapePath(Color.Black, 16.dp)
+                        .also { drawPath(path = it, Color.Black) }
```
andretortolano, line 100: wondering is Color.Black will work for Dark mode

andretortolano: also this looks as though this was copied from the skippable trait, I guess we could  define the xShapePath somewhere we can share in other places.

appcues/src/main/java/com/appcues/debugger/ui/CaptureConfirmDialog.kt
```diff
@@ -0,0 +1,220 @@
...
+                    )
+            )
+        }
+    }
+}
+
+private fun Modifier.conditionalClickable(enabled: Boolean, onClick: () -> Unit) = composed {
+    this.then(
+        if (enabled) {
+            Modifier.clickable {
```
andretortolano, line 168: lambda nitpicking Modifier.clickable(onClick = onClick)

appcues/src/main/java/com/appcues/debugger/ui/CaptureConfirmDialog.kt
```diff
@@ -0,0 +1,220 @@
...
+                onClick()
+            }
+        } else {
+            Modifier.alpha(ContentAlpha.disabled)
+        }
+    )
+}
+
+@Composable
+private fun CaptureButton(text: String, textColor: Color, modifier: Modifier) {
```
andretortolano, line 178: Modifier is usually the first parameter, and also includes default value == Modifier.
```

CaptureButton(
    modifier: Modifier = Modifier
    .
    .
    .
)
```

appcues/src/main/java/com/appcues/debugger/ui/DebuggerComposition.kt
andretortolano: found an interesting guide from androiddevelopers that showcase the showing of a modal.

https://developer.android.com/codelabs/basic-android-kotlin-compose-viewmodel-and-state#8

it seems like we would use a StateFlow and collectAsState to check its value. the way you have now is probably working tho

appcues/src/main/java/com/appcues/ui/theme/AppcuesColors.kt
```diff
@@ -6,7 +6,7 @@ internal object AppcuesColors {
 
     val WhisperBlue = Color(color = 0xFFE7ECF3)
     val HadfieldBlue = Color(color = 0xFF0D7EF9)
-    val ShadyNeonBlue = Color(color = 0xFF5C5CFF)
+    val ShadyNeonBlue = Color(color = 0xFF5C5CFF) // AKA Blurple
```
andretortolano, line 9: Im ok with renaming if you would like

andretortolano (approved):
Good, I think we can be more intentional about a class that does all the screen capturing work

appcues/src/main/java/com/appcues/debugger/ui/DebuggerComposition.kt
iujames: Interesting, I think this has some similarities already to what we are doing in the DebuggerComposition LaunchedUIStateEffect, where the VM UIState has `collectAsState` and then this dialog shows on `Expanded(ScreenCapture)` state (causing this recomposition to get the `debuggerState.screenCapture` here and check if not null)

appcues/src/main/java/com/appcues/debugger/ui/CaptureConfirmDialog.kt
iujames: yes on both points - I'll see if I can find a shared spot for that xShapePath, good call!

Dark mode - I did check this, in short, it's not handled at all throughout the debugger currently. I figured I would leave that be and we'd need to do a follow up pass to handle dark mode if desired.

appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
iujames: Yeah I like the idea of handling screen capture responsibilities in it's own class - I'm going to take a refactor pass at that. Thanks for the suggestions!

appcues/src/main/java/com/appcues/debugger/ui/CaptureConfirmDialog.kt
andretortolano: Dark mode: good point, I will create a task for it

appcues/src/main/java/com/appcues/debugger/screencapture/ScreenCaptureProcessor.kt
```diff
@@ -0,0 +1,147 @@
...
+import com.appcues.AppcuesConfig
+import com.appcues.R
+import com.appcues.monitor.AppcuesActivityMonitor
+import com.appcues.ui.ElementSelector
+import com.appcues.util.ContextResources
+import com.appcues.util.ResultOf
+import com.appcues.util.ResultOf.Success
+import java.util.Date
+
+internal class ScreenCaptureProcessor(
```
iujames, line 19: @andretortolano I've moved the details around generating and (soon to be) saving a screen capture into this `ScreenCaptureProcessor` to consolidate, if you'd like to take another look.

PR #341, opened 2023-02-23
Merge 1.3.1 to main

iujames:
@andretortolano I'd like to catch up main with the latest updates in the `release/1.3.1` branch - specifically the auth header additions, prior to working on the screen capture API updates - which will also have authed requests.


andretortolano (approved):
Good call, almost forgot about this one, we should always have a working PR for release branches out there, so we dont forget about those!

PR #342, opened 2023-02-24
Add support for token param on capture_screen deep link

iujames:
stacks on #340 

Small prep work for upcoming API integration - allowing a `token` query parameter on the deep link
```
appcues-{app_id}://sdk/capture_screen?token={jwt}
```

This will come from the builder QR code flow, and allow authentication for screen capture requests coming up. The updates here capture the token and pass it through to spots it will be needed.

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/screencapture/ScreenCaptureProcessor.kt
+++ b/appcues/src/main/java/com/appcues/debugger/screencapture/ScreenCaptureProcessor.kt
@@ -48,7 +48,8 @@ internal class ScreenCaptureProcessor(
         }
     }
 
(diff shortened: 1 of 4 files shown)


appcues/src/main/java/com/appcues/debugger/screencapture/ScreenCaptureProcessor.kt
```diff
@@ -48,7 +48,8 @@ internal class ScreenCaptureProcessor(
         }
     }
 
-    suspend fun save(capture: Capture): ResultOf<Capture, Error> {
+    @Suppress("UnusedPrivateMember")
```
iujames, line 51: this suppression will go away with changes for API calls coming up

andretortolano (approved):

PR #343, opened 2023-02-24
initial work to add network calls for screen capture

iujames:
stacks on #342 

The primary work here was to add the new Retrofit interfaces, and related files for the three new endpoint destinations coming: SDK Settings, Customer API, and Image Upload.

After Slack discussion, this led to the broader restructuring / generalization of our Retrofit layer - we now have these packages for the different pieces:
Screenshot 2023-02-24 at 4 08 16 PM
This led to a ton of namespace updates and file changes here that aren't meaningful.

The meaningful changes:
* Got rid of `AppcuesRemoteSource` interface and renamed `RetrofitAppcuesRemoteSource` to `AppcuesRemoteSource` (matching the name of the prior interface) like we chatted about
* Added `CustomerApiService` and `CustomerApiRemoteSource` that uses it
* Added `SdkSettingsService` and `SdkSettingsRemoteSource` that uses it
* Added `ImageUploadService` and `ImageUploadRemoteSource` that uses it
* Made `RetrofitWrapper` a bit more flexible and generic so that it could be shared and used in all these spots, including some flexibility about which interceptors are installed for each

None of these new endpoints are actually being used by anything yet. Adding the actual screen capture flow logic will be next.


appcues/src/main/java/com/appcues/data/remote/DataRemoteKoin.kt
```diff
@@ -1,30 +1,63 @@
...
-        scoped<AppcuesRemoteSource> {
+        scoped {
             val config: AppcuesConfig = get()
-            RetrofitAppcuesRemoteSource(
-                appcuesService = getAppcuesService(config.apiBasePath ?: BASE_URL),
-                accountId = config.accountId,
+            AppcuesRemoteSource(
+                service = RetrofitWrapper(
+                    baseUrl = (config.apiBasePath ?: AppcuesRemoteSource.BASE_URL).toHttpUrl(),
+                    interceptors = listOf(MetricsInterceptor()),
```
iujames, line 27: this is the only one that needs the MetricsInterceptor at this point - we can now configure interceptors on a per service basis.

appcues/src/main/java/com/appcues/data/remote/DataRemoteKoin.kt
```diff
@@ -1,30 +1,63 @@
...
+                ).create(SdkSettingsService::class),
+                config = get(),
+            )
+        }
+
+        scoped {
+            CustomerApiRemoteSource(
+                service = RetrofitWrapper(
+                    baseUrl = null,
+                    interceptors = listOf(CustomerApiHostInterceptor()),
```
iujames, line 51: My idea here (and I haven't fully tried it out yet) is to allow this interceptor to dynamically inject the customer API host path, based on the Sdk Settings response - since that endpoint is discovered at runtime. This interceptor concept is mostly based on the various discussions in this post https://stackoverflow.com/a/62150282 on how to handle a runtime configured host path with Retrofit/OkHttp.

appcues/src/main/java/com/appcues/data/remote/NetworkRequest.kt
```diff
@@ -0,0 +1,27 @@
...
+
+object NetworkRequest {
+    internal suspend fun <Success> execute(apiCall: suspend () -> Success): ResultOf<Success, RemoteError> =
+        try {
+            ResultOf.Success(apiCall.invoke())
+        } catch (exception: HttpException) {
+            ResultOf.Failure(RemoteError.HttpError(exception.code(), convertErrorBody(exception)))
+        } catch (exception: Exception) {
+            ResultOf.Failure(RemoteError.NetworkError(exception))
+        }
```
iujames, line 17: this helper for error handling was in the `RetrofitAppcuesRemoteSource` previously - moved out to this shared spot so that it could be used by all the remote sources making network calls now.

appcues/src/main/java/com/appcues/data/remote/RetrofitWrapper.kt
```diff
@@ -0,0 +1,48 @@
...
+import okhttp3.logging.HttpLoggingInterceptor
+import retrofit2.Retrofit
+import retrofit2.converter.moshi.MoshiConverterFactory
+import kotlin.reflect.KClass
+
+internal class RetrofitWrapper(
+    private val baseUrl: HttpUrl?,
+    private val isDebug: Boolean = BuildConfig.DEBUG,
+    private val interceptors: List<Interceptor> = listOf(),
+    private val okhttpConfig: (OkHttpClient.Builder) -> OkHttpClient.Builder = { it }
```
iujames, line 17: this is mostly the same as before - when it was only used for appcues, but with enhancements
* `baseUrl` is optional
* you can pass in the custom list of interceptors based on the service
* you can chain on your own custom builder calls - like appcues uses to set the 5 sec timeout
* the retrofit instance is now a lazy prop, just in case multiple services are being created to use it (none do this currently)

appcues/src/main/java/com/appcues/data/remote/appcues/AppcuesRemoteSource.kt
```diff
@@ -0,0 +1,114 @@
...
+import com.appcues.data.remote.appcues.response.experience.ExperienceResponse
+import com.appcues.util.ResultOf
+import okhttp3.Interceptor
+import okhttp3.Interceptor.Chain
+import okhttp3.MediaType.Companion.toMediaTypeOrNull
+import okhttp3.RequestBody.Companion.toRequestBody
+import okhttp3.Response
+import java.util.UUID
+
+internal class AppcuesRemoteSource(
```
iujames, line 20: this is just what use to be `RetrofitAppcuesRemoteSource` - moved/renamed to flatten the structure.

appcues/src/main/java/com/appcues/data/remote/customerapi/CustomerApiRemoteSource.kt
```diff
@@ -0,0 +1,70 @@
...
+}
+
+internal class CustomerApiHostInterceptor : Interceptor {
+
+    companion object {
+        // customer API host is looked up in settings, and must be set here
+        // before any usage
+        lateinit var host: String
+    }
+
```
iujames, line 55: this is where I'm going to try to dynamically inject the host at runtime (haven't tried yet)

appcues/src/main/java/com/appcues/data/remote/imageupload/ImageUploadRemoteSource.kt
```diff
@@ -0,0 +1,26 @@
...
+    suspend fun upload(
+        url: String,
+        capture: Capture,
+    ): ResultOf<Unit, RemoteError> {
+
+        return NetworkRequest.execute {
+            service.upload(
+                url = url,
+                // still need to figure out how to actually get the PNG content of the capture for the PUT body
+                image = capture.displayName.toRequestBody("application/json; charset=utf-8".toMediaTypeOrNull()),
```
iujames, line 22: placeholder content here - we'll need to get the capture.screenshot `Bitmap` and send the PNG data on this PUT

andretortolano (approved):
nice work here, thanks for bringing this up for discussion!

PR #344, opened 2023-02-27
Implement network calls to save screenshot and screen

iujames:
This change implements the sequence of calls to upload a screenshot and save the screen into the customer account for builder usage. Some cleanup items found and worked along the way

* capture proper screen rect for elements
* capture system window insets (top/bottom bars) for builder usage in metadata
* size bitmap to screen capture size in DP (not a much larger 3x image for instance)

### DIFF
--- a/appcues/src/main/java/com/appcues/data/remote/imageupload/ImageUploadRemoteSource.kt
+++ b/appcues/src/main/java/com/appcues/data/remote/imageupload/ImageUploadRemoteSource.kt
@@ -1,26 +1,36 @@
 package com.appcues.data.remote.imageupload
 
+import android.graphics.Bitmap
+import android.graphics.Bitmap.CompressFormat.PNG
 import com.appcues.data.remote.NetworkRequest
 import com.appcues.data.remote.RemoteError
-import com.appcues.debugger.screencapture.Capture
 import com.appcues.util.ResultOf
 import okhttp3.MediaType.Companion.toMediaTypeOrNull
 import okhttp3.RequestBody.Companion.toRequestBody
+import java.io.ByteArrayOutputStream
 
 internal class ImageUploadRemoteSource(
     private val service: ImageUploadService,
 ) {
+    companion object {
+        private const val IMAGE_QUALITY = 100
+    }
+
     suspend fun upload(
         url: String,
-        capture: Capture,
+        bitmap: Bitmap,
     ): ResultOf<Unit, RemoteError> {
-
         return NetworkRequest.execute {
             service.upload(
                 url = url,
-                // still need to figure out how to actually get the PNG content of the capture for the PUT body
-                image = capture.displayName.toRequestBody("application/json; charset=utf-8".toMediaTypeOrNull()),
+                image = bitmap.toData().toRequestBody("image/png".toMediaTypeOrNull()),
             )
         }
+    }
+
+    private fun Bitmap.toData(): ByteArray {
+        val stream = ByteArrayOutputStream()
+        this.compress(PNG, IMAGE_QUALITY, stream)
+        return stream.toByteArray()
     }
 }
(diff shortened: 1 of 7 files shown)


appcues/src/main/java/com/appcues/data/remote/imageupload/ImageUploadRemoteSource.kt
```diff
@@ -1,26 +1,36 @@
...
-        capture: Capture,
+        bitmap: Bitmap,
     ): ResultOf<Unit, RemoteError> {
-
         return NetworkRequest.execute {
             service.upload(
                 url = url,
-                // still need to figure out how to actually get the PNG content of the capture for the PUT body
-                image = capture.displayName.toRequestBody("application/json; charset=utf-8".toMediaTypeOrNull()),
+                image = bitmap.toData().toRequestBody("image/png".toMediaTypeOrNull()),
```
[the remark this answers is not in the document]
andretortolano: I've never done it, but it seems like if we notice we need to improve this for N reasons we can come back to [this](https://adinugroho.medium.com/upload-image-from-android-app-using-retrofit-2-ae6f922b184c)

appcues/src/main/java/com/appcues/debugger/screencapture/ScreenCaptureProcessor.kt
```diff
@@ -48,14 +60,87 @@ internal class ScreenCaptureProcessor(
...
-    @Suppress("UnusedPrivateMember")
-    suspend fun save(capture: Capture, token: String): ResultOf<Capture, Error> {
+    suspend fun save(capture: Capture, token: String) =
+        // Saving a screen is a 4-step chain. This is implemented here as a sequence of calls, chaining
+        // a success continuation on each call to move to the next step. If any step fails, the RemoteError
+        // is bubbled up to the caller of this function instead of the successful result, and an error
+        // can be shown and handled.
+
+        // step 1 - get the settings, with the path to customer API
+        getSettings { sdkSettings ->
```
andretortolano, line 70: wondering if you are interested in changing this piece of code to look like this:


```kotlin

    suspend fun save(capture: Capture, token: String): ResultOf<Capture, RemoteError> {
        // Saving a screen is a 4-step chain. This is implemented here as a sequence of calls, chaining
        // a success continuation on each call to move to the next step. If any step fails, the RemoteError
        // is bubbled up to the caller of this function instead of the successful result, and an error
        // can be shown and handled.
        return try {
            // step 1 - get the settings, with the path to customer API
            sdkSettingsRemoteSource.getSettings()
                // step 2 - use the customer API path to get the link to upload the screenshot
                .getUploadPath(capture, token)
                // step 3 - upload the screenshot
                .uploadImage(capture)
                // step 4 - update the screenshot link and save the screen
                .saveScreen(capture, token)
        } catch (e: ScreenCaptureSaveException) {
            Failure(e.error)
        }
    }
```

andretortolano (approved):

appcues/src/main/java/com/appcues/debugger/screencapture/ScreenCaptureProcessor.kt
andretortolano: I achieved this by removing the callbacks on success and chaining the return to one that made sense for the next step, on every remote call I would check if the result is failure and in case it is I throw this private ScreenCaptureSaveException that we try/catch on save to ensure that any exception here is captured and returned as failure later.

```kotlin
    private class ScreenCaptureSaveException(val error: RemoteError) : Exception("ScreenCaptureException: $error")

    // step 1 - get the settings, with the path to customer API
    private suspend fun SdkSettingsRemoteSource.getSettings(): SdkSettingsResponse {
        return sdkSettings().let {
            when (it) {
                is Failure -> throw ScreenCaptureSaveException(it.reason)
                is Success -> it.value
            }
        }
    }
```

iujames: Thanks for the idea! I do think it reads more cleanly this way, giving it a shot.

andretortolano (approved):
Looks good!

PR #345, opened 2023-02-28
♻️  update for ui tooltip tests

andretortolano:
minor changes to skippable and tooltip traits for consistency matching ui updates.

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/extensions/DrawScopeEx.kt
+++ b/appcues/src/main/java/com/appcues/ui/extensions/DrawScopeEx.kt
@@ -13,7 +13,7 @@ import androidx.compose.ui.unit.dp
 internal fun DrawScope.xShapePath(color: Color, deflateDp: Dp = 0.dp): Path {
     return Path()
         .apply {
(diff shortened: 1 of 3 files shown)


iujames (approved):

appcues/src/main/java/com/appcues/ui/extensions/DrawScopeEx.kt
```diff
@@ -13,7 +13,7 @@ import androidx.compose.ui.unit.dp
 internal fun DrawScope.xShapePath(color: Color, deflateDp: Dp = 0.dp): Path {
     return Path()
         .apply {
-            val strokeWidth = 2.dp.toPx()
+            val strokeWidth = 1.5.dp.toPx()
```
iujames, line 16: this will change the X on the screen capture confirmation dialog - but that is probably ok I think.

andretortolano: it changes the x confirmation dialog, and I think its ok

PR #346, opened 2023-02-28
Add success and failure toast messages on screen capture

iujames:
stacks on #344 

This adds the blue success and pink failure toasts per design. Failure allows a retry.

A new `ToastState` is used in the DebuggerViewModel, in addition to the `UIState`, as toasts are really independent of other UI and can show up on top of whatever other UI state is rendering. They auto dismiss after 3 seconds, or if the background is tapped.

Failure:

https://user-images.githubusercontent.com/19266448/221929787-9b65e1e4-4a21-42d3-a278-a18b6a6738f2.mov

Success:

https://user-images.githubusercontent.com/19266448/221929846-f391e510-66d4-4c33-91bf-ba4ab1a3153d.mov

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/ui/ToastView.kt
+++ b/appcues/src/main/java/com/appcues/debugger/ui/ToastView.kt
@@ -1,0 +1,157 @@
+package com.appcues.debugger.ui
+
+import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.foundation.background
+import androidx.compose.foundation.border
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.interaction.MutableInteractionSource
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.BoxScope
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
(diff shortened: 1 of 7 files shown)


appcues/src/main/java/com/appcues/debugger/ui/ToastView.kt
```diff
@@ -0,0 +1,157 @@
...
+import kotlinx.coroutines.delay
+
+private const val TOAST_LENGTH = 3_000L
+
+@Composable
+internal fun BoxScope.ToastView(debuggerState: MutableDebuggerState) {
+
+    when (val toastState = debuggerState.toast.targetState) {
+        is ScreenCaptureSuccess -> SuccessToast(toast = toastState, debuggerState = debuggerState)
+        is ScreenCaptureFailure -> FailureToast(toast = toastState, debuggerState = debuggerState)
```
iujames, line 43: There is some commonality in these toasts (general size and position) but enough difference (color, one has button, behaviors) that I found it simplest to just split out into two separate compositions that we can manage independently, IMO.

andretortolano (approved):
Compose master, looks good, seems to fit well with existing compositiong

PR #348, opened 2023-03-01
Apply content padding from modal to the sticky content within

iujames:
This will do the trick on the `release/1.3.1` branch - the sticky content composition needs the `contentPadding` passed down, just like the main content.

The change will differ on the `main` branch, as now this padding is part of the `modifier` passed through from the content wrapper (modal) -- will PR that separately and we'll handle merge conflicts when the branches sync up.

For example, consider this scrolling full screen modal with a button at the bottom. It has 20px padding all around defined at the modal level - no sticky content. Note, no margin set between the bottom of the last text and the top of the button at bottom.

https://user-images.githubusercontent.com/19266448/222220954-db6845a9-2f02-4c4e-8bfb-e9bb76239469.mov

Now, make that bottom button row `sticky=top`

| Before | After |
| --- | --- |
| Screenshot 2023-03-01 at 12 48 53 PM | Screenshot 2023-03-01 at 12 47 53 PM |

In the after screenshot, notice that the 20px padding leading/trailing/bottom are showing as expected on the sticky button as well.

It also applies the correct additional sticky content padding still, so that the scrolling content can scroll up enough to still be fully visible above the sticky bottom row - in this case, the text scrolls until its bottom lines up with the top of the button (no margin set here for accurate testing).

Screenshot 2023-03-01 at 12 52 10 PM

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/composables/StepComposition.kt
+++ b/appcues/src/main/java/com/appcues/ui/composables/StepComposition.kt
@@ -56,17 +56,21 @@ internal fun Step.ApplyOverlayStepTraits(boxScope: BoxScope, stepDecoratingPaddi
 }
 
 @Composable
-internal fun Step.ComposeStickyContent(boxScope: BoxScope, stepDecoratingPadding: StepDecoratingPadding) {
+internal fun Step.ComposeStickyContent(boxScope: BoxScope, contentPadding: PaddingValues?, stepDecoratingPadding: StepDecoratingPadding) {
     topStickyContent?.let {
         Box(
-            modifier = Modifier.alignStepOverlay(boxScope, Alignment.TopCenter, stepDecoratingPadding),
-            contentAlignment = Alignment.BottomCenter
+            modifier = Modifier
+                .then(if (contentPadding != null) Modifier.padding(contentPadding) else Modifier)
+                .alignStepOverlay(boxScope, Alignment.TopCenter, stepDecoratingPadding),
+            contentAlignment = Alignment.TopCenter
         ) { it.Compose() }
     }
 
     bottomStickyContent?.let {
         Box(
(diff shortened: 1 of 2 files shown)


appcues/src/main/java/com/appcues/ui/composables/StepComposition.kt
```diff
@@ -56,17 +56,21 @@ internal fun Step.ApplyOverlayStepTraits(boxScope: BoxScope, stepDecoratingPaddi
...
-internal fun Step.ComposeStickyContent(boxScope: BoxScope, stepDecoratingPadding: StepDecoratingPadding) {
+internal fun Step.ComposeStickyContent(boxScope: BoxScope, contentPadding: PaddingValues?, stepDecoratingPadding: StepDecoratingPadding) {
     topStickyContent?.let {
         Box(
-            modifier = Modifier.alignStepOverlay(boxScope, Alignment.TopCenter, stepDecoratingPadding),
-            contentAlignment = Alignment.BottomCenter
+            modifier = Modifier
+                .then(if (contentPadding != null) Modifier.padding(contentPadding) else Modifier)
+                .alignStepOverlay(boxScope, Alignment.TopCenter, stepDecoratingPadding),
+            contentAlignment = Alignment.TopCenter
```
iujames, line 65: note: this was `BottomCenter` before, on the top case, but seems it should match the alignment on the line above? in practice I don't think it really matters much, since the box height is the same as the sticky content height inside it here.

andretortolano (approved):

appcues/src/main/java/com/appcues/ui/composables/StepComposition.kt
andretortolano: yes it wont change much but its good to keep it matching the alignStepOverlay for consistency

PR #353, opened 2023-03-07
♻️  Updated logic around positioning the edge and corner radius for tooltip container

andretortolano:
Tried as much as possible to simplify how we were trying to handle some of pointer position in relation to the container.

with this change now when we position target-rect closer to the edge, we have special handling of its appearance.

https://user-images.githubusercontent.com/5244805/223538533-7ec0deab-a4c2-40f8-801d-52386f85ea0b.mov


iujames (approved):
looking good!

PR #354, opened 2023-03-07
Implement target element trait

iujames:
## Handling Trait Errors
A big chunk of this work was to support propagation of trait errors during rendering, and appropriately ending an experience when that occurs. For example: a selector not found error when trying to render a tooltip step with a target element trait. This was generally handled by changing the completion callback on render from: 
```
val presentationComplete: (() -> Unit),
```
to
```
val presentationComplete: ((ResultOf<Unit, Error>) -> Unit),
```
This applies to the BeginningStep state property that is used in transition to RenderingStep. Now, the UI layer can signal that the presentation failed, via the Error result there, and allow the state machine to be aware of that and transition back out of the experience and trigger the proper analytics.

To actually handle the error, when a trait exception arises during the processing of metadata, deep in the composition, it is bubbled back up to the ViewModel using the `LocalViewModel` (CompositionLocal). The VM can then handle informing the StateMachine, and finally dismissing the view. The dismissal is handled in the view model, now that it is aware of the `onDismiss` handler that gets passed through from wherever the view model was initiated (ie. AppcuesOverlayManager for tooltips, AppcuesActivity for modals).

## Target Element Trait Implementation
A few elements that are common to target-rectangle and target-element traits are moved to the shared `TargetContent.kt` file - `TargetRectangleInfo`, the string key for "targetRectangle" metadata, and `ContentPreferredPosition` enum. The `TargetElementTrait` then works to look up the given selector in the current view hierarchy and, if found, populate the "targetRectangle" metadata from it. The rest of the rendering logic for keyhole and tooltips is unchanged, as it just relies on this target rectangle.

open question: iOS has a way to detect when the screen has changed size or layout in some form and recalculate this rect - it is not clear if we need anything like this, or if the normal re-composition will handle it in this case, since it's all Compose here.

## Example
This shows a 3-step element targeted tooltip experience
* find view with contentDescription "Profile", for the profile tab, first step
* find view with tag "txtGivenName" for the second step
* find view with tag "btnSaveProfile" for the third step

https://user-images.githubusercontent.com/19266448/223548512-e1c1103b-afcf-467f-8aa4-1d6eb076184e.mov

> note: some additional logic is being used in the sample app here between step 1 and 2, since a navigation action is taking place. This requires that the host app uses the `NavigationHandler` to inform Appcues when the deep link navigation has completed, otherwise target element experiences can fail, if the view they are looking for has not been fully loaded into the current activity yet. This will likely be a point of customer complication for any more sophisticated walk through type of experiences.

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/appcues/BackdropKeyholeTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/BackdropKeyholeTrait.kt
@@ -63,8 +63,10 @@ internal class BackdropKeyholeTrait(
         val density = LocalDensity.current
         val metadata = LocalAppcuesStepMetadata.current
 
+        // trait does not apply if no target rectangle in metadata
+        val targetRectInfo = rememberTargetRectangleInfo(metadata) ?: return
+
         val shapeBlurRadius = if (shape == CIRCLE) configBlurRadius.toFloat() else 0.0f
-        val targetRectInfo = rememberTargetRectangleInfo(metadata)
         val targetRect = targetRectInfo.getRect(rememberAppcuesWindowInfo()).inflateOrEmpty(configSpreadRadius)
         val floatAnimation = rememberFloatStepAnimation(metadata)
         val encompassesDiameter = getRectEncompassesRadius(targetRect.width, targetRect.height, shapeBlurRadius) * 2
@@ -92,20 +94,22 @@ internal class BackdropKeyholeTrait(
                 .drawWithContent {
                     drawContent()
 
-                    val shapeCenter = Offset(positionPx.x + sizePx.width / 2, positionPx.y + sizePx.height / 2)
-                    val blurStartPoint = ((sizePx.width / 2) - blurRadiusPx) / (sizePx.width / 2)
+                    if (sizePx.width > 0 && sizePx.height > 0) {
+                        val shapeCenter = Offset(positionPx.x + sizePx.width / 2, positionPx.y + sizePx.height / 2)
+                        val blurStartPoint = ((sizePx.width / 2) - blurRadiusPx) / (sizePx.width / 2)
 
-                    drawRoundRect(
-                        topLeft = positionPx,
-                        brush = Brush.radialGradient(
-                            colorStops = arrayOf(blurStartPoint to Color.Transparent, 1.0f to Color.Black),
-                            center = shapeCenter,
-                            radius = encompassesRadiusPx.value
-                        ),
-                        size = sizePx,
-                        cornerRadius = rectCornerRadius,
-                        blendMode = BlendMode.DstIn
-                    )
+                        drawRoundRect(
+                            topLeft = positionPx,
+                            brush = Brush.radialGradient(
+                                colorStops = arrayOf(blurStartPoint to Color.Transparent, 1.0f to Color.Black),
+                                center = shapeCenter,
+                                radius = encompassesRadiusPx.value
+                            ),
+                            size = sizePx,
+                            cornerRadius = rectCornerRadius,
+                            blendMode = BlendMode.DstIn
+                        )
+                    }
                 }
         ) {
             content()
--- a/appcues/src/main/java/com/appcues/trait/appcues/BackdropKeyholeTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/BackdropKeyholeTrait.kt
@@ -63,8 +63,7 @@ internal class BackdropKeyholeTrait(
         val density = LocalDensity.current
         val metadata = LocalAppcuesStepMetadata.current
 
-        // trait does not apply if no target rectangle in metadata
-        val targetRectInfo = rememberTargetRectangleInfo(metadata) ?: return
+        val targetRectInfo = rememberTargetRectangleInfo(metadata)
 
         val shapeBlurRadius = if (shape == CIRCLE) configBlurRadius.toFloat() else 0.0f
         val targetRect = targetRectInfo.getRect(rememberAppcuesWindowInfo()).inflateOrEmpty(configSpreadRadius)
@@ -94,7 +93,7 @@ internal class BackdropKeyholeTrait(
                 .drawWithContent {
                     drawContent()
 
-                    if (sizePx.width > 0 && sizePx.height > 0) {
+                    if (encompassesRadiusPx.value > 0) {
                         val shapeCenter = Offset(positionPx.x + sizePx.width / 2, positionPx.y + sizePx.height / 2)
                         val blurStartPoint = ((sizePx.width / 2) - blurRadiusPx) / (sizePx.width / 2)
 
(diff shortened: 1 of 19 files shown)


appcues/src/main/java/com/appcues/trait/appcues/BackdropKeyholeTrait.kt
```diff
@@ -63,8 +63,10 @@ internal class BackdropKeyholeTrait(
         val density = LocalDensity.current
         val metadata = LocalAppcuesStepMetadata.current
 
+        // trait does not apply if no target rectangle in metadata
+        val targetRectInfo = rememberTargetRectangleInfo(metadata) ?: return
```
andretortolano, line 67: can you go over in more depth what was the issue here in this trait?

appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
```diff
@@ -101,8 +100,10 @@ internal class TooltipTrait(
         val density = LocalDensity.current
         val metadata = LocalAppcuesStepMetadata.current
 
+        // trait does not apply if no target rectangle in metadata
+        val targetRectInfo = rememberTargetRectangleInfo(metadata) ?: return
```
andretortolano, line 104: so in the case target rect is not available I had set to show the tooltip was preferred position NONE, which position the tooltip at the bottom of the view. wondering if now that we have a more defined behavior if we should remove that entirely

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
```diff
@@ -114,47 +111,52 @@ private fun BoxScope.ComposeLastRenderingState(state: Rendering) {
         val metadataSettingTraits = remember(state.position) { mutableStateOf(steps[state.position].metadataSettingTraits) }
         val previousStepMetaData = remember { mutableStateOf(AppcuesStepMetadata()) }
         val stepMetadata = remember(metadataSettingTraits.value) {
-
-            val actual = hashMapOf<String, Any?>().apply { metadataSettingTraits.value.forEach { putAll(it.produceMetadata()) } }
-
-            mutableStateOf(AppcuesStepMetadata(previous = previousStepMetaData.value.actual, actual = actual)).also {
-                previousStepMetaData.value = it.value
+            try {
```
andretortolano, line 114: maybe move this remember block to a function. called

rememberStepmetadata(metadataSettingTraits, onTraitException = { 
    viewModel.onTraitException(ex)
    return@MainSurface (something like that)
})

andretortolano: if its not onTraitException the return value is not null

andretortolano (approved):
Nice iteration over ElementTargeting code

iujames (commented):
Just realized I did not submit some of my explanation comments, whoops. Doing so now.

appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
iujames: right, I need to review this part again, I missed that and it should not just ignore the trait here.

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
iujames: I'm not sure I follow exactly yet. I can extract to a function, but the behavior right now does return a non-null value when it is not the exception case. The exception case returns a null value, which is then later used to prevent the rest of the composition from attempting to render with invalid metadata.

appcues/src/main/java/com/appcues/trait/appcues/BackdropKeyholeTrait.kt
```diff
@@ -92,20 +94,22 @@ internal class BackdropKeyholeTrait(
...
-                        brush = Brush.radialGradient(
-                            colorStops = arrayOf(blurStartPoint to Color.Transparent, 1.0f to Color.Black),
-                            center = shapeCenter,
-                            radius = encompassesRadiusPx.value
-                        ),
-                        size = sizePx,
-                        cornerRadius = rectCornerRadius,
-                        blendMode = BlendMode.DstIn
-                    )
+                    if (sizePx.width > 0 && sizePx.height > 0) {
```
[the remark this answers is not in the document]
andretortolano: is that for when we have the Keyhole trait with no extra configuration? or is that just the size of the rect ? there is probably a better way we could make this work, I'm thinking that even if we have a rect of 0x0, we should be able to draw the keyhole if it has set other settings like blurRadius for instance

appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
iujames: reverted this and tested it out - no target rectangle or element, and no backdrop-keyhole - just the default backdrop

It seems pretty close, some sort of sizing issue clipping off the bottom.

Screenshot 2023-03-08 at 9 06 24 AM

appcues/src/main/java/com/appcues/trait/appcues/BackdropKeyholeTrait.kt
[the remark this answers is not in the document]
iujames: it is not - reverting.

appcues/src/main/java/com/appcues/trait/appcues/BackdropKeyholeTrait.kt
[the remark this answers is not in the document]
iujames: this would be the case when there is either no target rect, or a target rect with size zero, in theory. Maybe a better way to guard is `encompassesRadiusPx.value` must be > 0.

appcues/src/main/java/com/appcues/trait/appcues/TargetElementTrait.kt
```diff
@@ -1,36 +1,88 @@
...
+        val insets = ViewCompat.getRootWindowInsets(rootView)?.getInsets(WindowInsetsCompat.Type.systemBars()) ?: Insets.NONE
+
+        // this is the position of the view relative to the entire screen
+        val actualPosition = Rect()
+        selectedView.getGlobalVisibleRect(actualPosition)
+
+        // this global position includes system bars, so we need to subtract those insets to get coordinates relative to the root
+        // view of what we are drawing content on top of
+        val x = actualPosition.left - insets.left
+        val y = actualPosition.top - insets.top
```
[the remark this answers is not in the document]
andretortolano: I think so, will keep an eye on this one

appcues/src/main/java/com/appcues/trait/appcues/BackdropKeyholeTrait.kt
iujames: updated to guard `encompassesRadiusPx.value` must be > 0 instead on the `drawRoundRect` block, which I think is the appropriate check.

PR #361, opened 2023-03-13
Fix/google api char limit

andretortolano:
even when we limit the char count to 500, in some situations google play counted more chars, I imagine this is because of "/n",  after some local testing even when trying to truncate shorter messages, I never get the exact char count. so as a simple fix since this is not that important lets change from 500 to 490 so we have some margins for any inconsistencies

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -24,9 +24,11 @@ platform :android do
 
     git_log = changelog_from_git_commits(
       commits_count: 5,
-      pretty: "- [%as] %s"
+      pretty: "- [%as] %<(70,trunc)%s"
     )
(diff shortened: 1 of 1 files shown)


mmaatttt (approved):
Should work fine 👍 

I was curious, and it turns out there's a way to truncate each commit message with a weird looking `trunc` syntax: https://git-scm.com/docs/git-log#Documentation/git-log.txt-emltltNgttruncltruncmtruncem

```sh
pretty: "- [%as] %<(70,trunc)%s"
```

```
// gets you
- [2023-02-17] 🔧 Add repo-specific whitesource/mend settings disable "Renovate" (a..
- [2023-02-15] ✅ Update experience decoding error test                              
```

iujames (approved):

PR #362, opened 2023-03-17
✨ Enable horizontal positioning for tooltip container

andretortolano:
Added Left and Right as preferredContentPosition supported for tooltips,

then added piece of calculation rules to determine where we should position the tooltip, and this is mirroring behavior described for IOS and WEB tooltip.

* Another pass on tooltip sizing related to task [#48999](https://app.shortcut.com/appcues/story/48999/android-update-tooltip-to-fit-in-visible-area)
* [This is](https://github.com/appcues/appcues-android-sdk/blob/2de61c114632faec66f4e9fa0da8e33d0e59f102/appcues/src/main/java/com/appcues/trait/extensions/TargetRectangleInfoExt.kt#L55) the method that should mirror IOS/WEB rules for tooltip positioning
* UI tests running OK

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/appcues/TooltipsPaths.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/TooltipsPaths.kt
@@ -58,6 +58,12 @@ internal sealed class TooltipPointerPosition {
         this._alignment.value = PointerAlignment(horizontalAlignment = alignment)
     }
 
+    /**
+     * returns ContainerCornerRadius that based on TooltipPointerPosition, will ignore/use
+     * the existing cornerRadius value for some of the edges.
+     *
+     * eg. When HorizontalAlignment is Left the container topStart should not apply cornerRadius.
+     */
     abstract fun toContainerCornerRadius(cornerRadius: Dp): ContainerCornerRadius
 
     object Top : TooltipPointerPosition() {
@@ -240,6 +246,13 @@ private fun getTooltipPointerPath(
     }
 }
 
+/**
+ * Calculates the pointerXOffset position, returns a pair of value containing:
+ * first: pointer x offset
+ * second: pointer tip offset
+ *
+ * during the processing of this function we also update the tooltipPointerPosition horizontal alignment
+ */
 private fun calculatePointerXOffset(
     containerDimens: TooltipContainerDimens,
     tooltipSettings: TooltipSettings,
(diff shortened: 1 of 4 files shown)


appcues/src/main/java/com/appcues/trait/appcues/TooltipsPaths.kt
```diff
@@ -163,89 +198,166 @@ private fun getTooltipPointerPath(
...
+    )
+
+    val (pointerOffsetY, horizontalTipOffset) = calculatePointerYOffset(
+        containerDimens = containerDimens,
+        tooltipSettings = tooltipSettings,
+        pointerOffsetY = with(density) { tooltipSettings.pointerOffsetY.value.toPx() },
+        cornerRadiusPx = cornerRadiusPx,
+    )
+
+    val tipLength = animateFloatAsState(
```
andretortolano, line 217: moving the animated tip length here since has similar logic untied to being horizontal or vertical positioning

appcues/src/main/java/com/appcues/trait/appcues/TooltipsPaths.kt
```diff
@@ -266,16 +378,17 @@ private fun getContainerPath(
...
-
-        val containerOffsetPx = Offset(
-            x = 0f,
-            y = when (tooltipSettings.tooltipPointerPosition) {
-                // offset entire content rectangle in case tooltip is at the top
-                is Top -> tooltipSettings.pointerLengthPx
-                else -> 0f
-            }
-        )
+        val size = if (tooltipSettings.tooltipPointerPosition.isVertical) {
```
andretortolano, line 381: size is calculated based on orientation

appcues/src/main/java/com/appcues/trait/appcues/TooltipsPaths.kt
```diff
@@ -74,18 +76,50 @@ internal sealed class TooltipPointerPosition {
...
+                topEnd = if (alignment.value.verticalAlignment == PointerVerticalAlignment.TOP) 0.dp else cornerRadius,
+                bottomEnd = if (alignment.value.verticalAlignment == PointerVerticalAlignment.BOTTOM) 0.dp else cornerRadius,
+                bottomStart = cornerRadius,
             )
         }
     }
 
     object None : TooltipPointerPosition() {
 
+        override val isVertical = false
```
iujames, line 121: Does this case represent when no position preference is given in the builder?

appcues/src/main/java/com/appcues/trait/appcues/TooltipsPaths.kt
```diff
@@ -163,89 +198,166 @@ private fun getTooltipPointerPath(
...
+        containerDimens = containerDimens,
+        tooltipSettings = tooltipSettings,
+        pointerOffsetY = with(density) { tooltipSettings.pointerOffsetY.value.toPx() },
+        cornerRadiusPx = cornerRadiusPx,
+    )
+
+    val tipLength = animateFloatAsState(
+        targetValue = when (tooltipSettings.tooltipPointerPosition) {
+            Top, Left -> -tooltipSettings.pointerLengthPx
+            Bottom, Right -> tooltipSettings.pointerLengthPx
```
iujames, line 220: can these two case be combined?

appcues/src/main/java/com/appcues/trait/appcues/TooltipsPaths.kt
```diff
@@ -163,89 +198,166 @@ private fun getTooltipPointerPath(
...
+    var pointerTipOffset = 0f
+    var pointerOffset = (containerDimens.widthPx / 2) - tooltipSettings.pointerBaseCenterPx + pointerOffsetX
+
+    when {
+        tooltipSettings.tooltipPointerPosition is Right -> {
+            pointerOffset = containerDimens.widthPx - tooltipSettings.pointerLengthPx
+        }
+        tooltipSettings.tooltipPointerPosition is Left -> {
+            pointerOffset = tooltipSettings.pointerLengthPx
+        }
```
iujames, line 265: I'm a bit lost on what these are doing to be honest

iujames (approved):
can't pretend to follow all the calculation details here, but thats what the tests are for! Nice work hammering away at all the cases.

appcues/src/main/java/com/appcues/trait/appcues/TooltipsPaths.kt
iujames: they all do the same thing right? return `tooltipSettings.pointerLengthPx`? Seemed like line 219 and 220 had identical result

appcues/src/main/java/com/appcues/trait/appcues/TooltipsPaths.kt
iujames: ah right, yeah i think the floating bottom case is valid, might be useful for toast like things. I recall something similar in a competitor product as well.

appcues/src/main/java/com/appcues/trait/appcues/TooltipsPaths.kt
iujames: ah ok the edge case for when the pointer target element is on the far edge of the screen - all good, maybe just a comment would help remind what case that is handling.

appcues/src/main/java/com/appcues/trait/appcues/TooltipsPaths.kt
andretortolano: line 219 is negativa value

iujames: ah! thank you

PR #363, opened 2023-03-22
Support configurable element targeting strategy and default Android View implementation

iujames:
This adds the capability to supply an `ElementTargeting` implementation into the config at initialization, which supports extensibility for other strategies around capturing screens and targeting/selecting elements. By default, it uses an Android View "standard" implementation. But, other uses like the React Native plugin, for example, can supply their own.

This adds a few types to the public API to support this

ElementTargeting - interface for the "strategy"
ElementSelector - interface to define selector matching behavior
ViewElement - the layout metadata for a View, used in capture and target element rendering


appcues/src/main/java/com/appcues/debugger/screencapture/AndroidViewTargeting.kt
```diff
@@ -0,0 +1,111 @@
...
+
+        (target as? AndroidViewSelector)?.let {
+            if (it.contentDescription != null && it.contentDescription == contentDescription) {
+                weight += 1
+            }
+
+            if (it.tag != null && it.tag == tag) {
+                weight += 1
+            }
+        }
```
iujames, line 38: definitely subject to evolution and change, but this allows for some weighted score based on multiple properties

want to look into using the resource ID as an option as well sometime soon

appcues/src/main/java/com/appcues/ElementTargeting.kt
```diff
@@ -0,0 +1,133 @@
...
+
+import android.view.View
+import com.squareup.moshi.JsonClass
+import java.util.UUID
+
+/**
+ * Defines an element targeting strategy type, which can be used to capture view layout
+ * information in the application UI as well as render element targeted experiences.
+ */
+interface ElementTargeting {
```
andretortolano, line 11: the commends define well that this is supposed to do. wondering if the naming here could include "strategy" to make it more clear, how does ElementTargeting sounds without any context do you think?

andretortolano: Android could be named AndroidTargetingStrategy, etc

appcues/src/main/java/com/appcues/ElementTargeting.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * Capture the layout hierarchy in the currently rendered screen of the application.
+     *
+     * @return: Root view element for the current screen, or `nil` if not available. The view element
```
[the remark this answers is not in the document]
iujames: copy paste from iOS 🙃 thank you!

andretortolano (approved):
Looks good!

PR #366, opened 2023-03-24
Fix compiler warnings for deprecated functions

iujames:
Some general clean up I thought we could include in next major update.

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/AppcuesActivity.kt
+++ b/appcues/src/main/java/com/appcues/ui/AppcuesActivity.kt
@@ -43,7 +43,9 @@ internal class AppcuesActivity : AppCompatActivity() {
 
     private val onBackPressCallback = object : OnBackPressedCallback(true) {
         override fun handleOnBackPressed() {
-            // disable our own back press callback
+            // disable our own back press callback - so that the check below for hasEnabledCallbacks
+            // will only be checking for callback additional to our own here (avoid infinite recursion
+            // in the onBackPressed call)
             this.isEnabled = false
 
             // then check if we have any other callback, i.e. debugger
--- a/appcues/src/main/java/com/appcues/ui/AppcuesActivity.kt
+++ b/appcues/src/main/java/com/appcues/ui/AppcuesActivity.kt
@@ -3,6 +3,7 @@ package com.appcues.ui
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
+import androidx.activity.OnBackPressedCallback
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.os.bundleOf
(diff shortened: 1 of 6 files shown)


appcues/src/main/java/com/appcues/util/PackageManagerExt.kt
```diff
@@ -0,0 +1,33 @@
...
+import android.content.pm.ActivityInfo
+import android.content.pm.PackageInfo
+import android.content.pm.PackageManager
+import android.content.pm.ResolveInfo
+import android.os.Build
+
+// extensions to handle deprecations in Android 33 of PackageManager functions
+// https://developer.android.com/reference/android/content/pm/PackageManager
+
+internal fun PackageManager.getPackageInfoCompat(packageName: String, flags: Int = 0): PackageInfo =
```
iujames, line 14: concept for these Compat extensions from https://stackoverflow.com/a/74741495

there is a feature request in with Google to support more officially in some form https://issuetracker.google.com/issues/246845196

appcues/src/main/java/com/appcues/ui/AppcuesActivity.kt
```diff
@@ -40,6 +41,24 @@ internal class AppcuesActivity : AppCompatActivity() {
...
+                // call it if so
+                onBackPressedDispatcher.onBackPressed()
+                // then re-enable our own callback
+                this.isEnabled = true
+            } else {
+                // otherwise initiate dismissal
+                viewModel.onBackPressed()
+            }
+        }
+    }
```
iujames, line 60: I believe this is functionally equivalent to what we had before, but this is the one I'd appreciate an extra careful look on please. I tested and when i have the debugger open on top of an experience, it appears to work the same as before: first back collapses the debugger, then second back dismisses the experience.

andretortolano: it looks  similar, I'm interested to know if there was any specific scenario that required you to change the state of "isEnabled" to false and back to enabled, instead of just disabling on the else branch

andretortolano (approved):
nice changes, interesting to know reasoning for changing "isEnabled" state

appcues/src/main/java/com/appcues/ui/AppcuesActivity.kt
iujames: I believe if we do not set `isEnabled` to false, then the check on line 50 will always return true
```
if (onBackPressedDispatcher.hasEnabledCallbacks()) {
```
and we'll enter into an infinite loop calling our own back press handler.

andretortolano: oh that makes sense, we need to account for this callback when asking if there is any enabled callback

iujames: I'll add a comment in the code to make it clear why it is doing that!

PR #368, opened 2023-03-29
Add support for close button styling

iujames:
This PR stacks on #367 and adds the new styling support in the `SkippableTrait`. 

I would suggest not really looking at the diff here and really just looking at `SkippableTrait` as a new "2.0" implementation of the close button, since it now has a much different capability set. Almost all the changes are in there, other than the one related change to `X` path drawing in `DrawScopeExt` (and updated usage in one other spot in `CaptureConfirmDialog`).

There were a lot of little details covered here, but in summary - it handles all of the spec updates defined in https://github.com/appcues/appcues-mobile-experience-spec/pull/165 for styling options in `buttonStyle` on the `@appcues/skippable` trait. 

I restructured the code a bit to pull out the distinct differences in how `default` and `minimal` appearances are handled. In a couple spots, I opted to use a specific implementation in the file here (border and shadow) rather than try to use the shared one in `ModifierExt` - the reason being that there were subtle specific details here around how the default corner radius should be handled, for back compat, and I didn't want to over complicate the shared version used by everything else for experience rendering.

Next I'll work on UI testing to match what is in https://github.com/appcues/appcues-mobile-experience-spec/pull/165, for the Android side. In the short term, I'll include some side by side screens with iOS and Android to demonstrate the core capabilities here. iOS on the left, Android on the right.

| test case | screenshot|
| --- | ---|
| default on dark, no styling | appcues-skippable-default |
| default on light, no styling | appcues-skippable-default-light |
| default, styled | appcues-skippable-default-styled |
| minimal on dark, no styling | appcues-skippable-minimal |
| minimal on light no styling | appcues-skippable-minimal-light |
| minmal, styled | appcues-skippable-minimal-styled |

notes:
* the default background circle should have the same discrepancy on ios/android as today - just due to iOS using the vibrancy/blur effect there, and android using a semi-transparent fixed `0x54000000`
* the shadow behind the `X` in minimal was a little tricky, but I think I have it working, though it looks slightly less pronounced on android than iOS. In any case, this is not expected to be a builder option anytime soon, maybe ever.
* the minimal 'X' approach was changed a bit to properly support optionally styling the foreground color - by default, it should be the same end result though, using the `BlendMode.Difference` with a `ColorFIlter.tint(...)` using 70% white -- this matches exactly what iOS is doing, from what I can tell.

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/appcues/SkippableTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/SkippableTrait.kt
@@ -1,45 +1,60 @@
 package com.appcues.trait.appcues
 
 import androidx.compose.foundation.background
+import androidx.compose.foundation.border
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.gestures.detectTapGestures
 import androidx.compose.foundation.interaction.MutableInteractionSource
+import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.size
-import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.ripple.rememberRipple
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
-import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.composed
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.draw.drawBehind
 import androidx.compose.ui.graphics.BlendMode
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.ColorFilter
+import androidx.compose.ui.graphics.Paint
+import androidx.compose.ui.graphics.PaintingStyle
+import androidx.compose.ui.graphics.Path
+import androidx.compose.ui.graphics.drawscope.DrawScope
+import androidx.compose.ui.graphics.drawscope.Stroke
+import androidx.compose.ui.graphics.drawscope.drawIntoCanvas
+import androidx.compose.ui.graphics.toArgb
 import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.Role
 import androidx.compose.ui.semantics.contentDescription
 import androidx.compose.ui.semantics.semantics
-import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import com.appcues.AppcuesCoroutineScope
 import com.appcues.R
 import com.appcues.data.model.AppcuesConfigMap
 import com.appcues.data.model.getConfig
 import com.appcues.data.model.getConfigOrDefault
+import com.appcues.data.model.getConfigStyle
+import com.appcues.data.model.styling.ComponentStyle.ComponentHorizontalAlignment
+import com.appcues.data.model.styling.ComponentStyle.ComponentVerticalAlignment
 import com.appcues.trait.BackdropDecoratingTrait
 import com.appcues.trait.ContainerDecoratingTrait
 import com.appcues.trait.ContainerDecoratingTrait.ContainerDecoratingType
-import com.appcues.trait.appcues.SkippableTrait.ButtonAppearance.Default
-import com.appcues.trait.appcues.SkippableTrait.ButtonAppearance.Hidden
-import com.appcues.trait.appcues.SkippableTrait.ButtonAppearance.Minimal
 import com.appcues.ui.ExperienceRenderer
+import com.appcues.ui.extensions.getBoxAlignment
+import com.appcues.ui.extensions.getColor
+import com.appcues.ui.extensions.getMargins
 import com.appcues.ui.extensions.xShapePath
+import com.appcues.util.ne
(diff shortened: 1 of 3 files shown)


appcues/src/main/java/com/appcues/ui/extensions/DrawScopeEx.kt
```diff
@@ -11,39 +9,25 @@ import java.lang.Float.max
...
-                start = pathRect.topLeft,
-                end = pathRect.bottomRight,
-                strokeWidth = strokeWidthPx,
-            )
-            drawLine(
-                color = color,
-                start = pathRect.bottomLeft,
-                end = pathRect.topRight,
-                strokeWidth = strokeWidthPx,
-            )
```
iujames, line 47: the issue with what was happening before here - this was actually drawing the Path, not just creating it! These `DrawScope.drawLine(...)` calls paint the path. I updated this to simply create and return the Path, and then let the callsite draw the path. That way, I could have fine control over what color and draw style get used, based on the styling options.

appcues/src/main/java/com/appcues/trait/appcues/SkippableTrait.kt
```diff
@@ -123,37 +160,171 @@ internal class SkippableTrait(
...
+                            style = drawStyle
+                        )
+                    } else {
+                        drawPath(
+                            path = it,
+                            color = Color.Black,
+                            colorFilter = ColorFilter.tint(Color(color = 0xB3FFFFFF), BlendMode.Difference),
+                            style = drawStyle
+                        )
+                    }
```
iujames, line 229: important part related to the `xShapePath` change - now I can draw the shadow first `drawPathShadow`, then draw the actual path, based on the given style properties. Previously, I could not get it to work, since `xShapePath` was actually drawing the path internally.

appcues/src/main/java/com/appcues/trait/appcues/SkippableTrait.kt
```diff
@@ -48,56 +63,78 @@ internal class SkippableTrait(
...
+        // the 'X' in a minimal style button uses a 1.5 dp stroke width for a 30 dp size button
+        // and scales accordingly with different sizes
+        private const val BUTTON_X_MINIMAL_STROKE_FACTOR = 1.5
 
-        object Hidden : ButtonAppearance(0.dp, 0.dp, 0.dp)
-        object Minimal : ButtonAppearance(4.dp, 30.dp, 8.dp)
-        object Default : ButtonAppearance(8.dp, 30.dp, 8.dp)
+        // the 'X' in a default style button uses a 2.0 dp stroke width for a 30 dp size button
+        // and scales accordingly with different sizes
+        private const val BUTTON_X_DEFAULT_STOKE_FACTOR = 2.0
```
iujames, line 87: these properties allow use to now scale with size of the X in a more consistent way with the symbol image / font size scaling used on iOS. The larger the button is, the more pronounced the X becomes. Also, on default, the X is a slightly heavier "weight" appearance, minimal it is thinner (2.0 vs 1.5 here)

appcues/src/main/java/com/appcues/ui/extensions/DrawScopeEx.kt
```diff
@@ -11,39 +9,25 @@ import java.lang.Float.max
...
     return Path()
         .apply {
-            val strokeWidthPx = strokeWidth.toPx()
             val pathSizePx = pathSize.toPx()
+            val sizeRect = size.toRect()
 
             // center the desired size within the current bounds
+            // find the delta to deflate the container rect in each direction
             val deltaX = max(size.width - pathSizePx, 0.0f) / 2
             val deltaY = max(size.height - pathSizePx, 0.0f) / 2
```
iujames, line 21: the centering detail here becomes important, as now the button height and width don't necessarily have to match - can have any rounded rect shape. We sill want the `X` to be visually centered within and kept in a square bounds itself. Prior to this with just a "deflate" option (previous PR) - you could end up with the `X` getting stretched on one axis.

appcues/src/main/java/com/appcues/trait/appcues/SkippableTrait.kt
```diff
@@ -123,37 +160,171 @@ internal class SkippableTrait(
...
-     * Apply different modifiers based on ButtonAppearance
-     */
-    private fun Modifier.drawSkippableButton(buttonAppearance: ButtonAppearance): Modifier {
+    // styles the close X button differently based on the appearance set in config
+    private fun Modifier.styleButton(isDark: Boolean): Modifier {
         return this.then(
             when (buttonAppearance) {
-                Hidden -> Modifier
-                Default ->
+                ButtonAppearance.HIDDEN -> Modifier
```
iujames, line 167: looking at this now, the HIDDEN case should be hoisted up even higher I think.  In `DecorateContainer`, it should not even add the `Spacer` at all, if HIDDEN, right?

iujames: updated at top level to not include the `Spacer` when `hidden`

andretortolano: I think before that I was basically not taking space, but this works better.

appcues/src/main/java/com/appcues/trait/appcues/SkippableTrait.kt
andretortolano: yes, given the customizations required for this new version it makes sense to change the xShapePath for better flexbility

appcues/src/main/java/com/appcues/ui/extensions/DrawScopeEx.kt
andretortolano: interesting fact, i had color being passed down twice basically because of this, I dont remember exactly how to tried to draw it using the lineTo but I had no success for some reason. Nice work here

andretortolano (approved):
Nice improvement over the existing skippableTrait, what I liked the most was the separation in various Modifier methods

PR #370, opened 2023-03-30
Refactor step container render into a testable function for snapshot testing

iujames:
The `CompositionTestHelper.ComposeContainer` is a debug only public helper that allows our snapshot test tools (different project) to send in experience and trait JSON data and render a step for snapshot purposes. It is similar in concept to how the existing `ComposeContent` works for primitives, but supports a more advanced case with the full experience and trait decoding and mapping. The given trait is injected into the specified step, for testing. This allows our test cases to have some base experience JSON and then add variations of traits as needed. (test updates will be in another PR in that repo)

The change in `AppcuesComposition` is just a refactor to make the render of a step container a testable piece with limited  dependencies, and internal visibility, in support of this new test helper.

### DIFF
--- a/appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
+++ b/appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
@@ -1,20 +1,32 @@
 package com.appcues
 
+import android.content.Context
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
 import coil.ImageLoader
 import com.appcues.action.ExperienceAction
 import com.appcues.analytics.ExperienceLifecycleEvent.StepInteraction.InteractionType
 import com.appcues.data.MoshiConfiguration
+import com.appcues.data.mapper.experience.ExperienceMapper
 import com.appcues.data.mapper.step.primitives.mapPrimitive
 import com.appcues.data.model.ExperienceStepFormState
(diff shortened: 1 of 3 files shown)


appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
```diff
@@ -35,6 +47,65 @@ fun ComposeContent(json: String, imageLoader: ImageLoader) {
     }
 }
 
+@Composable
+fun ComposeContainer(
+    context: Context,
+    experienceJson: String,
+    traitJson: String,
```
iujames, line 54: considering having this take a list of trait JSON to apply, but I think for now it's not needed and we can always expand that later

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
```diff
@@ -156,9 +163,9 @@ private fun BoxScope.ComposeLastRenderingState(state: Rendering) {
 
 @Composable
 private fun rememberMetadataSettingTraits(
-    viewModel: AppcuesViewModel,
     metadataSettingTraits: MutableState<List<MetadataSettingTrait>>,
 ): MutableState<AppcuesStepMetadata>? {
+    val traitExceptionHandler = LocalAppcuesTraitExceptionHandler.current
```
iujames, line 168: using this new trait exception handler, just a lambda that handles the exception, simplifies things so that the entire `AppcuesViewModel` does not need to be passed all the way down or used at all in testing scenarios.

andretortolano: sure, I guess we had discussed this option when we decided to capture exception before

andretortolano (approved):

PR #371, opened 2023-03-30
Fix crash when applying border to tooltip

iujames:
I was hacking around on some testing items and came across what seems to be the relatively simple fix for the crash seen when applying a border to a tooltip trait style. It simply needs to guard against an empty path during some layout passes. 

I'm guessing there is a first pass composition running before the tooltip size is first set and the `containerDimens` are fully known. In the `tooltipPath` helper, if no `containerDimens` are set yet, it does not draw any path. Later, if using this empty path when drawing a border, a crash is observed:

```
    java.lang.IllegalArgumentException: width and height must be > 0
    	at android.graphics.Bitmap.createBitmap(Bitmap.java:1095)
    	at androidx.compose.ui.graphics.Api26Bitmap.createBitmap-x__-hDU$ui_graphics_release(AndroidImageBitmap.android.kt:198)
    	at androidx.compose.ui.graphics.AndroidImageBitmap_androidKt.ActualImageBitmap-x__-hDU(AndroidImageBitmap.android.kt:44)
    	at androidx.compose.ui.graphics.ImageBitmapKt.ImageBitmap-x__-hDU(ImageBitmap.kt:255)
    	at androidx.compose.ui.graphics.ImageBitmapKt.ImageBitmap-x__-hDU$default(ImageBitmap.kt:249)
    	at androidx.compose.foundation.BorderKt.drawGenericBorder(Border.kt:474)
    	at androidx.compose.foundation.BorderKt.access$drawGenericBorder(Border.kt:1)
    	at androidx.compose.foundation.BorderKt$border$2$1.invoke(Border.kt:119)
    	at androidx.compose.foundation.BorderKt$border$2$1.invoke(Border.kt:100)
```

checking `!path.isEmpty` can avoid that and render as expected:

Screenshot 2023-03-30 at 1 14 21 PM

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
@@ -191,7 +191,8 @@ internal class TooltipTrait(
         path: Path,
         isDark: Boolean
     ) = this.then(
-        if (style?.borderWidth != null && style.borderWidth ne 0.0 && style.borderColor != null) {
+        @Suppress("ComplexCondition")
+        if (style?.borderWidth != null && style.borderWidth ne 0.0 && style.borderColor != null && !path.isEmpty) {
             Modifier
                 .border(style.borderWidth.dp, style.borderColor.getColor(isDark), GenericShape { _, _ -> addPath(path) })
                 .padding(style.borderWidth.dp)


appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
```diff
@@ -191,7 +191,8 @@ internal class TooltipTrait(
         path: Path,
         isDark: Boolean
     ) = this.then(
-        if (style?.borderWidth != null && style.borderWidth ne 0.0 && style.borderColor != null) {
+        @Suppress("ComplexCondition")
+        if (style?.borderWidth != null && style.borderWidth ne 0.0 && style.borderColor != null && !path.isEmpty) {
```
iujames, line 195: I considered other ways to rework this function but all of them led to what felt like an overall less readable solution than just the pretty simple 4 case conditional check here, so, I suppressed the Detekt warning (which wants 3 conditions max)

andretortolano: interesting, the isEmpty might be just what we need here to be real, what happens is that the first pass ever the containerDimens is empty. I did not find any way to go around this yet. we wait until the "content" returns a size to update containerDimens.value, which re-composes all these modifiers.

andretortolano (approved):
nice catch here.

PR #372, opened 2023-03-30
Improve error handling when reading target element trait from config

iujames:
This is a small, but useful improvement to how the target element trait reads its selector from the config JSON, during the mapping of an experience. This improves on the initial work in #363 .

I was really scratching my head and confusing myself earlier when working on React Native, as an experience I was using to test with was silently failing. No error for invalid selector, not even an error for JSON decoding issues. What was actually happening was that it was failing during the experience mapping, when the trait was trying to decode the config, and the exception thrown was silent since our `CoroutineExceptionHandler` catches (and optionally logs to logcat).

This change will make it so that the config props are read as a simple Map<String, String> rather than using the moshi adapter to try to decode the object. Then, the failure happens later when the selector is inflated and we can properly log the error. This way we get meaningful StepError information to diagnose things if a bad selector is ever encountered.

So instead of a silent error, we now get
Screenshot 2023-03-30 at 3 25 18 PM

This also matches how it works on iOS for this type of error logging.

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/appcues/TargetElementTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/TargetElementTrait.kt
@@ -4,11 +4,9 @@ import androidx.core.graphics.Insets
 import androidx.core.view.ViewCompat
 import androidx.core.view.WindowInsetsCompat
 import com.appcues.Appcues
-import com.appcues.ElementSelector
 import com.appcues.ViewElement
 import com.appcues.data.model.AppcuesConfigMap
 import com.appcues.data.model.getConfig
-import com.appcues.data.model.getConfigObject
 import com.appcues.data.model.getConfigOrDefault
 import com.appcues.debugger.screencapture.toDp
 import com.appcues.findMatches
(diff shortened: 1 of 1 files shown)


mmaatttt:
Ah I noticed this when testing sc-44389. Glad you're on top of it!

appcues/src/main/java/com/appcues/trait/appcues/TargetElementTrait.kt
```diff
@@ -24,7 +22,7 @@ internal class TargetElementTrait(
         const val TYPE = "@appcues/target-element"
     }
 
-    private val selector: ElementSelector? = config.getConfigObject("selector")
+    private val selectorProperties = config.getConfig<Map<String, String>>("selector") ?: mapOf()
```
andretortolano, line 25: was getConfigObject failing silently? so by changing to getConfig (possible null) and checking later during the viewMatchingSelector is whats doing the trick?

iujames: Yes, `getConfigObject` ended up failing in the Moshi deserialization layer, right here https://github.com/appcues/appcues-android-sdk/blob/main/appcues/src/main/java/com/appcues/data/remote/appcues/adapters/ElementSelectorAdapter.kt#L16

That exception would get thrown during the experience mapping process. The entire response would then effectively be thrown out, as that error would not be handled and propagated all the way out to the root level `CoroutineExceptionHandler`. To the end user it was just a silent failure if the `selector` on the trait had any unexpected structure, which was hard to diagnose.

PR #373, opened 2023-03-31
✨ Added property pointerCornerRadius for tooltip trait

andretortolano:
adding cornerRadius to the previous tooltip structure required some changes and a lot of testing. part of the effort here was to keep it organized so we can come back to it later if needed.

TooltipPath now contains all the Path related methods, auxiliary classes and methods were moved to TooltipPathExt for lack of a better name. with this change we are supporting cornerRadius on tooltips including some animations, but there are still some remaining work to safe guard odd scenarios (I will create a separate ticket for this one after)

https://user-images.githubusercontent.com/5244805/231225409-9165106c-a2d0-4343-87e7-2bccd21982f6.mov

tagging @mmaatttt here since he helped quite a lot with the trigonometry involved, + I know he wanted to know how this will be merged on android to compare implementation on the web side

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/appcues/TooltipPathExt.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/TooltipPathExt.kt
@@ -1,20 +1,9 @@
 package com.appcues.trait.appcues
 
-import androidx.compose.animation.core.AnimationSpec
-import androidx.compose.animation.core.animateDpAsState
-import androidx.compose.animation.core.animateFloatAsState
-import androidx.compose.foundation.shape.RoundedCornerShape
-import androidx.compose.runtime.Composable
+import android.graphics.PointF
 import androidx.compose.runtime.State
 import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
 import androidx.compose.ui.geometry.Offset
-import androidx.compose.ui.geometry.Size
-import androidx.compose.ui.graphics.Path
-import androidx.compose.ui.graphics.PathOperation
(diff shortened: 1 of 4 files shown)


appcues/src/main/java/com/appcues/trait/appcues/TooltipPathExt.kt
```diff
@@ -166,6 +160,7 @@ internal fun getTooltipSettings(
     distance: Dp,
     pointerBaseDp: Dp,
     pointerLengthDp: Dp,
+    pointerCornerRadius: Dp = 0.dp,
```
iujames, line 163: do we need a default value here any longer? I think it's always passed in now.

iujames (approved):
awesome! I'm certainly not the one to check the math here, but in my testing it out it is working quite well with the known limitations:
* will crash if container cornerRadius is too large for the container
* will crash if when a pointer corner radius is applied to a large width causing it to become larger than container
maybe those two things are really related - unable to position a pointer in a container when the pointer width does not fit

also
* a really large pointer length, larger than will fit, will cause the tooltip to disappear. I think this may be the subject of [other ticket](https://app.shortcut.com/appcues/story/48999/android-update-tooltip-to-fit-in-visible-area)

in terms of the organization of the code, looks great to me!

appcues/src/main/java/com/appcues/trait/appcues/TooltipPathExt.kt
andretortolano: nice catch, not being used anymore

PR #376, opened 2023-04-12
Add support for snapshot testing multiple traits with optional variable step content

iujames:
This adjustment will allow our UI test library to snapshot test combinations of traits, for example: tooltip + target-rectangle. It will also allow optionally overriding the step content with another predefined block of content json.

The test updates will be coming in the spec repo, but these allow for a test that would use a combination like this example
```kotlin
        snapshotTraits(
            composeTestRule,
            listOf(
                "appcues-backdrop",
                "appcues-target-rectangle-relative",
                "appcues-tooltip",
                "appcues-backdrop-keyhole-rectangle",
            ),
            "text-default",
        )
```

To produce an image like
com appcues host TooltipTraitTests_testDefault

### DIFF
--- a/appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
+++ b/appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
@@ -26,10 +26,13 @@ import com.appcues.ui.composables.LocalAppcuesActionDelegate
 import com.appcues.ui.composables.LocalExperienceStepFormStateDelegate
 import com.appcues.ui.composables.LocalImageLoader
 import com.appcues.ui.composables.LocalLogcues
+import com.appcues.ui.composables.isBackdropVisible
 import com.appcues.ui.composables.isContentVisible
 import com.appcues.ui.primitive.Compose
 import com.appcues.ui.theme.AppcuesTheme
 
+// This helper supports testing a single experience primitive (can have nested children),
+// defined in the given json string.
 @Composable
 fun ComposeContent(json: String, imageLoader: ImageLoader) {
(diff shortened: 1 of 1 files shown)


appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
```diff
@@ -47,33 +50,36 @@ fun ComposeContent(json: String, imageLoader: ImageLoader) {
...
+// optional content is injected into the single step, if present.
 @Composable
-fun ComposeContainer(
-    context: Context,
-    experienceJson: String,
-    traitJson: String,
-    groupIndex: Int,
-    stepIndex: Int,
-    imageLoader: ImageLoader)
+fun ComposeContainer(context: Context, stepContentJson: String?, traitJson: List<String>, imageLoader: ImageLoader)
```
andretortolano, line 57: I'm probably not getting it right, but I think that what @mmaatttt did [here](https://github.com/appcues/appcues-mobile-experience-spec/pull/168/files) is a bit different, it seems like traits where should refer to a different file in a source folder, that would contain the entire json for that trait name.

iujames: in this case, `traitJson` is a list of JSON objects, one per trait, to be injected. They get deserialized and mapped down on line 65 below. Does that clarify?

andretortolano: oh yeah sorry, after this one we need to update the ScreenshotTestExt.kt to support calling traits from separate files then

andretortolano (approved):

PR #380, opened 2023-04-18
♻️ Adjusting tooltip composition to fix pointer positioning

andretortolano:
the way the tooltip is built at the moment sometimes gives us unexpected results, where the pointer is in the wrong position and/or the content is not visible.

Some adjustments to align positioning logic between android and ios.

* TargetRectangleInfo?.getTooltipPointerPosition(..) we added distance as one parameter that can influence available space, and changed the containerDimens with contentDimens.
* Added contentDimens that is calculated based on existing containerDimens and known tooltip position.
* Fixed tooltipSize to mirror rules from IOS
* general improvement on how the composition is organized (fixing some of the detekt warnings)

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
@@ -8,10 +8,8 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.requiredHeightIn
-import androidx.compose.foundation.layout.requiredWidthIn
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.shape.GenericShape
@@ -19,6 +17,7 @@ import androidx.compose.foundation.verticalScroll
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.State
+import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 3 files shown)


appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
```diff
@@ -330,38 +353,61 @@ internal class TooltipTrait(
...
-        return remember(targetRectangleInfo, containerDimens) {
-            val position = targetRectangleInfo.getTooltipPointerPosition(
+        pointerLength: Dp,
+        contentSized: Boolean,
+    ): TooltipPointerPosition {
+        // pointer position is only calculated once for a given target rectangle, after the container
+        // is sized. The tooltip may end up getting a new container layout if it shifts the pointer
+        // to another edge (not preferred edge) based on this calculation, but we stick with that
+        // calculated position from that point forward and don't allow it to flip back
+        return remember(targetRectangleInfo, contentSized) {
```
iujames, line 390: I'm not sure if this remember is 1000% valid, but maybe good enough. Since `contentSized` gets set to true once the first step runs its layout and sets a size - but what if the next step uses the exact same `targetRectangleInfo`? Maybe that is not possible - but what I'm trying to get at: every step should run this pointer position code once, picking the desired position for that step, and then it only resets when a new step comes through (or some other significant relayout happens like an orientation change?)

iujames (approved):
Have a bit of uncertainty about the one remember I added, noted below. Can you also change the gitmoji from 🚧 to ♻️ I think? since this change is no longer WIP.

appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
andretortolano: its not intuitive but the targetRectangle cant be the same for two steps I dont think, but in any case, if both steps would share the same, then the contentSized would trigger again when the content size changes

iujames: I don't think `contentSized` would trigger again though would it? since it was already sized and moved from `false` -> `true` on the first ever layout. it would never change again (since never `null` again)

andretortolano: since contentDimens is a key to a remember in that block, when steps change, the size will most likely change as well, and it will cause the entire block to recompose. thats my assumption

PR #381, opened 2023-04-18
Handle un-targeted tooltip layout and transitions to and from

iujames:
This update will handle the layout of any tooltip without a target rect - setting width to 400dp (or max for screen), and removing any pointer parameters in the draw. It also improves a little around the logic when transitioning from one step to the next, ensuring the proper `contentDimens` are used when running the next calculation on where to position the pointer.

End result should be a pretty robust handling of un-targeted tips anywhere in the sequence of tooltip steps, showing Android (left) and iOS (right) side-by-side here.
https://user-images.githubusercontent.com/19266448/232903037-8177d99d-d9a1-47e4-a006-f5efd6d0638e.mov

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
@@ -79,6 +79,8 @@ internal class TooltipTrait(
 
         private const val POINTER_BASE_DEFAULT = 16.0
         private const val POINTER_LENGTH_DEFAULT = 8.0
+
+        private val MAX_TOOLTIP_WIDTH = 400.0.dp
     }
 
     private val style = config.getConfigStyle("style")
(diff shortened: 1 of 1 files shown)


appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
```diff
@@ -113,7 +115,7 @@ internal class TooltipTrait(
         val containerDimens = remember { mutableStateOf<TooltipContainerDimens?>(null) }
         // this is like containerDimens, but removing the size of the pointer from the width or height,
         // depending on the pointer position - used in layout calculation of desired pointer position
-        val contentDimens = remember { mutableStateOf<TooltipContainerDimens?>(null) }
+        val contentDimens = remember(targetRectInfo) { mutableStateOf<TooltipContainerDimens?>(null) }
```
iujames, line 118: this update to the remember allows the next pointer position check to run only after the new content size is found for the new step

appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
```diff
@@ -154,7 +156,7 @@ internal class TooltipTrait(
                 ) {
                     Box(
                         modifier = Modifier
-                            .tooltipSize(style, windowInfo, tooltipSettings)
+                            .tooltipSize(style, windowInfo, tooltipSettings, targetRect == null)
```
iujames, line 159: `targetRect == null` is how we know it is "bottom toast" style now

andretortolano (approved):

PR #383, opened 2023-04-19
♻️  limit pointer width based on tooltip contraints

andretortolano:
safeguarding pointer base values based on content available space to position the pointer.

also as part of this task, after calculating how much space the pointer base will take, we compare with the available space to figure out if we should round the base corners of the pointer. (matching IOS behavior)

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/appcues/TooltipPathExt.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/TooltipPathExt.kt
@@ -135,6 +135,7 @@ internal data class TooltipSettings(
     val pointerBasePx: Float,
     val pointerLengthPx: Float,
     val pointerCornerRadiusPx: Float,
+    val isRoundingBase: Boolean,
 ) {
 
     val pointerBaseCenterPx = pointerBasePx / 2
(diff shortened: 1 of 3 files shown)


appcues/src/main/java/com/appcues/trait/appcues/TooltipPathExt.kt
```diff
@@ -317,18 +320,18 @@ fun calculateTooltipWidth(base: Float, length: Float, radius: Float, roundedBase
     //
     // (3) multiply by two to account for the two sides of the pointer
     //     and add the existing width to the end result
-    return (tan(angle) * radius * roundedBase) + base
+    return (tan(angle) * radius * 2) + base
 }
 
-data class CornerPoint(
+internal data class CornerPoint(
```
iujames, line 326: nice catch on internals, we can generate the dokka docs perhaps on main to make sure nothing unexpected got into public api before release

iujames (approved):

appcues/src/main/java/com/appcues/trait/appcues/TooltipPathExt.kt
andretortolano: nice todo before 2.0

PR #384, opened 2023-04-19
Account for the pointer length in the horizontal tooltip padding start calculation

iujames:
These updates are to ensure that the desired target width gets satisfied and the start padding is not too large, in some horizontal tooltip cases. It was observed in tests that content positioned to left (pointing to right) could sometimes be compressed in width by the pointer length, leading to a narrow than expected tooltip.


andretortolano (approved):
🚀

PR #386, opened 2023-04-20
📝 Updating docs for 2.0 release

andretortolano:
Updating docs and revising method/class visibility (from public to internal)

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/composables/CompositionLocals.kt
+++ b/appcues/src/main/java/com/appcues/ui/composables/CompositionLocals.kt
@@ -63,6 +63,7 @@ internal val LocalStackScope = compositionLocalOf<StackScope> { ColumnStackScope
 internal val LocalChromeClient = compositionLocalOf { AccompanistWebChromeClient() }
 
 internal val LocalAppcuesTraitExceptionHandler = compositionLocalOf { AppcuesTraitExceptionHandler {} }
+
 internal data class AppcuesTraitExceptionHandler(val onTraitException: (AppcuesTraitException) -> Unit)
 
 internal sealed class StackScope(private val childrenCount: Int) {
@@ -94,11 +95,23 @@ internal sealed class StackScope(private val childrenCount: Int) {
     class ColumnStackScope(val width: Double?, childrenCount: Int) : StackScope(childrenCount)
 }
 
+/**
+ * Use LocalAppcuesStepMetadata to access the shared information between steps
+ *
+ * Example usage:
+ * @sample com.appcues.trait.appcues.BackdropKeyholeTrait
+ */
 val LocalAppcuesStepMetadata = compositionLocalOf<AppcuesStepMetadata> { noLocalProvidedFor("LocalAppcuesStepMetadata") }
 
+/**
+ * AppcuesStepMetadata used to share trait information between traits, generated from produceMetadata.
+ *
+ * Previous and current refer to the previously generated metadata values and the current values.
+ * It may be desired to reference both values to support transitions between them.
+ */
 data class AppcuesStepMetadata(
     val previous: Map<String, Any?> = hashMapOf(),
-    val actual: Map<String, Any?> = hashMapOf()
+    val current: Map<String, Any?> = hashMapOf()
 )
 
 private fun noLocalProvidedFor(name: String): Nothing {
--- a/docs/Traits.md
+++ b/docs/Traits.md
@@ -4,9 +4,16 @@ Experiences displayed by the Appcues Android SDK are customizable and extensible
 
 ## Overview
 
-An ``ExperienceTrait`` modifies the how an entire experience, or a particular step in an experience is displayed. 
-Each type of ``ExperienceTrait`` has the capability to apply specific changes to the rendered content in an experience.
-Also, some traits are applied to the whole experience, some to a set of steps, and others are specific to a step.
+An `ExperienceTrait` modifies the how an entire experience, or a particular step in an experience is displayed. 
+A trait of a specific _type_ has _capabilities_ that modify the way an experience is displayed to the user. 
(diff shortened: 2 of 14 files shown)


docs/VersionMigration.md
iujames, line 1: lets name this one `2.0VersionMigration.md` - since we may have more of these in the future

andretortolano: I was thinking we could use the same file and add like migration instructions on the top for newer versions

appcues/src/main/java/com/appcues/ui/composables/CompositionLocals.kt
```diff
@@ -94,8 +95,18 @@ internal sealed class StackScope(private val childrenCount: Int) {
...
+ * Use LocalAppcuesStepMetadata to access the shared information between steps
+ *
+ * Example usage:
+ * @sample com.appcues.trait.appcues.BackdropKeyholeTrait
+ */
 val LocalAppcuesStepMetadata = compositionLocalOf<AppcuesStepMetadata> { noLocalProvidedFor("LocalAppcuesStepMetadata") }
 
+/**
+ * AppcuesStepMetadata used to share trait information between traits, generated from produceMetadata.
+ * Previous and actual refers to the steps inside of the step group
```
iujames, line 108: Wondering if this could be more clear? `previous` refers to the previously known metadata values that could be used for animating from a previous state, is that correct? And `actual` - wondering if it should be renamed `current`? That is the current latest known metadata values, correct?

Theoretically, it is not tied to the step though right? Meaning, a single step might have the ability to have metadata updated multiple times. I know this is not really how it is used right now, its more about step transitions, but in theory, the metadata can update and publish at any time - say for instance orientation changes, or some timed behaviors

docs/VersionMigration.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Overview
+
+There are a number of breaking changes from 1.x to 2.0, mostly related to the experience trait system. If you haven't implemented any custom actions or traits, the migration will involve little or no changes in your code.
+
+## General Changes
+
+### Removed
+
+- Properties are no longer supported for anonymous users. `Appcues.anonymous(properties: Map<String, Any>?)` has been superseded by ``Appcues.anonymous()``. If you require properties, use ``Appcues.identify(userId: String, properties: Map<String, Any>? = null)``.
```
iujames, line 11: some of these have double back ticks `` which I don't think are needed here? That may have been a carry over from DocC stuff on iOS side where it creates links?

docs/VersionMigration.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Changed
+
+- ``AppcuesNavigationDelegate`` is now responsible for handling universal links in addition to scheme links. Consequently, ``AppcuesNavigationDelegate/navigate(to:openExternally:completion:)`` has a new parameter `openExternally` which indicates whether a link is expecting to be opened in an external browser.
```
iujames, line 15: this is the one we chatted about removing for Android side

iujames (commented):
couple iOS specific things I think we want to remove first

appcues/src/main/java/com/appcues/ui/composables/CompositionLocals.kt
andretortolano: maybe the wording is not precise indeed, but the metadata is tied to the composition, it refreshes on step change, but will also re-create, update on orientation change like you said its just not something controlled by the trait that is implementing MetadataTrait.

docs/VersionMigration.md
andretortolano: It doesnt change anything, I will clean this one up.

docs/VersionMigration.md
andretortolano: shoot, forgot to remove the line 👎

appcues/src/main/java/com/appcues/ui/composables/CompositionLocals.kt
```diff
@@ -94,11 +95,22 @@ internal sealed class StackScope(private val childrenCount: Int) {
...
+ *
+ * Example usage:
+ * @sample com.appcues.trait.appcues.BackdropKeyholeTrait
+ */
 val LocalAppcuesStepMetadata = compositionLocalOf<AppcuesStepMetadata> { noLocalProvidedFor("LocalAppcuesStepMetadata") }
 
+/**
+ * AppcuesStepMetadata used to share trait information between traits, generated from produceMetadata.
+ * previous and current refers to the the previous generated metadata and current as sometimes you may want to
+ * consult both values to enable for nice transition between the two.
```
iujames, line 109: Suggestion: 
> Previous and current refer to the previously generated metadata values and the current values. It may be desired to reference both values to support transitions between them.

iujames (approved):

iujames (approved):
Looking good. I think the one item left to review (doesn't have to be in this PR) is the Traits doc here https://github.com/appcues/appcues-android-sdk/blob/main/docs/Traits.md to carry over any related updates similar to the iOS update here https://github.com/appcues/appcues-ios-sdk/commit/259823a24f94aec06d5806138bdcfa697c6058d8

docs/Traits.md
```diff
@@ -84,8 +101,52 @@ In practice this distinction looks like this in the experience data model:
...
                 // Step-level traits for the second step
+                // Presenting and Creating traits are ignored
+                // Decorating traits only
             ]
         }
     ]
 }
 ```
+
+Only a single instance of each trait type may be included in the `traits` array at any given level. If multiple traits with the same `id` are included in the same array, the experience is invalid, an error will be logged, and the experience will not display.
```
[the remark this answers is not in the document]
iujames: iOS side fixup https://github.com/appcues/appcues-ios-sdk/pull/393

PR #387, opened 2023-04-21
Fix handling of pointerBase larger than available space

iujames:
This fixes the issue where the pointerBase max was causing the pointer alignment to be incorrect in some cases. Validated with snapshot tests coming. Turned out to just be a simple off by one case in calculating the pointer x/y offsets.

| After | Before |
| --- | --- |
| com appcues host TooltipTraitTests_testPointerBaseGreaterThanContentTop | com appcues host TooltipTraitTests_testPointerBaseGreaterThanContentTop |


andretortolano (approved):
Nice, thanks for tracking this one down!

PR #388, opened 2023-04-21
🐛 fixing animated values related to positioning the tooltip

andretortolano:
did some small change related to the padding modifier, in order to keep the value of the animated padding values even when contentDimens is not available.

### DIFF
--- a/gradle/tools/detekt/detekt-config.yml
+++ b/gradle/tools/detekt/detekt-config.yml
@@ -116,6 +116,7 @@ complexity:
   LongMethod:
     active: true
     threshold: 60
+    ignoreAnnotated: [ 'Composable' ]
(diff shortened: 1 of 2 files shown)


gradle/tools/detekt/detekt-config.yml
```diff
@@ -116,6 +116,7 @@ complexity:
   LongMethod:
     active: true
     threshold: 60
+    ignoreAnnotated: [ 'Composable' ]
```
andretortolano, line 119: little change to detekt-config rules since I've been fighting this rule for no reason on Composable methods for a while now.

I guess it makes sense most of the time but for composable we can ignore since things can get bigger and sometimes breaking into smaller methods might not be the best

iujames (approved):
nice!

PR #391, opened 2023-04-27
💥  Removed Appcues param from action.execute method

andretortolano:
At the moment Appcues instance is not required on ExperienceAction.execute call, when necessary we can let it be declared as a constructor of the implementing class.

This changes removes appcues from the execute method


iujames (approved):
Looks good - I think there are just some unit tests that need updated to match, to build and test successfully.

PR #392, opened 2023-04-27
✨ Added target-interaction trait

andretortolano:
This adds an implementation of @appcues/target-interaction (https://github.com/appcues/appcues-mobile-experience-spec/pull/175). This trait can be used to create a transparent View that lays over the target rectangle, from metadata, to captures taps and long presses. 

Additionally this trait can also provide with a list of actions that should be executed when those actions occur.

```json
{
    "type": "@appcues/target-interaction",
    "config": {
        "actions": [
            {
                "on": "tap",
                "type": "@appcues/continue",
                "config": {
                    "offset": 1
                }
            }
        ]
    }
}
```

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/appcues/TargetInteractionTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/TargetInteractionTrait.kt
@@ -1,0 +1,89 @@
+package com.appcues.trait.appcues
+
+import androidx.compose.foundation.ExperimentalFoundationApi
+import androidx.compose.foundation.combinedClickable
+import androidx.compose.foundation.gestures.detectTapGestures
+import androidx.compose.foundation.interaction.MutableInteractionSource
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.BoxScope
+import androidx.compose.foundation.layout.offset
+import androidx.compose.foundation.layout.size
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.input.pointer.pointerInput
+import androidx.compose.ui.unit.DpOffset
+import androidx.compose.ui.unit.DpSize
+import androidx.compose.ui.unit.dp
+import com.appcues.action.ActionProcessor
+import com.appcues.action.ActionRegistry
+import com.appcues.action.ExperienceAction
+import com.appcues.analytics.ExperienceLifecycleEvent.StepInteraction.InteractionType
+import com.appcues.analytics.ExperienceLifecycleEvent.StepInteraction.InteractionType.TARGET_LONG_PRESSED
+import com.appcues.analytics.ExperienceLifecycleEvent.StepInteraction.InteractionType.TARGET_TAPPED
+import com.appcues.data.model.AppcuesConfigMap
+import com.appcues.data.model.getConfigActions
+import com.appcues.trait.BackdropDecoratingTrait
+import com.appcues.trait.extensions.getRect
+import com.appcues.trait.extensions.rememberTargetRectangleInfo
+import com.appcues.ui.composables.AppcuesActionsDelegate
+import com.appcues.ui.composables.LocalAppcuesStepMetadata
+import com.appcues.ui.extensions.toLongPressMotionOrNull
+import com.appcues.ui.extensions.toTapMotionOrEmpty
+import com.appcues.ui.utils.rememberAppcuesWindowInfo
+
+internal class TargetInteractionTrait(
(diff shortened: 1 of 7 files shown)


appcues/src/main/java/com/appcues/data/model/ModelExtensions.kt
```diff
@@ -54,6 +57,11 @@ internal fun AppcuesConfigMap.getConfigPrimitive(key: String): ExperiencePrimiti
     }
 }
 
+internal fun AppcuesConfigMap.getConfigActions(key: String, actionRegistry: ActionRegistry): List<Action> {
+    return getConfig<List<Any>>(key)?.map { MoshiConfiguration.fromAny<ActionResponse>(it) }
+        ?.mapNotNull { response -> response?.toAction(actionRegistry) } ?: arrayListOf()
+}
```
iujames, line 63: This seems fine, just noting that this `ModelExtensions.kt` file has evolved to really just be stuff for `AppcuesConfigMap` used during trait config decoding right? And this one in particular is very specific to a single trait that needs it. It could just be included in the new `TargetInteractionTrait.kt` where it is being used, as an alternative. I don't feel strongly, just made me ponder the organization of this stuff - checking if you have any opinions.

appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
```diff
@@ -95,18 +96,22 @@ internal class ExperienceMapper(
...
-                    add(LinkAction(
-                        redirectUrl = it,
-                        linkOpener = scope.get()
-                    ))
+                    add(
+                        LinkAction(
+                            redirectUrl = it,
+                            linkOpener = scope.get()
+                        )
+                    )
```
iujames, line 104: Looks like this will have some sort of conflict with https://github.com/appcues/appcues-android-sdk/pull/391 perhaps, depending which goes first.

appcues/src/main/java/com/appcues/trait/appcues/TargetInteractionTrait.kt
```diff
@@ -0,0 +1,91 @@
...
+    override fun BoxScope.BackdropDecorate(content: @Composable BoxScope.() -> Unit) {
+        val targetRectInfo = rememberTargetRectangleInfo(LocalAppcuesStepMetadata.current)
+
+        val rect = targetRectInfo.getRect(rememberAppcuesWindowInfo()) ?: Rect(Offset(0f, 0f), Size(0f, 0f))
+        val sizeDp = DpSize(rect.width.dp, rect.height.dp)
+        val offsetDp = DpOffset(rect.left.dp, rect.top.dp)
+
+        // calling content before our composable makes so we are on top of all other backdrop traits
+        // wrapping content call
+        content()
```
iujames, line 70: ah ok was wondering how that detail was handled

appcues/src/main/java/com/appcues/trait/appcues/TargetInteractionTrait.kt
```diff
@@ -0,0 +1,91 @@
...
+                            indication = null,
+                            onLongClick = actions.toLongPressMotionOrNull(actionDelegate, TARGET_LONG_PRESSED, VIEW_DESCRIPTION),
+                            onClick = actions.toTapMotionOrEmpty(actionDelegate, TARGET_TAPPED, VIEW_DESCRIPTION),
+                        )
+                    }
+                    // when no actions are listed, the default behavior is to
+                    else Modifier.pointerInput(Unit) { detectTapGestures { } }
+                )
+        )
+    }
```
iujames, line 90: can this composition be avoided entirely in the case of no target rect in metadata?

iujames (approved):
looking good! couple minor notes below

appcues/src/main/java/com/appcues/data/model/ModelExtensions.kt
andretortolano: maybe renaming to AppcuesConfigMapExt would make more sense, but I think its ok to leave those methods grouped up and available for other traits, even when its very specific they have similar implementation

appcues/src/main/java/com/appcues/trait/appcues/TargetInteractionTrait.kt
andretortolano: thats a good catch, did some changes here that might cover this scenario

appcues/src/main/java/com/appcues/data/model/ModelExtensions.kt
iujames: sounds good, yeah agree rename would be fine if you'd like.

PR #399, opened 2023-05-03
Update the introduction for 2.0 migration doc

iujames:
This is a requested change from Chris, to lead with the positive/common case first.


andretortolano (approved):
nice catch

PR #400, opened 2023-05-03
Remove AppcuesActivity and use AppcuesOverlayViewManager for modals

iujames:
This turned out to be a very easy change to try to standardize approach between modals and tooltips. Want to be sure to think this through and test everything out, but posting the code changes here for any input.


andretortolano (approved):
I'm ok with this change, seems to be working ok. 

there is a method in AppcuesOverlayViewManager.stop can be set to private if that makes sense, not sure if we will ever use it outside of that class

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
```diff
@@ -32,19 +32,25 @@ internal class AppcuesOverlayViewManager(private val scope: Scope) : DefaultLife
...
-        viewModel?.onPause()
         shakeGestureListener?.stop()
     }
 
-    fun start() {
+    fun start(): Boolean {
+        val activity = AppcuesActivityMonitor.activity
+
+        // inform the caller that we cannot currently add a view in the app
+        if (activity == null || AppcuesActivityMonitor.isPaused) {
```
andretortolano, line 47: thats an interesting state to safe guard against, the experience showing while the activity is paused would surely cause a problem

andretortolano (approved):

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
iujames: yes, it essentially leaves the state machine in a hung state - as it cannot progress from BeginningStep to RenderingStep, the UI never moves it forward.

PR #401, opened 2023-05-03
Enable explicit API mode

iujames:
https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors

This feature will help us avoid any accidental leakage of items into the public interface of the SDK in the future.


andretortolano (approved):
is everything correct here for CompositionTestHelper.kt ? not sure if it is only scanning for problems in the current variant, wanted to doublecheck

iujames:
> is everything correct here for CompositionTestHelper.kt

@andretortolano can you clarify what you mean there?

PR #402, opened 2023-05-03
Update Detekt config to require comments on public API

iujames:
stacks on #401 

enable Detekt [comments rules](https://detekt.dev/docs/rules/comments)
* OutdatedDocumentation
* UndocumentedPublicClass
* UndocumentedPublicFunction
* UndocumentedPublicProperty

+ fix up a few spots it found

(excludes the rule on the sample app)

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/StickyContentPadding.kt
+++ b/appcues/src/main/java/com/appcues/trait/StickyContentPadding.kt
@@ -6,6 +6,9 @@ import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.ui.unit.Density
 
+/**
+ * Type that defines the padding values necessary to account for sticky content elements when composing the step.
+ */
 public class StickyContentPadding(private val density: Density) {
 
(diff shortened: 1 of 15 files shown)


appcues/src/main/java/com/appcues/trait/StickyContentPadding.kt
```diff
@@ -6,37 +6,69 @@ import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.ui.unit.Density
 
+/**
+ * Type that defines the padding values necessary to account for sticky
+ * content elements in a composition.
```
andretortolano, line 11: not sure if we can be more specific with something like:

Type that defines the padding values necessary to account for sticky content elements when composing the step

andretortolano (approved):

appcues/src/main/java/com/appcues/trait/StickyContentPadding.kt
iujames: yeah that sounds good to me, updated, thanks

PR #403, opened 2023-05-08
Add support for XML resource ID based selectors

iujames:
A pretty simple update in our selector structure, this adds `resourceName`, using the resource ID string representation from how the view was defined in XML layouts, as a property you can use for target element selection.

This approach would only add that property if the `View.isClickable` was `true` - for commonly interactive elements like buttons, tabs, menus. It would avoid a lot of the extra noise in the screenshot for container elements or internal system views, but still make a lot of common tooltip targets selectable "for free".  I'm wondering if there is a better filter to use here than `isClickable` still?


andretortolano (approved):
attributing different weight for different types of selectors 👍 
getting the resource entryName condition makes sense to me 👍

PR #405, opened 2023-05-10
🐛 removing customer view from talkback stack when experiences are showing

andretortolano:
Sets underlaying views to be ignored by talkback while experiences are showing.

by changing the importantForAccessibility flag on specific view, we ensure talkback will ignore it while we have our overlayed view on screen


iujames (approved):
Seems reasonable to me, we just have to keep this list of excluded view IDs in mind should we ever add anything else, but I can't think of a cause for concern there currently.

PR #407, opened 2023-05-22
✨ Added displayName property when capturing view elements

andretortolano:
Initial idea adding property displayName as part of the information builder will use to target an element.

As is now we are basically extending the existing information and adding that to a new property called displayName, we can later explore other ideas around gathering more information on specificy View types to improve displayName  and type properties


appcues/src/main/java/com/appcues/ElementTargetingStrategy.kt
```diff
@@ -97,6 +97,11 @@ public data class ViewElement(
      */
     val type: String,
 
+    /**
+     * A value representing the display name of the view.
+     */
+    val displayName: String?,
```
iujames, line 103: This is exactly where I would expect it (not in the selector itself, but part of the ViewElement)

Currently, the expectation is that this is only required to be set if the view has a selector - so many views may not have a display name and that is OK. Pondering if we should make `displayName` a required prop on the `ElementSelector` interface. Considering cross platform, like React Native, where it has it's own implementation of ElementSelector and it's own targeting strategy completely - it will need updated as well to provide a displayName.

I think it's probably best to _not_ require it on `ElementSelector` and leave it up to each implementation to handle as desired. Which is in line with what is here, just thinking it through.

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
```diff
@@ -21,9 +21,18 @@ internal data class AndroidViewSelector(
...
     val isValid: Boolean
         get() = contentDescription != null || tag != null || resourceName != null
 
+    val displayName: String?
+        get() = when {
+            resourceName != null -> "RES/$resourceName"
+            tag != null -> "TAG/$tag"
+            contentDescription != null -> contentDescription
+            else -> null
+        }
```
iujames, line 34: seems reasonable - maybe next step is to generate a snapshot from the test app and run it by Sam? Just to see if he has any opinions on the user facing aspect of this. Does he want to include `RES/` prefix for instance? I'm not sure if that has meaning to the end user in the builder or not.

Beyond this scope - do we want to try to identify if the type has `Button` or  `Label` or `Image` etc in the name and attempt to be making any edits to the name like `Button: $resource_name` ? I kind of hope not as I'm not sure how maintainable that is, but interested in his expectations.

appcues/src/main/java/com/appcues/ElementTargetingStrategy.kt
andretortolano: I think it makes sense to leave displayName here because it would be part of the interface contract the builder would expect to display some "proper name", this is in line with open builder concept right, this is why I asked about the type, right now type is not used for anything but we could enhance it and use that as a parameter to know which icons to show in builder, etc.

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
andretortolano: so as a first take I thought how we could differ the display name, my thought process here was to follow the same priority order we do for weight inside evaluateMatch. but the first problem is

1. we are basically in need of similar logic for both places to keep in sync.
2. if thats the case, changing weight or adding new selector types could change how the view is displayed.

things to ponder. I could not solve this one, I know its a good starting point but needs to evolve

iujames: I noted a discussion topic for tomorrow just to get any input Sam has on desired UX in builder. Could you generate a snapshot with the examples as we have it here for him?

andretortolano: as discussed with the team, looks like we can move on with this for now and revisit if necessary in the future

iujames (approved):

PR #408, opened 2023-05-22
⬆️  Update gradle tools to 8.0

andretortolano:
Updating gradle tools to 8.0, for some reason now we need to run cmd commands on JDK 17, so if you are not on 17 yet, make sure to install it and update the path.

`brew install openjdk@17`

`echo 'export PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH"' >> ~/.zshrc`

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -4,11 +4,6 @@
     <option name="RIGHT_MARGIN" value="140" />
     <option name="SOFT_MARGINS" value="140" />
     <JetCodeStyleSettings>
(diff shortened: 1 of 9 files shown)


iujames (commented):
Looks like we'll either need to hold on this one, or find a way to update the build machine in CircleCI correct? As discussed in chat, I think for our snapshot tests at least we need to stay on JDK 11 for Shot usage, for now.

For the `.idea/**` files we can exclude from `.gitignore` I think.

andretortolano:
I'm closing this one. lets wait and see

PR #409, opened 2023-05-30
🐛 Fix tooltip max height when it overflows available space

andretortolano:
Adding a maxHeight value to Top and Bottom tooltip position, that can contain a value when we fall under the two last scenarios, where there is no good place to position the tooltip, and then we calculate the preferred position considering all the limitations. Now along with that, reporting back the available maxHeight makes so that at least we have matching behavior for those edge cases between android and IOS.

image
image
image

There were two bugs related to this issue, [#51484](https://app.shortcut.com/appcues/story/51484/ios-and-android-behavior-difference-with-large-pointer-length-on-tooltip) and [#53188](https://app.shortcut.com/appcues/story/53188/long-tooltip-content-overlaps-target-area)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@ fastlane/Preview.html
 fastlane/screenshots/**/*.png
 fastlane/test_output
 vendor/
+/.idea/kotlinc.xml
--- a/appcues/src/main/java/com/appcues/trait/appcues/TooltipPathExt.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/TooltipPathExt.kt
@@ -42,6 +42,7 @@ internal data class ContainerCornerRadius(
 internal sealed class TooltipPointerPosition {
 
     abstract val isVertical: Boolean
+    abstract val maxHeight: Dp?
 
     private val _alignment = mutableStateOf(PointerAlignment())
     val alignment: State<PointerAlignment> = _alignment
@@ -66,7 +67,7 @@ internal sealed class TooltipPointerPosition {
      */
     abstract fun toContainerCornerRadius(cornerRadius: Dp): ContainerCornerRadius
 
-    object Top : TooltipPointerPosition() {
+    class Top(override val maxHeight: Dp? = null) : TooltipPointerPosition() {
 
         override val isVertical = true
 
@@ -82,7 +83,7 @@ internal sealed class TooltipPointerPosition {
         }
     }
 
-    object Bottom : TooltipPointerPosition() {
+    class Bottom(override val maxHeight: Dp? = null) : TooltipPointerPosition() {
 
         override val isVertical = true
 
@@ -101,6 +102,7 @@ internal sealed class TooltipPointerPosition {
     object Left : TooltipPointerPosition() {
 
         override val isVertical = false
+        override val maxHeight: Dp? = null
 
         override fun toContainerCornerRadius(cornerRadius: Dp): ContainerCornerRadius {
             val ignoreRounding = alignment.value.ignoreRounding
@@ -117,6 +119,7 @@ internal sealed class TooltipPointerPosition {
     object Right : TooltipPointerPosition() {
 
         override val isVertical = false
+        override val maxHeight: Dp? = null
 
         override fun toContainerCornerRadius(cornerRadius: Dp): ContainerCornerRadius {
             val ignoreRounding = alignment.value.ignoreRounding
@@ -133,6 +136,7 @@ internal sealed class TooltipPointerPosition {
     object None : TooltipPointerPosition() {
 
         override val isVertical = false
+        override val maxHeight: Dp? = null
 
         override fun toContainerCornerRadius(cornerRadius: Dp): ContainerCornerRadius {
             return ContainerCornerRadius(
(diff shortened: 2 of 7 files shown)


appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
```diff
@@ -387,12 +387,13 @@ internal class TooltipTrait(
...
         tooltipSettings: TooltipSettings,
         useBottomToastStyle: Boolean,
+        maxHeight: Dp?,
     ): Modifier =
         composed {
             // keep containerMaxSize here as a rememberable so it force recomposition when value change
             val containerMaxSize = DpSize(
                 width = windowInfo.widthDp - (SCREEN_HORIZONTAL_PADDING * 2),
-                height = windowInfo.heightDp - (SCREEN_VERTICAL_PADDING * 2)
+                height = maxHeight ?: (windowInfo.heightDp - (SCREEN_VERTICAL_PADDING * 2))
```
andretortolano, line 396: maxHeight coming from tooltipPosition, if null we use previous value (screen size - paddings)

.gitignore
```diff
@@ -51,4 +51,5 @@ fastlane/report.xml
 fastlane/Preview.html
 fastlane/screenshots/**/*.png
 fastlane/test_output
-vendor/
\ No newline at end of file
+vendor/
+/.idea/kotlinc.xml
```
iujames, line 55: nit: maybe bundle up with the others starting on line 21 to organize, or just exclude `/.idea/*` at this point?

edit: i see right below at least one .idea file for detekt that would make the * a bad idea probably.

appcues/src/main/java/com/appcues/trait/extensions/TargetRectangleInfoExt.kt
```diff
@@ -75,14 +76,21 @@ internal fun TargetRectangleInfo?.getTooltipPointerPosition(
...
         // passed the preference positions we position the tooltip wherever is available.
-        canPositionVertically -> if (excessSpaceTop > excessSpaceBottom) Bottom else Top
+        canPositionVertically -> if (excessSpaceTop > excessSpaceBottom) Bottom() else Top()
         canPositionHorizontally -> if (excessSpaceLeft > excessSpaceRight) Right else Left
         // Doesn't fit anywhere so pick the top/bottom side that has the most space.
         // Allowing left/right here would mean the width gets compressed and that opens a can of worms.
-        excessSpaceTop > excessSpaceBottom -> Bottom
-        else -> Top
+        excessSpaceTop > excessSpaceBottom -> Bottom(availableHeightTop)
+        else -> Top(availableHeightBottom)
```
andretortolano, line 93: calculate availableHeightTop and Bottom, and pass it forward when we hit the edge cases when it wont fit anywhere

appcues/src/main/java/com/appcues/trait/appcues/TooltipPathExt.kt
```diff
@@ -66,7 +67,7 @@ internal sealed class TooltipPointerPosition {
      */
     abstract fun toContainerCornerRadius(cornerRadius: Dp): ContainerCornerRadius
 
-    object Top : TooltipPointerPosition() {
+    class Top(override val maxHeight: Dp? = null) : TooltipPointerPosition() {
```
iujames, line 70: should it be a `data class` ?

iujames (approved):
I'm a bit surprised at the amount of change necessary I guess. I thought it would basically act as a min height 48 on the content and allow scaling down as needed to get closer to the target. Having to change Top and Bottom to classes and pass data around in multiple spots, and change other `when` usage feels a bit unfortunate. Is that to avoid having to do height calculations in two places? (1) finding pointer position (2) actually setting the tooltip size in the trait. Just thinking over if it can be any simpler, but the end result looks good!

appcues/src/main/java/com/appcues/trait/appcues/TooltipPathExt.kt
andretortolano: data class generates a lot of boilerplate we wont use, wont change much here, we could edit it if we want.

on the topic of changing from object to class, which led to a lot of small changes like  including "is" in front of classes on when statements, I feel like its a value that is tied to the tooltipPosition calculation, so it would make sense to have it there, I did it this way to avoid passing in an extra argument for maxHeight like we already had before. 

this way we avoid having to gather all this values in two different places, and its also an information that we may/may not need, depending on which scenario we fall into when considering the pref position and the excess and available spaces on screen

iujames: Yeah, I think I can see how it would require handling in two different places otherwise, so that is fair. I also tested the scenario where the content is tall but there is still enough space to position the tooltip left or right, rather than shrink content, and that seems to work as expected 👍. It keeps the tall content and moves the pointer to the right side in this example, like I'd expect.

Screenshot 2023-05-31 at 9 15 56 AM

andretortolano: yea we are not changing the exiting logic, the maxHeight is calculated for when we fall into the edge cases only (two last conditions on the when statement [here](https://github.com/appcues/appcues-android-sdk/pull/409/files#diff-ade2386127bf7b29bbf36b66130e680bc4e704eddfa49fa010e1797a1983f49eR93))

PR #413, opened 2023-06-05
♻️  update capture screen displayName logic

andretortolano:
changing the logic around displayName, and also including another change to support appcuesId as Compose appcues tag property coming from the JSON selector

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
+++ b/appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
@@ -7,38 +7,52 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.compose.ui.semantics.SemanticsNode
 import androidx.compose.ui.semantics.SemanticsOwner
+import androidx.compose.ui.semantics.SemanticsProperties
 import androidx.compose.ui.semantics.SemanticsPropertyKey
 import androidx.compose.ui.semantics.SemanticsPropertyReceiver
 import androidx.core.view.children
 import com.appcues.ElementSelector
 import com.appcues.ElementTargetingStrategy
 import com.appcues.ViewElement
+import com.appcues.debugger.screencapture.AndroidViewSelector.Companion.SELECTOR_APPCUES_ID
+import com.appcues.debugger.screencapture.AndroidViewSelector.Companion.SELECTOR_CONTENT_DESCRIPTION
+import com.appcues.debugger.screencapture.AndroidViewSelector.Companion.SELECTOR_RESOURCE_NAME
+import com.appcues.debugger.screencapture.AndroidViewSelector.Companion.SELECTOR_TAG
 import com.appcues.isAppcuesView
 import com.appcues.monitor.AppcuesActivityMonitor
 
-internal data class AndroidViewSelector(
-    var contentDescription: String? = null,
-    var tag: String? = null,
-    var resourceName: String? = null
+internal class AndroidViewSelector(
+    private val properties: Map<String, String?>,
+    val type: String? = null,
 ) : ElementSelector {
+
+    companion object {
+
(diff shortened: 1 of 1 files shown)


appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
```diff
@@ -7,38 +7,52 @@ import android.view.View
...
+        const val SELECTOR_APPCUES_ID = "appcuesId"
+    }
+
+    private val contentDescription: String? = properties[SELECTOR_CONTENT_DESCRIPTION]
+    private val tag: String? = properties[SELECTOR_TAG]
+    private val resourceName: String? = properties[SELECTOR_RESOURCE_NAME]
+    private val appcuesId: String? = properties[SELECTOR_APPCUES_ID]
+
     val isValid: Boolean
         get() = contentDescription != null || tag != null || resourceName != null
```
iujames, line 43: i think need to add the `appcuesId` case here

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
```diff
@@ -54,6 +68,10 @@ internal data class AndroidViewSelector(
                 weight += 100
             }
 
+            if (it.appcuesId != null && it.appcuesId == appcuesId) {
+                weight += 100
+            }
```
iujames, line 73: thinking we should make this match the [iOS side](https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift#L50-L52) where appcuesID matches the highest level selector (so the 1000 case here matching resource name). In practice, right now it would always be mutually exclusive, granted.

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
```diff
@@ -7,38 +7,52 @@ import android.view.View
...
 
     val displayName: String?
         get() = when {
-            resourceName != null -> "RES/$resourceName"
-            tag != null -> "TAG/$tag"
+            resourceName != null -> "$type ($resourceName)"
+            tag != null -> "$type (tag $tag)"
+            appcuesId != null -> appcuesId
             contentDescription != null -> contentDescription
             else -> null
```
iujames, line 51: I think of `resourceName` analogous to `accessibilityIdentifier` on the [iOS side](https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift#L85-L96)

So matching that pattern, how about
```kotlin
when {
    resourceName != null -> resourceName
    appcuesId != null -> appcuesId
    tag != null -> "$type (tag $tag)"
    contentDescription -> "$type ($contentDescription)"
    else -> null

}
```

andretortolano: I will change that, I inverted resourceName with contentDescription there

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
```diff
@@ -193,7 +209,7 @@ private fun SemanticsNode.asCaptureView(context: Context): ViewElement {
             height = unclippedGlobalBounds.height().toDp(density),
             displayName = it?.displayName,
             selector = it,
-            type = "Composable #$id",
+            type = it?.type ?: "Composable #$id",
```
andretortolano, line 212: just a note, since type can be null here, i kept the "Composable #$id" which only appears on views that fail to provide a selector thus ignored by builder later on

PR #419, opened 2023-06-15
Add doc on instrumenting views for element targeting

iujames:
For [sc-53872]

Covering the general element targeting concept here, and how to instrument android views, but keeping it worded more specifically to apply to "Anchored Tooltips" to try to match the customer expectations around feature naming and what they'd be looking for information about. Open to any ideas on that though.

This general format will be applied to iOS and React Native once we're happy with it.

Thinking maybe this one should be linked to from the [Getting Started](https://github.com/appcues/appcues-android-sdk/blob/main/docs/GettingStarted.md) page.


andretortolano (approved):
maybe a new section for "Tooltips" could go over a brief explanation on how that works and linking to this doc

PR #420, opened 2023-06-15
Improve handling of margin for system bars

iujames:
The previous way we were checking for the status bar height seems to not work on certain devices (Pixel 7 at least in testing)

This approach using WindowInsets for systemBars seems more reliable.


appcues/src/main/res/layout/appcues_activity_layout.xml
iujames, line 1: this was just an old file that is no longer referenced that I came across.

andretortolano (approved):
Lets push that and wait to test it out more

appcues/src/main/res/layout/appcues_activity_layout.xml
andretortolano: yea nice find

PR #421, opened 2023-06-21
Add support for multi window scenarios in element targeting + overlay

iujames:
This handles situations where simply taking the
```kotlin
activity.window.decorView.rootView
```
to be the root view for injecting overlay content  is not sufficient.

If a dialog view / modal is being shown in the customer application, this will cause another window and decorView to show up in the view hierarchy. We need to actually get all of the top level windows in the Activity and find the topmost item (last) and use this for screen capture, element targeting, and content overlay.

This [post](https://stackoverflow.com/a/19776489) had some helpful tips about how to get this info, which is not always straightforward. We use a multi-layer approach, where API 29+ can use a [handy WindowInspector function](https://developer.android.com/reference/android/view/inspector/WindowInspector#getGlobalWindowViews()) to get this info easily. Below that level, we have to make a _safeguarded_ attempt at reflection to get this info. If all else fails, we just use the main window for the Activity like before. This is not ideal, but the vast majority of the time it is correct.

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
+++ b/appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
@@ -1,8 +1,11 @@
 package com.appcues.ui
 
 import android.app.Activity
+import android.os.Build.VERSION
+import android.os.Build.VERSION_CODES
 import android.view.View
 import android.view.ViewGroup
+import android.view.inspector.WindowInspector
 import androidx.compose.runtime.remember
 import androidx.compose.ui.platform.ComposeView
 import androidx.core.view.ViewCompat
@@ -10,7 +13,12 @@ import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.children
 import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.ViewTreeLifecycleOwner
+import androidx.lifecycle.ViewTreeViewModelStoreOwner
 import androidx.lifecycle.findViewTreeLifecycleOwner
+import androidx.lifecycle.findViewTreeViewModelStoreOwner
+import androidx.savedstate.findViewTreeSavedStateRegistryOwner
+import androidx.savedstate.setViewTreeSavedStateRegistryOwner
 import com.appcues.R
 import com.appcues.databinding.AppcuesOverlayLayoutBinding
 import com.appcues.logging.Logcues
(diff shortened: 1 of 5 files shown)


appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
```diff
@@ -123,15 +132,59 @@ internal class AppcuesOverlayViewManager(private val scope: Scope) : DefaultLife
...
+
+    if (ViewTreeViewModelStoreOwner.get(decorView) == null) {
+        val viewModelStoreOwner = window.decorView.findViewTreeViewModelStoreOwner()
+        ViewTreeViewModelStoreOwner.set(decorView, viewModelStoreOwner)
+    }
+
+    if (decorView.findViewTreeSavedStateRegistryOwner() == null) {
+        val savedStateRegistryOwner = window.decorView.findViewTreeSavedStateRegistryOwner()
+        decorView.setViewTreeSavedStateRegistryOwner(savedStateRegistryOwner)
+    }
```
iujames, line 187: workarounds for issues like those noted in https://issuetracker.google.com/issues/180691023 - we can't rely on the decorView to be well formed in some cases other than the normal main window decorView.

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
```diff
@@ -123,15 +132,59 @@ internal class AppcuesOverlayViewManager(private val scope: Scope) : DefaultLife
...
+        try {
+            // this is the less desirable method for API 21-28, using reflection to try to get the root views
+            val windowManagerClass = Class.forName("android.view.WindowManagerGlobal")
+            val windowManager = windowManagerClass.getMethod("getInstance").invoke(null)
+            val getViewRootNames: Method = windowManagerClass.getMethod("getViewRootNames")
+            val getRootView: Method = windowManagerClass.getMethod("getRootView", String::class.java)
+            val rootViewNames = getViewRootNames.invoke(windowManager) as Array<Any?>
+            val rootViews = rootViewNames.map { getRootView(windowManager, it) as View }
+            rootViews.last()
+        } catch (e: Exception) {
```
andretortolano, line 164: not sure if makes difference to detekt but can replace e for _

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
andretortolano: good catch

andretortolano (approved):

PR #422, opened 2023-06-21
♻️ [Embed - Part 1] - isolating state machine to ensure its only accessible through ExperienceRenderer

andretortolano:
The goal here was to isolate the StateMachine, and ensure that anything that used to use the StateMachine object now should rely on ExperienceRenderer instance instead. That included changes to Actions, Tests, but the main change is on ExperienceRenderer. 

Not ideal but as part of this I took some side-quests, which includes:

- Simplifying AnalyticsPolicy since we are not using AppcuesActivity
- Reorganizing some of the classes and its responsibilities in the hopes of solving the circular dependency we have today with AnalyticsTracker (More of it in the following PR) 

**There are some concerns pointed out here in comments that I hope we can evolve on the following PRs.**

So far we these other two PRs that includes at lot of changes towards the end goal of embeds.

#424 - Support to multiple state machines.
#425 - Improving communication between ViewModel and StateMachine (stateFlow)

TODO:

- Properly get the frameID during RenderContext mapping
- create the API call to register a view tied to a frameID
- Create @embed trait and start working on the logic of finding the frameID where we should display the content
- Handle the special case where the frameID is not there, but could be registered later.
- Add specific analytics calls to experience failed/ recovered to match iOS (that is supposed to be matching Web)

🚀


appcues/src/main/java/com/appcues/analytics/AnalyticsPolicy.kt
```diff
@@ -1,92 +1,16 @@
...
-            if (skip) {
-                logcues.info("skipping duplicate screen tracking after experience end")
-                return false
-            }
-        }
-
-        lastScreen = title
-        return true
-    }
+    fun canTrackScreen() = basicAnalyticsPolicy("unable to track screen")
```
andretortolano, line 13: cleared this method as this extra check is no longer needed due to our change from Activity to custom view

appcues/src/main/java/com/appcues/statemachine/Transitions.kt
```diff
@@ -230,11 +230,11 @@ private fun errorTransition(experience: Experience, currentStepIndex: Int, messa
     )
 }
 
-private fun stepNotFoundErrorTransition(experience: Experience, groupIndex: Int, stepIndex: Int): Transition {
+private fun stepNotFoundErrorTransition(experience: Experience, groupIndex: Int): Transition {
```
andretortolano, line 233: small adjustment to this error since its only being called passing step group and step 0,0

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -92,20 +81,20 @@ internal class ExperienceLifecycleTracker(
         analyticsTracker.track(event.name, properties, interactive = false, isInternal = true)
     }
 
-    private fun State.shouldTrack(): Boolean =
+    private fun State.shouldNotTrackUnpublished(): Boolean =
```
iujames, line 84: this feels a little less readable to me, personally. Could keep `shouldTrack()` and check the `.not()` case above on line 35?

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -16,73 +17,61 @@ import com.appcues.statemachine.State.EndingExperience
...
-                            // track the experience_completed event
-                            trackLifecycleEvent(ExperienceCompleted(it.experience))
-                        } else {
-                            // otherwise its considered experience_dismissed (not completed)
-                            trackLifecycleEvent(ExperienceDismissed(it.experience, it.flatStepIndex))
-                        }
-                    }
-                    else -> Unit
+        when (this) {
+            is BeginningExperience -> config.experienceListener?.experienceStarted(experience.id)
```
iujames, line 38: ok so this is replacing what used to be in the state machine logic, trying to follow along.

appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
```diff
@@ -61,7 +58,7 @@ internal class AppcuesViewModel(
 
     init {
         viewModelScope.launch {
-            stateMachine.stateFlow.collectLatest { result ->
+            experienceRenderer.stateFlow.collectLatest { result ->
```
iujames, line 61: im not sure if i follow the change yet to hoist the stateFlow up to the experienceRenderer, rather than the state machine. This view model instance is tied to one particular experience right? so shouldn't it be listening to the state changes from a single state machine operating on that experience? What if multiple things are running at the same time, wouldn't the experienceRendering now be sending updates through this flow for all of them? even things unrelated to the experience being rendered by this view model?

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -1,40 +1,79 @@
...
+        experienceLifecycleTracker.onState(state)
+
+        _stateFlow.emit(state)
+    }
+
+    override suspend fun onError(error: Error) {
+        experienceLifecycleTracker.onError(error)
+
+        _errorFlow.emit(error)
+    }
```
iujames, line 71: as noted prior, these are the things Im wondering if should remain owned by the stateMachine (can have N active instances) versus the experienceRenderer (single instance)

iujames (commented):
In general, the concept of insulating the stateMachines inside of ExperienceRenderer, and then having outside traits and actions interact with the ExperienceRenderer rather than a state machine directly makes sense to me, and very much aligns with the iOS structure being worked out 👍

I'm not certain about some of the details about state flow observation and analytics that got moved though, may need to chat that over.

appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
andretortolano: you are right!

andretortolano: we need to change a couple of things here, I will see if I can make it work

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
andretortolano: I was trying to move stuff out of the statemachine and leave if with as less dependencies as possible, since this is related to the experience lifecycle i was hoping we could plug it in here

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
iujames: Nitpick, but following changes here, I think this was updated to `shouldTrackUnpublished`? I was thinking just keep `shouldTrack()` as the implementation details of what constitutes a valid item to track should be contained within (in this case, checks for published content).

PR #424, opened 2023-06-29
Feature/embed/supporting multiple state machines

andretortolano:


appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
```diff
@@ -84,7 +84,7 @@ internal sealed class ExperienceLifecycleEvent(
         get() = hashMapOf<String, Any?>(
             "experienceId" to experience.id.appcuesFormatted(),
             "experienceName" to experience.name,
-            "experienceType" to experience.type,
+            "experienceType" to experience.renderContext.toString(),
```
andretortolano, line 87: this is an important callout since we have a override fun toString() that returns the experienceType.

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -72,6 +72,7 @@ internal class ExperienceLifecycleTracker(
             is Error.ExperienceError -> trackLifecycleEvent(ExperienceError(this))
             is Error.StepError -> trackLifecycleEvent(StepError(this))
             is Error.ExperienceAlreadyActive -> Unit
+            is Error.NoActiveStateMachine -> Unit
```
andretortolano, line 75: On the next PR I ended up removing this error type.

appcues/src/main/java/com/appcues/data/model/RenderContext.kt
```diff
@@ -0,0 +1,13 @@
+package com.appcues.data.model
+
+internal sealed class RenderContext {
+    object Modal : RenderContext()
+    data class Embed(val frameId: String) : RenderContext()
+
+    override fun toString(): String {
```
andretortolano, line 7: Not sure if these are the values we want here, we can change later and make sure this matches iOS

appcues/src/main/java/com/appcues/statemachine/Error.kt
```diff
@@ -1,12 +1,17 @@
...
 
 internal sealed class Error(open val message: String) {
+
     val id: UUID = UUID.randomUUID()
 
     data class ExperienceAlreadyActive(val ignoredExperience: Experience, override val message: String) : Error(message)
     data class ExperienceError(val experience: Experience, override val message: String) : Error(message)
     data class StepError(val experience: Experience, val stepIndex: Int, override val message: String) : Error(message)
+
+    data class NoActiveStateMachine(val renderContext: RenderContext) :
```
andretortolano, line 15: as I noted before, this will be removed on the following PR.

appcues/src/main/java/com/appcues/statemachine/StateMachineFactory.kt
```diff
@@ -0,0 +1,16 @@
+package com.appcues.statemachine
+
+import com.appcues.AppcuesCoroutineScope
+import com.appcues.action.ActionProcessor
+import com.appcues.data.model.RenderContext
+import com.appcues.statemachine.StateMachine.StateMachineListener
+
+internal class StateMachineFactory(
```
andretortolano, line 8: a Factory of StateMachines, the create method receives the renderContext and a StateMachineListener (this will be refactored later, since we will be moving away from using this StateMachineListener and revert back to StateFlow

appcues/src/main/java/com/appcues/action/AppcuesActionFactories.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.appcues.action.appcues.LaunchExperienceAction
+import com.appcues.action.appcues.LinkAction
+import com.appcues.action.appcues.RequestReviewAction
+import com.appcues.action.appcues.SubmitFormAction
+import com.appcues.action.appcues.TrackEventAction
+import com.appcues.action.appcues.UpdateProfileAction
+import org.koin.core.parameter.parametersOf
+import org.koin.core.scope.Scope
+
+internal fun Scope.closeActionFactory(): ActionFactoryBlock {
```
andretortolano, line 14: created this file for both actions and traits. trying to ensure readability on their main class (ActionRegistry and TraitRegistry)

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -1,50 +1,42 @@
...
 
-    private val repository by inject<AppcuesRepository>()
-    private val stateMachine by inject<StateMachine>()
-    private val sessionMonitor by inject<SessionMonitor>()
-    private val config by inject<AppcuesConfig>()
-
-    // lazy prop inject here to avoid circular dependency with AnalyticsTracker
-    // AnalyticsTracker > AnalyticsQueueProcessor > ExperienceRenderer(this) > AnalyticsTracker
-    private val analyticsTracker by inject<AnalyticsTracker>()
+    private val stateMachines: HashMap<RenderContext, StateMachine> = hashMapOf()
```
iujames, line 39: will need something like this. Im not yet following how we'll track what `AppcuesFrame` (view) owns each state machine for the embed views, and do so in a way that allows for memory to be managed when those views go out of scope. See the iOS side where there is the one `stateMachine` in the `ExperienceRenderer`, then the `StateMachineDirectory` that manages this mapping and lookup with weak references to the `StateMachineOwning` (views)

iujames (commented):
There is a lot of good progress here. I made a few specific comments on spots and tried to understand the bigger picture as best I could in the first pass.

Honestly, it's just too much to comprehend and too many different changes going on. I really suggest we step back and take the time to structure in smaller increments and be able to update tests as things progress to ensure to regressions.

From what I'm understanding, perhaps one approach to break it down into work chunks:

1. Introduce RenderContext and set it on Experience models appropriately during decoding
2. Update ExperienceRenderer functions to depend on RenderContext, but do nothing with it yet. Call sites pass in fixed `Modal` value to start
3. Update Traits and Actions to init with RenderContext, pass it through all the call sites as needed, and update ExperienceRenderer call sites to use real value - still no actual usage in ExperienceRenderer
4. Updates to support creating and tracking multiple state machines in the ExperienceRender, using the RenderContext - the dictionary management
5. Possibly unrelated other refactoring of analytics tracker / experience render? (tbd)

appcues/src/main/java/com/appcues/data/model/RenderContext.kt
```diff
@@ -0,0 +1,13 @@
+package com.appcues.data.model
```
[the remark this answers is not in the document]
andretortolano: I think if we are adding a new property to the Experience model, then it should sit in data.model

PR #425, opened 2023-06-30
♻️ [Embed - Part 3] - Fixing ViewModel - StateMachine communication.

andretortolano:
@iujames pointed out that the current idea to use a callback was not going to work properly since any viewModel would be collecting events from all StateMachines.

This changes includes a new method in ExperienceRenderer called collectState that should be sufficient to make it all work (pending rigorous testing when we get embeds to fully work)

but in order to make that work some changes where required, including passing renderContext in some other places, and also trying to simplify in other places. for instance every State and Error should belong to an experience, so I enforced those sealed classes to include the experience, this makes it easier to connect the experience references where its needed.


appcues/src/main/java/com/appcues/statemachine/StateMachineFactory.kt
```diff
@@ -2,15 +2,14 @@ package com.appcues.statemachine
...
+import com.appcues.data.model.Experience
 
 internal class StateMachineFactory(
     private val appcuesCoroutineScope: AppcuesCoroutineScope,
     private val actionProcessor: ActionProcessor,
 ) {
 
-    fun create(renderContext: RenderContext, listener: StateMachineListener): StateMachine {
-        return StateMachine(renderContext, appcuesCoroutineScope, actionProcessor).also { it.listener = listener }
+    fun create(experience: Experience): StateMachine {
```
andretortolano, line 12: the dynamic information every state machine needs is the experience now

appcues/src/main/java/com/appcues/trait/AppcuesTraitFactories.kt
```diff
@@ -55,6 +47,14 @@ internal fun Scope.backgroundContentTraitFactory(): TraitFactoryBlock {
     return { config, level, _ -> get<BackgroundContentTrait> { parametersOf(config, level) } }
 }
 
+internal fun Scope.modalTraitFactory(): TraitFactoryBlock {
```
andretortolano, line 50: all presenting traits needs to know whats the renderContext ( this was an oversight on my last pass, where i focused so  much more on Actions and did not think too much about traits)

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
```diff
@@ -70,7 +74,7 @@ internal class AppcuesOverlayViewManager(private val scope: Scope) : DefaultLife
         if (parentView.findViewById<ComposeView>(R.id.appcues_overlay_layout) == null) {
             // then we add
             val binding = AppcuesOverlayLayoutBinding.inflate(layoutInflater)
-            if (viewModel == null) viewModel = AppcuesViewModel(scope, ::onCompositionDismiss)
+            if (viewModel == null) viewModel = AppcuesViewModel(scope, renderContext, ::onCompositionDismiss)
```
andretortolano, line 77: PresentingTraits needs to be aware of the RenderContext so they can create an instance of AppcuesViewModel

appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
```diff
@@ -56,9 +57,9 @@ internal class AppcuesViewModel(
 
     init {
         viewModelScope.launch {
-            experienceRenderer.stateFlow.collectLatest { result ->
+            experienceRenderer.collectState(renderContext) { result ->
```
andretortolano, line 60: knowing the renderContext is important because of this utility method that collects only events for that specific renderContext

appcues/src/main/java/com/appcues/ui/AppcuesViewModelFactory.kt
```diff
@@ -1,16 +0,0 @@
-package com.appcues.ui
```
andretortolano, line 1: file was not being used since we are not using Activity anymore.

On second thought, I would be interesting to investigate if we really dont need it or if we should instead find a way to use this factory while creating a custom view, as far as I know this ensures that the viewModel can be recreated with the same dependencies.  (did not look into it)

appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
iujames: yes, this seems the key change here - being able to listen to the state machine updates for the specific state machine that this VM is rendering, through the shared ExperienceRenderer reference.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -11,52 +12,40 @@ import com.appcues.data.model.ExperiencePriority.NORMAL
...
-    private val _stateFlow = MutableSharedFlow<State>(1)
-    val stateFlow: SharedFlow<State>
-        get() = _stateFlow
+    private val _allStates = MutableSharedFlow<State>(1)
 
-    private val _errorFlow = MutableSharedFlow<Error>(1)
-    val errorFlow: SharedFlow<Error>
-        get() = _errorFlow
+    suspend fun collectState(renderContext: RenderContext, action: suspend (value: State) -> Unit) =
+        _allStates.filter { it.experience.renderContext == renderContext }.collectLatest(action)
```
iujames, line 46: is it necessary to have this `_allStates` aggregate flow? Could the VM usage instead just listen to the `StateMachine` for it's `RenderContext` directly?

iujames (commented):
I'm not yet understanding the changes around having every State and Error capture the Experience where we did not have previously.

appcues/src/main/java/com/appcues/statemachine/State.kt
```diff
@@ -3,42 +3,32 @@ package com.appcues.statemachine
 import com.appcues.data.model.Experience
 import com.appcues.util.ResultOf
 
-internal sealed class State {
+internal sealed class State(open val experience: Experience) {
 
-    object Idling : State()
-    data class BeginningExperience(val experience: Experience) : State()
+    data class Idling(override val experience: Experience) : State(experience)
```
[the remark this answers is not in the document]
andretortolano: whenever we instanciate a new StateMachine, it will be tied to a particular experience. when the Experience is no longer visible, moves to idle or throws an error, we still know which experience it is.

this helps when we need to get the experience.renderContext when filtering which states should be sent to a particular subscriber from ExperienceRenderer

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
andretortolano: probably, I was imagining we would want allStates for debugger later, but maybe not

appcues/src/main/java/com/appcues/statemachine/State.kt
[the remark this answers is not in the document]
mmaatttt: @andretortolano the lightbulb moment for me was realizing that we need a state machine for each current render context (always a modal one running, and perhaps more for embed views), not each experience.

I also started by doing it with state machine associated with a particular experience, but realized it was a much nicer (conceptually and in the code). My initial thoughts are [here](https://appcues.slack.com/archives/C031RLGS4F8/p1686832146095029).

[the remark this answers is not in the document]
andretortolano: I'm not sure I follow, I understand there is an issue related to frameId not being present when an embedded PresenterTrait tries to present(), so to avoid that I was imagining that ExperienceRenderer would manage the experiences that are being displayed or not somehow. then the embeds that dont match a registered view would be stored in a cache list without being started, then when a Appcues.registerFrame(frameId) gets called we would check if that frameId represents a frameId we have a cached experience for.

to james argument I initially thought that there would be no difference between a modal renderContext and an embed, aside from the fact that Modal would be unique where Embed(frameId) can have multiple (1 for each frameId string)

PR #428, opened 2023-07-11
♻️  ExperienceRenderer depending on RenderContext to access StateMachine

andretortolano:
Changes included.

* Centralizing access to StateMachine by calling ExperienceRenderer
* Adding property RenderContext to ExperienceRenderer calls tied to StateMachine

### DIFF
--- a/appcues/src/main/java/com/appcues/analytics/AnalyticsPolicy.kt
+++ b/appcues/src/main/java/com/appcues/analytics/AnalyticsPolicy.kt
@@ -5,16 +5,17 @@ import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.ProcessLifecycleOwner
 import com.appcues.AppcuesCoroutineScope
 import com.appcues.SessionMonitor
+import com.appcues.data.model.RenderContext
 import com.appcues.logging.Logcues
 import com.appcues.statemachine.State.BeginningExperience
 import com.appcues.statemachine.State.EndingExperience
-import com.appcues.statemachine.StateMachine
+import com.appcues.ui.ExperienceRenderer
 import kotlinx.coroutines.launch
 
 internal class AnalyticsPolicy(
     private val sessionMonitor: SessionMonitor,
     appcuesCoroutineScope: AppcuesCoroutineScope,
-    private val stateMachine: StateMachine,
+    private val experienceRenderer: ExperienceRenderer,
     private val logcues: Logcues,
 ) : DefaultLifecycleObserver {
 
@@ -28,7 +29,7 @@ internal class AnalyticsPolicy(
     init {
         appcuesCoroutineScope.launch {
             ProcessLifecycleOwner.get().lifecycle.addObserver(this@AnalyticsPolicy)
-            stateMachine.stateFlow.collect {
+            experienceRenderer.getStateFlow(RenderContext.Modal).collect {
                 when (it) {
                     is BeginningExperience -> {
                         experienceActiveCount++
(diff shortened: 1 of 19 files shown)


appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
```diff
@@ -225,7 +226,7 @@ internal class DebuggerViewModel(
 
     fun captureScreen(debuggerState: MutableDebuggerState) {
         appcuesCoroutineScope.launch {
-            experienceRenderer.dismissCurrentExperience(markComplete = false, destroyed = false)
+            experienceRenderer.stop()
```
andretortolano, line 229: not sure if this impacts anything, instead of calling dismissCurrentExperience I figured we could call stop instead before capturing the screen.

internally I noticed the difference is that stop() calls  action with "destroyed = yes". when destroyed is true it means there was an external exit from the experience and we should not rely on UI to move the StateMachine, I think its safe to change to destroyed = false since we are in control here.

appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
```diff
@@ -210,7 +210,7 @@ internal class StateMachine(
 
     fun stop() {
         appcuesCoroutineScope.launch {
-            handleAction(EndExperience(markComplete = false, destroyed = true))
+            handleAction(EndExperience(markComplete = false, destroyed = false))
```
andretortolano, line 213: this is what I mean, when stop is called I think its intentional and we could handle it as destroyed = false, so it can be used in different places, but let me know if this looks wrong we can revert this piece and leave as is.

appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
andretortolano: the other alternative I think would be to close only the Modal experience, not sure whats correct here.

iujames (commented):
looking good to me - good step in the right direction for future flexibility while keeping existing things mostly unchanged.

let me handle the patch release now before merging this, just in case, and then this can go after whatever final touch up or testing you would like

iujames (approved):

appcues/src/main/java/com/appcues/analytics/AnalyticsPolicy.kt
```diff
@@ -28,7 +28,7 @@ internal class AnalyticsPolicy(
     init {
         appcuesCoroutineScope.launch {
             ProcessLifecycleOwner.get().lifecycle.addObserver(this@AnalyticsPolicy)
-            stateMachine.stateFlow.collect {
+            experienceRenderer.stateFlow.collect {
```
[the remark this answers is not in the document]
andretortolano: yes. I can patch it next in another PR

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -35,6 +39,8 @@ internal class ExperienceRenderer(
     private val config by inject<AppcuesConfig>()
     private val analyticsTracker by inject<AnalyticsTracker>()
 
+    val stateFlow = stateMachine.stateFlow
```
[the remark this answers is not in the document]
andretortolano: yes, did the small possible change for now, missed this but yeah I can include it here

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -113,8 +123,9 @@ internal class ExperienceRenderer(
         stateMachine.stop()
     }
 
-    suspend fun dismissCurrentExperience(markComplete: Boolean, destroyed: Boolean): ResultOf<State, Error> =
-        stateMachine.handleAction(EndExperience(markComplete, destroyed))
+    suspend fun dismissCurrentExperience(renderContext: RenderContext, markComplete: Boolean, destroyed: Boolean): ResultOf<State, Error> {
```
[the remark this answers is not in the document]
andretortolano: I think on IOS matt named dismiss.

appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
andretortolano: I will leave as it was, we can revisit this later.

PR #429, opened 2023-07-11
Support screenshot capture from plugins

iujames:
This is the extra extensibility point I think we'd need to support a Flutter plugin screen capture. The Flutter plugin could then provide the bitmap and sizing information needed that we cannot access directly from the normal screen capture method.


andretortolano (approved):
nice change allowing for custom screen capturing strategy on xplat plugins

PR #430, opened 2023-07-11
♻️ Update actions and traits to properly pass RenderContext

andretortolano:
Changes to:
* Actions
* Traits
* ViewModel

to account for dynamic RenderContext being passed down to use experienceRenderer


iujames (approved):
Rolling along now! Looks great, nice work

PR #431, opened 2023-07-14
♻️  Support Multiple state machines tied to each RenderContext

andretortolano:
A couple of changes here in order to support multiple state machines being created for each render context.

* Updated ExperienceRenderer: Added the dictionary with state machine slots, whenever a new RenderContext comes, if we dont have a state machine for it yet then we create, this will happen for any type of RenderContext, supposedly making it a generic logic that should work to any type of RenderContext now and in the future.

* Updated ExperienceLifecycleTracker: now during start it receives the state machine so it can observe states and errors. the start happens once for each state machine when its fabricated during ExperienceRenderer show method.

* Various changes to accommodate multiple state machines

Different from my last approach, we are not deleting any renderContext and its state machine from the map, they should just move to Idling and sit there unused.

Worth noting that all this is running Unit tests and UI test 🆗

### DIFF
--- a/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
+++ b/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
@@ -1,5 +1,6 @@
 package com.appcues.analytics
 
+import com.appcues.AppcuesCoroutineScope
 import com.appcues.Storage
 import com.appcues.analytics.ExperienceLifecycleEvent.ExperienceCompleted
 import com.appcues.analytics.ExperienceLifecycleEvent.ExperienceDismissed
@@ -9,6 +10,8 @@ import com.appcues.analytics.ExperienceLifecycleEvent.StepCompleted
 import com.appcues.analytics.ExperienceLifecycleEvent.StepError
 import com.appcues.analytics.ExperienceLifecycleEvent.StepSeen
 import com.appcues.statemachine.Error
+import com.appcues.statemachine.Error.ExperienceAlreadyActive
+import com.appcues.statemachine.Error.RenderContextNotActive
 import com.appcues.statemachine.State
 import com.appcues.statemachine.State.BeginningExperience
 import com.appcues.statemachine.State.BeginningStep
@@ -20,29 +23,28 @@ import com.appcues.statemachine.StateMachine
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.withContext
 import org.koin.core.component.KoinScopeComponent
 import org.koin.core.component.inject
 import org.koin.core.scope.Scope
 import java.util.Date
 
 internal class ExperienceLifecycleTracker(
-    override val scope: Scope
+    override val scope: Scope,
 ) : KoinScopeComponent {
 
     // lazy property injection to avoid circular DI reference in constructor
     // AnalyticsTracker -> this <- Analytics Tracker
     private val analyticsTracker: AnalyticsTracker by inject()
-    private val stateMachine: StateMachine by inject()
     private val storage: Storage by inject()
+    private val appcuesCoroutineScope: AppcuesCoroutineScope by inject()
 
-    suspend fun start(dispatcher: CoroutineDispatcher = Dispatchers.IO): Unit = withContext(dispatcher) {
-        launch { observeState() }
-        launch { observeErrors() }
+    fun start(stateMachine: StateMachine, dispatcher: CoroutineDispatcher = Dispatchers.IO) {
+        appcuesCoroutineScope.launch(dispatcher) { stateMachine.observeState() }
+        appcuesCoroutineScope.launch(dispatcher) { stateMachine.observeErrors() }
     }
 
-    private suspend fun observeState() {
-        stateMachine.stateFlow.collect {
+    private suspend fun StateMachine.observeState() {
+        stateFlow.collect {
             if (it.shouldTrack()) { // will not track for unpublished (preview) experiences
                 when (it) {
                     is RenderingStep -> {
@@ -74,13 +76,14 @@ internal class ExperienceLifecycleTracker(
         }
     }
 
-    private suspend fun observeErrors() {
-        stateMachine.errorFlow.collect {
+    private suspend fun StateMachine.observeErrors() {
+        errorFlow.collect {
             if (it.shouldTrack()) { // will not track for unpublished (preview) experiences
                 when (it) {
                     is Error.ExperienceError -> trackLifecycleEvent(ExperienceError(it))
                     is Error.StepError -> trackLifecycleEvent(StepError(it))
-                    is Error.ExperienceAlreadyActive -> Unit
+                    is ExperienceAlreadyActive -> Unit
+                    is RenderContextNotActive -> Unit
                 }
             }
         }
@@ -106,6 +109,7 @@ internal class ExperienceLifecycleTracker(
         when (this) {
             is Error.ExperienceError -> experience.published
             is Error.StepError -> experience.published
-            is Error.ExperienceAlreadyActive -> false
+            is ExperienceAlreadyActive -> false
+            is RenderContextNotActive -> false
         }
 }
--- a/appcues/src/main/java/com/appcues/statemachine/Error.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/Error.kt
@@ -1,12 +1,16 @@
 package com.appcues.statemachine
 
 import com.appcues.data.model.Experience
+import com.appcues.data.model.RenderContext
 import java.util.UUID
 
 internal sealed class Error(open val message: String) {
+
     val id: UUID = UUID.randomUUID()
 
-    data class ExperienceAlreadyActive(val ignoredExperience: Experience, override val message: String) : Error(message)
+    object ExperienceAlreadyActive : Error("Experience already active")
+
+    data class RenderContextNotActive(val renderContext: RenderContext) : Error("RenderContext $renderContext is not active.")
     data class ExperienceError(val experience: Experience, override val message: String) : Error(message)
     data class StepError(val experience: Experience, val stepIndex: Int, override val message: String) : Error(message)
 }
--- a/appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
+++ b/appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
@@ -3,6 +3,7 @@ package com.appcues.ui
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.appcues.AppcuesCoroutineScope
+import com.appcues.action.ActionProcessor
 import com.appcues.action.ExperienceAction
 import com.appcues.analytics.ExperienceLifecycleEvent.StepInteraction.InteractionType
 import com.appcues.analytics.SdkMetrics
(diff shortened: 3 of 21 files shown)


appcues/src/main/java/com/appcues/statemachine/Error.kt
```diff
@@ -1,12 +1,15 @@
...
+import com.appcues.data.model.RenderContext
 import java.util.UUID
 
 internal sealed class Error(open val message: String) {
+
     val id: UUID = UUID.randomUUID()
 
-    data class ExperienceAlreadyActive(val ignoredExperience: Experience, override val message: String) : Error(message)
+    data class ExperienceAlreadyActive(val experience: Experience) : Error("Experience ${experience.name} already active")
+    data class RenderContextNotActive(val renderContext: RenderContext) : Error("RenderContext $renderContext is not active.")
```
andretortolano, line 12: small quality of life change here, since both errors ExperienceAlreadyActive and RenderContextNotActive only have a single Message, we can abstract it and have a hardcoded value here

appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
```diff
@@ -109,11 +111,7 @@ internal class AppcuesViewModel(
     }
 
     fun onActions(actions: List<ExperienceAction>, interactionType: InteractionType, viewDescription: String?) {
-        uiState.value.let { state ->
-            if (state is Rendering) {
-                experienceRenderer.process(state.experience.renderContext, actions, interactionType, viewDescription)
-            }
-        }
+        actionProcessor.process(renderContext, actions, interactionType, viewDescription)
```
andretortolano, line 114: this is how it was before, instead of calling experienceRenderer we can leave actionProcessor here for now since the process call had the same signature.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -2,91 +2,109 @@ package com.appcues.ui
...
-        val priorityOverride = experience.priority == NORMAL && stateMachine.state != Idling
-        if (priorityOverride) {
-            return dismiss(RenderContext.Modal, markComplete = false, destroyed = false).run {
-                when (this) {
-                    is Success -> show(experience) // re-invoke show on the new experience now after dismiss
-                    is Failure -> false // dismiss failed - can't continue
-                }
-            }
-        }
+        return this != null && this != newExperience && newExperience.priority == NORMAL
```
andretortolano, line 94: I thought about it when I was trying to make a test work. not sure if this is ok so I want to confirm.

instead of checking the stateMachine state, we can (I think) check the exiting experience (this) is different than null, and if (this) is different than the incoming experience.

this works well because during tests we init a StateMachine with a specific state like RenderingStep(experience ...) then since its the same the dismiss is ignored keeping that mocked machine on its current state even if the priority of the new experience is set to NORMAL. 

again if this looks wrong we can revert it and make sure that the mocked experience in tests are set to LOW (that works too)

appcues/src/test/java/com/appcues/action/ActionProcessorTest.kt
```diff
@@ -153,8 +157,10 @@ internal class ActionProcessorTest : KoinTest {
                                 experienceRenderer = get(),
                             )
                         }
-                        scoped { StateMachine(get(), get(), get(), initState) }
+                        factory { StateMachine(get(), get(), get(), initState) }
```
andretortolano, line 160: changing here from scoped to factory makes so whenever we "get" its a new instance.

appcues/src/main/java/com/appcues/statemachine/Error.kt
iujames: I think a subtle bug here maybe - the parameter for ExperienceAlreadyActive is the ignored experience - `ignoredExperience` and the message is now saying that experience is already active, rather than the name of the experience that is active? I guess it should just go back to the previous value
```
"Experience already active"
```
from L135 of StateMachine previous version, and keep the name `ignoredExperience` to make that clear

appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
iujames: Is it a concern that it is no longer checking if the uiState is Rendering? What do you mean by - how it was before?

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -2,91 +2,109 @@ package com.appcues.ui
...
     private val config by inject<AppcuesConfig>()
     private val analyticsTracker by inject<AnalyticsTracker>()
-    private val actionProcessor by inject<ActionProcessor>()
+    private val lifecycleTracker by inject<ExperienceLifecycleTracker>()
+
+    private val slots: HashMap<RenderContext, StateMachine> = hashMapOf()
 
     fun getStateFlow(renderContext: RenderContext): SharedFlow<State> {
-        return stateMachine.stateFlow
+        return slots[renderContext]?.stateFlow ?: MutableSharedFlow()
```
iujames, line 47: I think I'd rather return `null` on an invalid `renderContext` and make that explicit to the caller to handle

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -2,91 +2,109 @@ package com.appcues.ui
...
     }
 
-    fun process(renderContext: RenderContext, actions: List<ExperienceAction>, interactionType: InteractionType, viewDescription: String?) {
-        actionProcessor.process(renderContext, actions, interactionType, viewDescription)
+    fun getState(renderContext: RenderContext): State {
+        return slots[renderContext]?.state ?: Idling
```
iujames, line 51: similarly, can return `State?` and allow `null` so that the caller is responsible for handling issues correctly rather than hiding it with a fake State?

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -2,91 +2,109 @@ package com.appcues.ui
...
     // AnalyticsTracker > AnalyticsQueueProcessor > ExperienceRenderer(this) > AnalyticsTracker
     private val repository by inject<AppcuesRepository>()
-    private val stateMachine by inject<StateMachine>()
     private val sessionMonitor by inject<SessionMonitor>()
     private val config by inject<AppcuesConfig>()
     private val analyticsTracker by inject<AnalyticsTracker>()
-    private val actionProcessor by inject<ActionProcessor>()
+    private val lifecycleTracker by inject<ExperienceLifecycleTracker>()
+
+    private val slots: HashMap<RenderContext, StateMachine> = hashMapOf()
```
iujames, line 44: seems ok for now but interested in how this may evolve - I think we'll actually need the reference to the state machine owner - the view for embed case - similar to how iOS is keeping the `WeakStateMachineOwning` reference here. Eventually, the embed trait will need to be able to get the view reference for a given render context. But this can evolve later.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
iujames: I need to wrap my head around this, im not sure. My gut instinct is to retain existing behavior wherever possible to minimize chance of regression, but I don't fully understand the change yet.

iujames (commented):
I think looking pretty good to me, made a first pass with some thoughts but I need to think over some details a bit more and have to go pickup my son, so will get back to it!

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -20,29 +23,28 @@ import com.appcues.statemachine.StateMachine
...
-        launch { observeErrors() }
+    fun start(stateMachine: StateMachine, dispatcher: CoroutineDispatcher = Dispatchers.IO) {
+        appcuesCoroutineScope.launch(dispatcher) { stateMachine.observeState() }
+        appcuesCoroutineScope.launch(dispatcher) { stateMachine.observeErrors() }
     }
 
-    private suspend fun observeState() {
-        stateMachine.stateFlow.collect {
+    private suspend fun StateMachine.observeState() {
+        stateFlow.collect {
```
iujames, line 47: Is there a way to end this collection on the flow when the state machine completes? Wondering if this will leave any memory references alive indefinitely now that we have the potential for multiple state machines during the app lifetime.

I guess a similar situation/question exists for the AppcuesViewModel usage as well.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
iujames: OK I went through this again and I agree I think it is logically the same. Since the only case where `stateMachine.state.currentExperience` is `null` i the `Idling` case, checking for `this != null` has the same effect as checking `stateMachine.state != Idling` in the previous logic.

I am curious which test exactly needed the change and how it worked before without the change, just to understand that part.

andretortolano: thanks for doublechecking on that. I added more context in [here](https://appcues.slack.com/archives/C031RLGS4F8/p1689601592014059)

appcues/src/main/java/com/appcues/statemachine/Error.kt
andretortolano: it makes sense, since we are eventually logging this one I wish we could use the ignoredExperience value as part of the message, or else we could just remove the val ignoredExperience from the class since its not used anywhere.

I was maybe thinking of"

"Experience already active, experience ..... ignored."

appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
andretortolano: in one of the last PRs this was just a single line calling the processor, i changed that but it was not really necessary as the method signature remained the same for no extra benefit, actionProcessor also requires renderContext as its not tied to a single StateMachine. not sure if we are going to change that but for now we could just continue using the actionProcessor here instead of experienceRenderer.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
andretortolano: not sure how thats going to work yet

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
andretortolano: yes once we get to the point were we want to clean state machines lets say on Screen calls, we can then store those launch into Job variables and cancel those whenever we want

appcues/src/main/java/com/appcues/statemachine/Error.kt
```diff
@@ -1,12 +1,16 @@
...
 import com.appcues.data.model.Experience
+import com.appcues.data.model.RenderContext
 import java.util.UUID
 
 internal sealed class Error(open val message: String) {
+
     val id: UUID = UUID.randomUUID()
 
-    data class ExperienceAlreadyActive(val ignoredExperience: Experience, override val message: String) : Error(message)
+    class ExperienceAlreadyActive : Error("Experience already active")
```
iujames, line 11: could be an `object` ?

iujames (approved):
Looks good! nice progress on the foundation pieces

appcues/src/main/java/com/appcues/statemachine/Error.kt
andretortolano: could, it is now

PR #434, opened 2023-07-17
Update to Gradle 8 and other related updates

iujames:
Updates related to the latest Android Studio Flamingo default versions, getting into compatibility with other tools releasing similar updates. A lot of these are cascading updates from the Gradle 8 change really, for example the Kotlin version which then requires a Compose Compiler / Compose update.

* Android Gradle Plugin 7.2.2 -> 8.0.2
* JDK 11 -> 17
* Kotlin 1.7.20 - 1.8.21
* Compose 1.3.0 -> 1.4.3
* Compose Compiler 1.3.2 -> 1.4.7
* Accompanist 0.27.0 -> 0.30.1

As part of Compose 1.4.3 - now using the official Pager for our carousel and removed the usage of the Accompanist Pager.

Ended up also needing to explicitly declare OkHttp 4.11.0 as a dependency to[ resolve a minification issue with the 4.10.0 version](https://github.com/square/okhttp/issues/7777) that was previously being implicitly included through other dependencies (Retrofit / Coil),

Also updated Koin 3.1.6 -> 3.4.2 to stay current (customer request). We had held back due to an Expo (react native) conflict, but this [appears to be updated now as well](https://github.com/expo/expo/commit/681f12047dfd3015c951bf54e2a2ba7f839d5b1f). There is a breaking issue with [how Koin handles parameters](https://github.com/InsertKoinIO/koin/issues/1607), so a change was made to use indexed based param lookup so that our code works on both Koin old and new versions.

Also other build/util tools and test lib updates
* Dokka (api doc gen) 1.7.20 -> 1.8.20
* Jacoco (test coverage report gen) 0.8.7 -> 0.8.9
* Updated unit test dependencies and verified tests (required a few small test code updates)

Working on related UI test updates in the spec library, Shot now supports JDK 17 and all the updates so hopefully that all works out.

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/appcues/CarouselTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/CarouselTrait.kt
@@ -1,7 +1,11 @@
 package com.appcues.trait.appcues
 
+import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
+import androidx.compose.foundation.pager.HorizontalPager
+import androidx.compose.foundation.pager.PagerState
+import androidx.compose.foundation.pager.rememberPagerState
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.mutableStateMapOf
@@ -16,10 +20,6 @@ import com.appcues.trait.ContentHolderTrait
 import com.appcues.trait.ContentHolderTrait.ContainerPages
 import com.appcues.ui.composables.AppcuesPaginationData
 import com.appcues.ui.composables.LocalAppcuesPaginationDelegate
-import com.google.accompanist.pager.ExperimentalPagerApi
-import com.google.accompanist.pager.HorizontalPager
-import com.google.accompanist.pager.PagerState
-import com.google.accompanist.pager.rememberPagerState
 import kotlinx.coroutines.flow.drop
 
 internal class CarouselTrait(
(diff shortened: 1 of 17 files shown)


appcues/src/main/java/com/appcues/trait/appcues/CarouselTrait.kt
```diff
@@ -1,7 +1,11 @@
 package com.appcues.trait.appcues
 
+import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
+import androidx.compose.foundation.pager.HorizontalPager
+import androidx.compose.foundation.pager.PagerState
+import androidx.compose.foundation.pager.rememberPagerState
```
iujames, line 8: the changes in this file are to use the new HorizontalPager and related items from Compose, instead of Accompanist

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
```diff
@@ -176,14 +176,14 @@ internal fun Activity.getParentView(): ViewGroup {
...
         val lifecycleOwner = window.decorView.findViewTreeLifecycleOwner()
-        ViewTreeLifecycleOwner.set(decorView, lifecycleOwner)
+        decorView.setViewTreeLifecycleOwner(lifecycleOwner)
     }
 
-    if (ViewTreeViewModelStoreOwner.get(decorView) == null) {
+    if (decorView.findViewTreeViewModelStoreOwner() == null) {
         val viewModelStoreOwner = window.decorView.findViewTreeViewModelStoreOwner()
-        ViewTreeViewModelStoreOwner.set(decorView, viewModelStoreOwner)
+        decorView.setViewTreeViewModelStoreOwner(viewModelStoreOwner)
```
iujames, line 186: this was an API change in a dependency that comes in through Compose somehow it seems - these are now functions on View rather than static functions on the classes.

appcues/src/test/java/com/appcues/data/AppcuesRepositoryTest.kt
```diff
@@ -140,6 +140,10 @@ internal class AppcuesRepositoryTest {
         val activityStorage = ActivityStorage(UUID.randomUUID(), "123", "userId", "data", null)
         coEvery { appcuesLocalSource.getAllActivity() } returns listOf(activityStorage)
 
+        // two lines below added related to https://github.com/mockk/mockk/issues/321
+        coEvery { appcuesRemoteSource.postActivity(any(), any(), any()) } returns Success(mockk(relaxed = true))
+        coEvery { appcuesRemoteSource.qualify(any(), any(), any(), any()) } returns Success(mockk(relaxed = true))
```
iujames, line 145: I had to update mockk, I assume for the Kotlin update, and without that many things failed.

However, it did introduce this weird issue in this class only where mocked return types with Generics have some unexpected behavior. In our case, its the ResultOf (Success/Failure) types used in the lower level network layer as the return value of `postActivity` and `qualify` here. It is basically exactly as noted in this issue https://github.com/mockk/mockk/issues/321 but this workaround to more explicitly set the mock response rather than using a relaxed mock works ok for now.

andretortolano (approved):
Im Ok with this change, I know there is some merging conflicts with my own PR, I can work on that later

PR #435, opened 2023-07-18
[WIP] - ♻️  Updating OverlayViewManager into RenderViewManager to support Embedded content

andretortolano:
This is an attempt to change OverlayViewManager into a more generic view handler to help different renderContext to exist on screen.

the main idea is that we we will ALWAYS need to add our custom view, so we can open URLs internall in the app, so changing from an approach to add/exclude our overlay view, we now add or get the existing, control visibility and (setContent) to ensure it stills work on regular modals/tooltips

* Also added Embeded Trait, but still needs to test that one out, wanted to put it out there for review of the idea, we can refine from this point


appcues/src/main/java/com/appcues/trait/appcues/EmbedTrait.kt
```diff
@@ -0,0 +1,140 @@
...
+        return slideOutVertically(tween(durationMillis = 250)) { it / slideOffsetDivider } +
+            scaleOut(tween(durationMillis = 250), targetScale = 0.85f) +
+            fadeOut(tween(durationMillis = 200))
+    }
+
+    override fun present() {
+        val success = RenderViewManager(scope, renderContext).start()
+
+        if (!success) {
+            throw AppcuesTraitException("unable to create modal overlay view")
```
iujames, line 137: copy/paste of `modal overlay`

appcues/src/main/java/com/appcues/ui/RenderViewManager.kt
```diff
@@ -0,0 +1,169 @@
...
+        val parentView = AppcuesActivityMonitor.activity?.getParentView()
+
+        parentView?.post {
+            parentView.findViewTreeLifecycleOwner()?.lifecycle?.removeObserver(this@RenderViewManager)
+
+            if (renderContext == Modal) {
+                parentView.getOverlayLayoutBinding().let {
+                    it.appcuesOverlayComposeView.isVisible = false
+                    it.appcuesOverlayComposeView.setContent { }
+                }
```
iujames, line 152: I guess I'm not yet understanding this part from the description

> the main idea is that we we will ALWAYS need to add our custom view, so we can open URLs internall in the app

Could you say more about that?

appcues/src/main/java/com/appcues/trait/appcues/EmbedTrait.kt
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    private fun exitTransitionDerivedOf(windowInfo: AppcuesWindowInfo): State<ExitTransition> {
+        return derivedStateOf {
+            when (windowInfo.deviceType) {
+                MOBILE -> exitTransition()
+                TABLET -> dialogExitTransition()
+            }
+        }
+    }
```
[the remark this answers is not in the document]
andretortolano: yes you are right, upon trying to run UI tests and reflecting I will be probably breaking this change into a smaller change for the moment

PR #438, opened 2023-07-20
♻️  Updating OverlayViewManager removing appcues_overlay_layout

andretortolano:
This took longer than it should to get it done, but small details here and there between test runs pushed this off until I got it right.

* Refactoring of OverlayViewManager including the removal of overlay_layout (creating ComposeView instead)

This is not a dependency per say for embeds but its a good change to come before since we are not using the external xml layout to provide a view for Chrome Client


appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
```diff
@@ -1,195 +1,129 @@
...
     private val renderContext: RenderContext
 ) : DefaultLifecycleObserver, ActivityMonitorListener {
 
-    private val logcues: Logcues by lazy { scope.get() }
-
-    private var viewModel: AppcuesViewModel? = null
-
-    private var shakeGestureListener: ShakeGestureListener? = null
+    private lateinit var viewModel: AppcuesViewModel
+    private lateinit var gestureListener: ShakeGestureListener
```
andretortolano, line 30: setting those as lateinit var will work the same as we had before, but this way we can avoid calling "!!" when creating the Composition

appcues/src/main/java/com/appcues/ui/utils/ActivityExt.kt
```diff
@@ -0,0 +1,67 @@
...
+import androidx.lifecycle.ViewTreeViewModelStoreOwner
+import androidx.lifecycle.findViewTreeLifecycleOwner
+import androidx.lifecycle.findViewTreeViewModelStoreOwner
+import androidx.savedstate.findViewTreeSavedStateRegistryOwner
+import androidx.savedstate.setViewTreeSavedStateRegistryOwner
+import java.lang.reflect.Method
+
+@Suppress("UNCHECKED_CAST")
+@SuppressLint("PrivateApi")
+internal fun Activity.getParentView(): ViewGroup {
```
andretortolano, line 20: moved it into a different file since its being used in other places and its generic enough to be extracted

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
iujames: What if I background and foreground without ever having an experience shown (no call to `start()`) - wouldn't that end up in `onResume` and reference `gestureListener` before it was initialized, causing an exception?

andretortolano: no because even though we are observing the lifecycle we only do it after we create the instances.

https://github.com/appcues/appcues-android-sdk/pull/438/files#diff-574c69d3146da26d9f6652cfdfcc11b4357f3dec4e153f40de758d6306794472R68

iujames: ah! right, thank you!

PR #439, opened 2023-07-24
✨ Added EmbedTrait

andretortolano:
This PR is quite big but here we are working on a couple of things.

* Adding EmbedTrait
-> New file similar to ModalTrait, WrapContent like a Dialog presentation style would. 
-> Using newly added EmbedViewPresenter to inflate the Compose view we need to inflate our experience

* Changed ModalTrait and TooltipTrait 
->  Both are now using the refreshed version of OverlayViewManager -> OverlayViewPresenter

* Created ViewPresenter as an abstract class that is extended by OverlayViewPresenter and EmbedViewPresenter

* Changed ExperienceRender show method to account for different type of render contexts

* Other small changes to make embeds work for the initial version  
-------

As a plus, I added a new Tab in the sample app to mirror IOS, this one is a initial version just so we can have embed views registered for initial debugging and checking.

image


appcues/src/main/java/com/appcues/trait/TraitKoin.kt
```diff
@@ -27,6 +28,7 @@ internal object TraitKoin : KoinScopePlugin {
         factory { params -> BackdropKeyholeTrait(params.getOrNull()) }
         factory { params -> ModalTrait(params.getOrNull(), params.get(), get()) }
         factory { params -> TooltipTrait(params.getOrNull(), params.get(), get()) }
+        factory { params -> EmbedTrait(params.getOrNull(), params.get(), get()) }
```
iujames, line 31: based on prior Koin convo and [change in PR](https://github.com/appcues/appcues-android-sdk/pull/434/commits/de3818eda79d1439e85bf9095bae7630ae3f2bfc#diff-f0670ef1ed9e62e4df09f5d2a4cc4680206e0a3629a9ff828f1dec73260c174a), we'll want to evolve to
```
EmbedTrait(it[0], it[1], get())
```

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -107,15 +108,20 @@ internal class ExperienceRenderer(
...
+        val modalSuccess = show(qualifiedExperiences.first { it.renderContext is RenderContext.Modal })
+        if (!modalSuccess) {
             val remainingExperiences = qualifiedExperiences.drop(1)
             if (remainingExperiences.isNotEmpty()) {
                 // fallback logic - try the next remaining experience, if available
                 return show(remainingExperiences)
             }
         }
-        return success
+        return modalSuccess
```
iujames, line 124: if I'm following correctly - we'll need to trim out the embed experiences before doing any sort of recursion on line 121 to try lower priority modal experiences - so the embeds don't get re-triggered on line 113?

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -107,15 +108,20 @@ internal class ExperienceRenderer(
             return true
         }
 
-        val success = show(qualifiedExperiences.first())
-        if (!success) {
+        qualifiedExperiences.filter { it.renderContext is RenderContext.Embed }.forEach {
+            // TODO what to do here
+            show(it)
+        }
```
iujames, line 114: using iOS as a guide, this is where the basic caching stuff comes into play
* keep some set of potentially renderable experiences https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift#L76
* try and show things right away, if possible https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift#L78-L80
* remove things from cache when an experience is dismissed/completed https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift#L232

The full set of rules is written up here for reference https://github.com/appcues/appcues-mobile-experience-spec/blob/main/rfc/embedded-experiences.md#processing-embeds-in-qualification-response

I think these behaviors can be layered on after this PR though as a next step, no issue with that.

appcues/src/main/java/com/appcues/ui/EmbedViewManager.kt
```diff
@@ -0,0 +1,90 @@
...
+import com.appcues.RenderContextManager
+import com.appcues.data.model.RenderContext
+import com.appcues.monitor.AppcuesActivityMonitor
+import com.appcues.monitor.AppcuesActivityMonitor.ActivityMonitorListener
+import com.appcues.ui.composables.AppcuesComposition
+import com.appcues.ui.primitive.EmbedChromeClient
+import com.appcues.ui.utils.getParentView
+import org.koin.core.scope.Scope
+
+internal class EmbedViewManager(
```
iujames, line 18: curious what led to the modal/embed split here versus the previous OverlayViewManager. Looks like some duplication but I'm guessing other things led to too much complication when trying to combine?

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
andretortolano: you are right here

appcues/src/main/java/com/appcues/ui/EmbedViewManager.kt
andretortolano: this was a first pass, I tried to not mess with appcuesOverlay too much, I can come back now and see if there is a better way to structure it.

but one thing I'm thinking here is that maybe we would want to have different classes instead of 1 class with too much responsibility

appcues/src/main/java/com/appcues/trait/ContentHolderTrait.kt
```diff
@@ -54,19 +54,22 @@ public interface ContentHolderTrait : ExperienceTrait {
         /**
          * set the pagination data based on custom [CreateContentHolder] implementation
          */
-        public fun setPaginationData(paginationData: AppcuesPaginationData) {
-            appcuesPaginationData.value = paginationData
+        @Composable
```
andretortolano, line 57: this is bad news, somehow we need to update the pagination data, that includes changing this method to be composable and this is a change in our API.

appcues/src/main/java/com/appcues/trait/ContentHolderTrait.kt
```diff
@@ -54,19 +54,22 @@ public interface ContentHolderTrait : ExperienceTrait {
...
 
         /**
          * the SDK will invoke this function to ensure we have the minimum
          * required [AppcuesPaginationData] that can be used by other traits
-         * in case no [setPaginationData] was called during [CreateContentHolder]
+         * in case no [UpdatePaginationData] was called during [CreateContentHolder]
          */
-        internal fun syncPaginationData() {
+        @Composable
+        internal fun SyncPaginationData() {
```
andretortolano, line 70: the other API change is here (same class)

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
iujames: yes, seems totally fine to me - makes sense to layer that cache behavior on top in the next round of work.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
andretortolano: the show(experiences) method doesnt need to return true or false.

created two separate functions for embeds and modal, embeds will evolve later, modal is now doing what the original show was doing. 

trying to show a modal experience, if it fails it will drop that and call it again until the list is empty.

appcues/src/main/java/com/appcues/RenderContextManager.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    private val slots: HashMap<RenderContext, WeakReference<ViewGroup>> = hashMapOf()
+
+    fun registerEmbedFrame(frameId: String, view: ViewGroup) {
+        slots[Embed(frameId)] = WeakReference(view)
+    }
+
+    fun getEmbedView(renderContext: RenderContext): ViewGroup? {
+        return when (renderContext) {
+            is Embed -> slots[renderContext]?.get().also { if (it == null) slots.remove(renderContext) }
```
iujames, line 18: this is the part where we would not actually call `slots.remove()` yet, based on convo about cell re-use (recyclers), correct?

appcues/src/main/java/com/appcues/trait/ContentHolderTrait.kt
iujames: Is the change in function name just to get upper case start for Composable convention?

Maybe this becomes a 3.0 release, we'll have to chat about that, i guess so. It's certainly not part of the public API that I suspect anyone is using, but still. But that's ok.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -100,22 +100,26 @@ internal class ExperienceRenderer(
...
-        if (!success) {
-            val remainingExperiences = qualifiedExperiences.drop(1)
-            if (remainingExperiences.isNotEmpty()) {
-                // fallback logic - try the next remaining experience, if available
-                return show(remainingExperiences)
-            }
+    private suspend fun showEmbeds(experiences: List<Experience>) {
+        experiences.forEach {
+            // TODO what to do hereª
+            show(it)
```
iujames, line 121: I think this is in progress based on prev convos - but lmk if other open questions. Basically, show the ones we can immediately, and the others have experience_error and then can later have experience_recovered if the view shows up.

appcues/src/main/java/com/appcues/ui/presentation/EmbedViewPresenter.kt
```diff
@@ -0,0 +1,28 @@
...
+        val embedView = renderContextManager.getEmbedView(renderContext) ?: return null
+
+        return ComposeView(context).also {
+            embedView.addView(it)
+        }
+    }
+
+    override fun ViewGroup.removeView() {
+        post {
+            renderContextManager.getEmbedView(renderContext)?.removeAllViews()
```
iujames, line 25: is there any chance that other views might exist in the container and we should only remove our own? I'm guessing no, but checking.

iujames (commented):
nice! I like how you pulled the common parts out of ViewPresenter and then have the overlay vs embed derived types 👍

This all seems great progress - my only open question is on the 3.0 public API change. I'm wondering if we should keep this as a (semi) long running embeds feature branch and do PRs into it rather than merge to main then? In case we need to do some patch releases in the meantime with small changes to the 2.x code?

appcues/src/main/java/com/appcues/trait/ContentHolderTrait.kt
andretortolano: PRing this to feature/embeds instead of main for now!

appcues/src/main/java/com/appcues/RenderContextManager.kt
andretortolano: good catch

appcues/src/main/java/com/appcues/ui/presentation/EmbedViewPresenter.kt
andretortolano: Maybe it would be better off creating our own ViewGroup child like AppcuesEmbeddedContent: FrameLayout

iujames (approved):
approving this one to merge over to `feature/embeds` as our feature branch for the 3.0.0 SDK work (for any public API changes like the `ContentHolderTrait`

PR #441, opened 2023-07-24
Update build machine to latest Android for all jobs

iujames:
This approach should ensure that we always have a consistent Android build machine specified across all the jobs.

Previously, I accidentally only updated the machine in the `validate_code` section (PR validation) but missed the `update_code_coverage` section, resulting in the [build failure here](https://app.circleci.com/pipelines/github/appcues/appcues-android-sdk/1947/workflows/add1b292-d035-4905-9f30-96d75f410d24/jobs/1836).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,16 @@ orbs:
   codecov: codecov/codecov@3.1.1
 
 # -------------------------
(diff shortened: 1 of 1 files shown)


andretortolano (approved):
👍

PR #442, opened 2023-07-27
✨ Added AppcuesFrameView to replace generic ViewGroup

andretortolano:
An additional PR to the original embed trait PR.

this one is aiming to replace the generic ViewGroup with a specific AppcuesFrameView, the reasoning behind it is to limit customer options here, not letting them adding ANY type of ViewGroup, in addition this is extending from AbstractComposeView, with similar implementation as ComposeView, which should be perfect for our use case.

deriving two classes, one internal and the other (AppcuesFrameView) public for use during appcues.registerEmbed("frame", appcuesFrameView)

### DIFF
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -3,7 +3,6 @@ package com.appcues
 import android.app.Activity
 import android.content.Context
 import android.content.Intent
-import android.view.ViewGroup
 import com.appcues.action.ActionRegistry
 import com.appcues.action.ExperienceAction
 import com.appcues.analytics.ActivityScreenTracking
@@ -237,13 +236,13 @@ public class Appcues internal constructor(koinScope: Scope) {
     }
 
     /**
-     * register an embed view for given frameId.
+     * Register an AppcuesFrameView for a given frameId.
      *
-     * @param frameId unique string set in builder that is used to identify this embed view with qualified experiences
-     * @param view frame used to inflate the embedded experience
+     * @param frameId unique string set in builder that is used to identify this AppcuesFrameView with qualified experiences
+     * @param frame frame used to inflate the embedded experience
      */
-    public fun registerEmbed(frameId: String, view: ViewGroup) {
-        renderContextManager.registerEmbedFrame(frameId, view)
+    public fun registerEmbed(frameId: String, frame: AppcuesFrameView) {
+        renderContextManager.registerEmbedFrame(frameId, frame)
     }
 
     /**
--- a/appcues/src/main/java/com/appcues/AppcuesViews.kt
+++ b/appcues/src/main/java/com/appcues/AppcuesViews.kt
@@ -1,0 +1,100 @@
+package com.appcues
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
+import android.widget.FrameLayout
+import androidx.compose.ui.platform.ComposeView
+
+/**
+ * AppcuesFrameView should is used when customers want to define a specific place for inflating Embed content
+ * in their app. By placing an AppcuesFrameView in their layout, then can later register this view with a frameId.
+ *
+ * appcues.registerEmbed("frame1", appcuesFrameView)
+ */
+public class AppcuesFrameView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : FrameLayout(context, attrs, defStyleAttr) {
+
+    private val composeView = ComposeView(context)
+
+    internal fun setupComposeView(): ComposeView {
+        // when index is -1 it means the view is not added yet
+        if (indexOfChild(composeView) == -1) {
+            addView(composeView)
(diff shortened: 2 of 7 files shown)


appcues/build.gradle
```diff
@@ -83,7 +83,7 @@ dependencies {
     implementation "androidx.browser:browser:1.4.0"
     implementation "androidx.startup:startup-runtime:1.1.1"
     implementation "androidx.navigation:navigation-compose:2.5.3"
-    implementation "androidx.compose.ui:ui:$compose_version"
+    api "androidx.compose.ui:ui:$compose_version"
```
andretortolano, line 86: extending AbstractComposeView require us to change here from implementation to api, meaning we are including the compose dependency in our graph, this is not a big issue since people should already have compose implemented. just noting here the reason why this was changed.

iujames: does this open any concern about version conflicts with the version we are using versus the customer app? Wondering if we can conceal this by just having an `AppcuesFrameView` that is a specific `View` that we provide that internally contains the compose view?

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -242,7 +241,7 @@ public class Appcues internal constructor(koinScope: Scope) {
      * @param frameId unique string set in builder that is used to identify this embed view with qualified experiences
      * @param view frame used to inflate the embedded experience
      */
-    public fun registerEmbed(frameId: String, view: ViewGroup) {
+    public fun registerEmbed(frameId: String, view: AppcuesEmbedView) {
```
iujames, line 244: we've been referring to this as the `Frame` in documents, and the `frameID` in the json. iOS calls it the `AppcuesFrameView` and has a register frame function - I think would be ideal to stick to that nomenclature.

iujames (commented):
No issue with having the `AppcuesComposeView` - just wondering if we can encapsulate/hide that detail from the consumer a bit more, just exposing an `AppcuesFrameView` (a FrameLayout perhaps?) that keeps the compose stuff internal in a subview

appcues/src/main/java/com/appcues/Appcues.kt
andretortolano: can surely change from AppcuesEmbedView to AppcuesFrameView

appcues/src/main/java/com/appcues/AppcuesViews.kt
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Base class for Appcues Compositions, provides a similar shell as native ComposeView but it adds internal
+ * flexibility since ComposeView is final.
+ */
+public abstract class AppcuesComposeView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : FrameLayout(context, attrs, defStyleAttr) {
```
andretortolano, line 18: using FrameLayout now instead of AbstractComposeView, but that entails adding some extra restraints on addView(***) methods, same thing that AbstractComposeView does to ensure no other view is added here.

I'm also adding and removing everytime we set/clear content to make sure that customers wont removeAllViews or removewView(index = 0) and break our Custom Layout

appcues/src/main/java/com/appcues/AppcuesViews.kt
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    internal fun clearComposition() {
+        super.removeAllViews()
+        composeView.setContent { }
+    }
+
+    // Below: enforce restrictions on adding child views to this ViewGroup
+    private fun checkAddView(child: View?) {
+        if (child != composeView) {
```
andretortolano, line 38: the only child we want to add is the composeView

iujames: Just an opinion, but I'd personally prefer the SDK not be opinionated here and simply provide the view with the understanding that if you mess with adding contents it is at your own peril (no exception throwing). Basically, not need any of these `checkAddView` calls and override safeguards. In practice, probably no one will do it and it won't matter either way - so it's harmless to include probably.

appcues/src/main/java/com/appcues/AppcuesViews.kt
```diff
@@ -0,0 +1,96 @@
...
+public class AppcuesFrameView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : FrameLayout(context, attrs, defStyleAttr) {
+
+    private val composeView = ComposeView(context)
+
+    internal fun setupComposeView(): ComposeView {
+        addView(composeView)
```
iujames, line 24: maybe come back to this - but wondering if there is ever a case where we need to check if it is already added before adding again? Thinking of the recycler view case specifically - but maybe not an issue since a new frame view would get created on the new view inflate, or if something was re-used, it would have `clearComposition` called on it prior to reuse.

iujames (approved):
Looks good to me! Thanks for going the extra step to minimize the public API exposure!

PR #445, opened 2023-08-03
Fix image layout issues related to aspectRatio handling

iujames:
This is the result of several days of exploration into our layout structures after discovering some Paparazzi snapshot tests were failing, then discovering that same layouts actually were not working in our SDK after the change to use the overlay view in the customer Activity instead of the old AppcuesActivity approach.

There are several bits of cleanup here, but to summarize: we needed a better `.aspectRatio` modifier. This is what `AspectRatioContentModeModifier` now handles. It properly accounts for FIT vs FILL, as well as fixed content width/height values when determining the resulting size. It also now handles intrinsic sizes properly so that we can use `.height(Min)` on our `Row` elements, and have side by side content size correctly based on the sibling heights.

There is no longer a need to track the StackScope children and their size as we had in the previous aspect ratio implementation, so this is simplified to be just an enum value and it is now only used in one spot - the `Spacer` primitive (which is not really used at this point either). Our layout should now calculate and render in a single pass, which is how Compose desires it to be, and not require any `.onSizeChanged` tracking of children as was previously done in `ExperiencePrimitive` (removed).

I've done some pretty extensive testing with our snapshots, and will continue to do so for a bit, but I think this is ready to go now. I'll link a video separately since it is too big, but the Paparazzi snapshots now match the expected iOS and the previous Shot versions.

Linking a few references here just for background info / future reference
* simple example that talks about how IntrinsicSize.Min on the Row works https://stackoverflow.com/a/67678041
* more details about filling width in row https://medium.com/rocknnull/jetpack-compose-filling-max-width-or-height-94e3af129a7b
* a custom aspectRatio modifier that was some inspiration to re-work ours https://proandroiddev.com/aspect-ratio-with-reference-in-jetpack-compose-9bf9b79016ee
* info about controlling intrinsic size values that I ended up using some in our custom modifier https://developer.android.com/jetpack/compose/layouts/intrinsic-measurements


appcues/src/main/java/com/appcues/ui/primitive/HorizontalStackPrimitive.kt
```diff
@@ -39,11 +41,11 @@ internal fun HorizontalStackPrimitive.Compose(modifier: Modifier) {
 
     val verticalAlignment = style.getVerticalAlignment(CenterVertically)
     Row(
-        modifier = modifier,
+        modifier = modifier.height(IntrinsicSize.Min),
```
iujames, line 44: the key update for `Row` to handle side by side sizing

andretortolano (approved):
Good work getting this resolved, also good changes towards a more simple code

PR #446, opened 2023-08-04
♻️  ExperienceRenderer changes around how embeds are displayed

andretortolano:
the goal here is to cover some of the edge cases for embeds when its registered an re-registered.

the initial scenario I was going for is to have proxy experiences to show on Embeds tab, and then interacting with those and changing between tabs should be able to work well.

for next step I will be looking into RecyclerViews

### DIFF
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -74,7 +74,6 @@ public class Appcues internal constructor(koinScope: Scope) {
     private val debuggerManager by koinScope.inject<AppcuesDebuggerManager>()
     private val appcuesCoroutineScope by koinScope.inject<AppcuesCoroutineScope>()
     private val analyticsPublisher by koinScope.inject<AnalyticsPublisher>()
-    private val renderContextManager by koinScope.inject<RenderContextManager>()
 
     /**
      * Set the listener to be notified about the display of Experience content.
@@ -242,7 +241,9 @@ public class Appcues internal constructor(koinScope: Scope) {
      * @param frame frame used to inflate the embedded experience
      */
     public fun registerEmbed(frameId: String, frame: AppcuesFrameView) {
-        renderContextManager.registerEmbedFrame(frameId, frame)
+        appcuesCoroutineScope.launch {
+            experienceRenderer.startFrame(frameId, frame)
+        }
     }
 
     /**
--- a/appcues/src/main/java/com/appcues/RenderContextManager.kt
+++ b/appcues/src/main/java/com/appcues/RenderContextManager.kt
@@ -1,25 +1,80 @@
 package com.appcues
 
+import com.appcues.analytics.ExperienceLifecycleTracker
+import com.appcues.data.model.Experience
+import com.appcues.data.model.ExperienceTrigger
+import com.appcues.data.model.ExperienceTrigger.Qualification
 import com.appcues.data.model.RenderContext
 import com.appcues.data.model.RenderContext.Embed
+import com.appcues.statemachine.StateMachine
+import org.koin.core.component.KoinScopeComponent
+import org.koin.core.component.get
+import org.koin.core.component.inject
+import org.koin.core.scope.Scope
 import java.lang.ref.WeakReference
 
-internal class RenderContextManager {
+internal class RenderContextManager(override val scope: Scope) : KoinScopeComponent {
 
-    private val slots: HashMap<RenderContext, WeakReference<AppcuesFrameView>> = hashMapOf()
+    private val lifecycleTracker by inject<ExperienceLifecycleTracker>()
+
+    private val frameSlots: HashMap<RenderContext, WeakReference<AppcuesFrameView>> = hashMapOf()
+
+    private val stateMachineSlots: HashMap<RenderContext, StateMachine> = hashMapOf()
+
+    private val potentiallyRenderableExperiences = hashMapOf<RenderContext, List<Experience>>()
 
     fun registerEmbedFrame(frameId: String, frame: AppcuesFrameView) {
-        slots[Embed(frameId)] = WeakReference(frame)
+        val renderContext = Embed(frameId)
+
+        // silently removing the existing state machine to ensure we get a new one when trying
+        // to start an experience for this render context
+        stateMachineSlots.remove(renderContext)
+        frameSlots[Embed(frameId)] = WeakReference(frame)
     }
 
     fun getEmbedFrame(renderContext: RenderContext): AppcuesFrameView? {
         return when (renderContext) {
-            is Embed -> slots[renderContext]?.get()
+            is Embed -> frameSlots[renderContext]?.get()
             else -> null
         }
     }
 
-    fun clear() {
-        slots.clear()
+    fun getOrCreateStateMachines(renderContext: RenderContext): StateMachine {
+        return stateMachineSlots[renderContext] ?: run {
+            get<StateMachine>()
+                .also { stateMachineSlots[renderContext] = it }
+                .also { lifecycleTracker.start(it, { potentiallyRenderableExperiences.remove(renderContext) }) }
+        }
+    }
+
+    fun getStateMachine(renderContext: RenderContext): StateMachine? {
+        return stateMachineSlots[renderContext]
+    }
+
+    fun putExperiences(experiences: List<Experience>, trigger: ExperienceTrigger) {
+        if (trigger is Qualification && trigger.reason == "screen_view") {
+            // clear list in case this was a screen_view qualification
+            potentiallyRenderableExperiences.clear()
+
+            stateMachineSlots.forEach { map ->
+                val renderContext = map.key
+                if (renderContext is Embed && getEmbedFrame(map.key) == null) {
+                    stateMachineSlots.remove(map.key)
+                }
+            }
+        }
+
+        // populate new experiences into the dictionary
+        experiences.filter { it.renderContext !is RenderContext.Modal }.groupBy { it.renderContext }.forEach {
+            potentiallyRenderableExperiences[it.key] = it.value
+        }
+    }
+
+    fun getPotentialExperiences(renderContext: RenderContext): List<Experience> {
+        return potentiallyRenderableExperiences[renderContext] ?: arrayListOf()
+    }
+
+    fun stop() {
+        stateMachineSlots.values.forEach { it.stop() }
     }
 }
--- a/appcues/src/main/java/com/appcues/analytics/AnalyticsTrackerExt.kt
+++ b/appcues/src/main/java/com/appcues/analytics/AnalyticsTrackerExt.kt
@@ -1,7 +1,9 @@
 package com.appcues.analytics
 
+import com.appcues.data.model.Experience
 import com.appcues.data.model.Experiment
 import com.appcues.util.appcuesFormatted
+import java.util.UUID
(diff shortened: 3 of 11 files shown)


appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -243,6 +244,10 @@ public class Appcues internal constructor(koinScope: Scope) {
      */
     public fun registerEmbed(frameId: String, frame: AppcuesFrameView) {
         renderContextManager.registerEmbedFrame(frameId, frame)
+
+        appcuesCoroutineScope.launch {
+            experienceRenderer.show(RenderContext.Embed(frameId))
+        }
```
iujames, line 250: possibly could be one call into `ExperienceRenderer` which would then register the frame and start it up right away, if possible - similar to https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Appcues.swift#L259

appcues/src/main/java/com/appcues/analytics/AnalyticsTrackerExt.kt
```diff
@@ -20,3 +22,33 @@ internal fun AnalyticsTracker.track(experiment: Experiment) {
...
+    experience.renderErrorId = errorId
+}
+
+internal fun AnalyticsTracker.trackExperienceRecovery(experience: Experience) {
+    if (experience.renderErrorId == null) return
+
+    val errorId = experience.renderErrorId
+
+    track(
+        event = AnalyticsEvent.ExperienceError,
```
iujames, line 46: should be `AnalyticsEvent.ExperienceRecovery`

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -38,12 +39,12 @@ internal class ExperienceLifecycleTracker(
     private val storage: Storage by inject()
     private val appcuesCoroutineScope: AppcuesCoroutineScope by inject()
 
-    fun start(stateMachine: StateMachine, dispatcher: CoroutineDispatcher = Dispatchers.IO) {
-        appcuesCoroutineScope.launch(dispatcher) { stateMachine.observeState() }
+    fun start(stateMachine: StateMachine, onDismissed: (Experience) -> Unit, dispatcher: CoroutineDispatcher = Dispatchers.IO) {
```
iujames, line 42: maybe call this callback `onExperienceEnded` to more generally capture the state transition to `EndingExperience` - which could be dismissal or completion

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -65,6 +66,7 @@ internal class ExperienceLifecycleTracker(
                             // if ending on the last step OR an action requested it be considered complete explicitly,
                             // track the experience_completed event
                             trackLifecycleEvent(ExperienceCompleted(it.experience))
+                            onCompleted(it.experience)
```
iujames, line 69: interesting, yeah I'm not quite sure here. I think we need to clear it out of the list on completion or dismiss - any time it goes into EndingExperience state. I think this is what happens on iOS here https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift#L232

However, I think you noted a point about this - if I leave a screen and the view is torn down, that might trigger a dismiss - but if I return to that screen I might expect the embed to be back, since it wasn't really dismissed by the user? Maybe this is ok, as returning to the screen should trigger a new screen view and qualification in theory? Otherwise, we'd need a way to know about non-modal dismiss that differs from the modal case where we handle the view being destroyed as a dismiss (i.e. a deep link coming in that navigates to another Activity, for example)

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -101,25 +110,40 @@ internal class ExperienceRenderer(
...
     suspend fun show(experiences: List<Experience>) {
-        showEmbeds(experiences.filter { it.renderContext is RenderContext.Embed })
-        showModal(experiences.filter { it.renderContext is RenderContext.Modal })
+        // get any experiences if exist, and check if qualification trigger is screen_view
+        experiences.firstOrNull()?.let {
+            if (it.trigger is Qualification && it.trigger.reason == "screen_view") {
+                // clear list in case this was a screen_view qualification
+                potentiallyRenderableExperiences.clear()
+            }
+        }
```
iujames, line 119: following along - so just looking at the first experience here should be good enough, since all in a qualification batch would have the same trigger reason, i think that works.

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
andretortolano: thanks for bringing that up. not sure what to do here.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
andretortolano: would there be any other to check this?

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
iujames: I think the right thing to do at this point is to remove it from the list on any transition to `EndingExperience` - completed or dismissed.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
iujames: No, I think this works. Technically the `reason` is at the root of the response - and then an array of experiences are provided. But we do copy that `reason` to every experience I think during mapping - so this works.

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -242,7 +241,9 @@ public class Appcues internal constructor(koinScope: Scope) {
      * @param frame frame used to inflate the embedded experience
      */
     public fun registerEmbed(frameId: String, frame: AppcuesFrameView) {
-        renderContextManager.registerEmbedFrame(frameId, frame)
+        appcuesCoroutineScope.launch {
+            experienceRenderer.startFrame(frameId, frame)
```
andretortolano, line 245: instead of registering we "start", this means that we will also try to init an experience to this frame if exists

appcues/src/main/java/com/appcues/RenderContextManager.kt
```diff
@@ -1,25 +1,80 @@
...
 import com.appcues.data.model.RenderContext.Embed
+import com.appcues.statemachine.StateMachine
+import org.koin.core.component.KoinScopeComponent
+import org.koin.core.component.get
+import org.koin.core.component.inject
+import org.koin.core.scope.Scope
 import java.lang.ref.WeakReference
 
-internal class RenderContextManager {
+internal class RenderContextManager(override val scope: Scope) : KoinScopeComponent {
```
andretortolano, line 16: initially had RenderContextManager only for storing the AppcuesFrameView but I think its better to have it responsible for everything related to renderContext, instead of having half of the work here and the other half on ExperienceRenderer

appcues/src/main/java/com/appcues/RenderContextManager.kt
```diff
@@ -1,25 +1,80 @@
...
+                .also { stateMachineSlots[renderContext] = it }
+                .also { lifecycleTracker.start(it, { potentiallyRenderableExperiences.remove(renderContext) }) }
+        }
+    }
+
+    fun getStateMachine(renderContext: RenderContext): StateMachine? {
+        return stateMachineSlots[renderContext]
+    }
+
+    fun putExperiences(experiences: List<Experience>, trigger: ExperienceTrigger) {
```
andretortolano, line 54: just moving logic around, here we have the new experience list and the trigger, then we check if its screen_view in order to run some cleanup

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -38,12 +40,20 @@ internal class ExperienceLifecycleTracker(
     private val storage: Storage by inject()
     private val appcuesCoroutineScope: AppcuesCoroutineScope by inject()
 
-    fun start(stateMachine: StateMachine, dispatcher: CoroutineDispatcher = Dispatchers.IO) {
-        appcuesCoroutineScope.launch(dispatcher) { stateMachine.observeState() }
-        appcuesCoroutineScope.launch(dispatcher) { stateMachine.observeErrors() }
+    var stateJob: Job? = null
+    var errorJob: Job? = null
+
+    private fun cancelJobs() {
```
andretortolano, line 46: cancel both jobs if we hit the end of the experience or an error.  (maybe we should only cancel on EndExperience) ?

appcues/src/main/java/com/appcues/RenderContextManager.kt
```diff
@@ -1,25 +1,80 @@
...
 
-internal class RenderContextManager {
+internal class RenderContextManager(override val scope: Scope) : KoinScopeComponent {
 
-    private val slots: HashMap<RenderContext, WeakReference<AppcuesFrameView>> = hashMapOf()
+    private val lifecycleTracker by inject<ExperienceLifecycleTracker>()
+
+    private val frameSlots: HashMap<RenderContext, WeakReference<AppcuesFrameView>> = hashMapOf()
+
+    private val stateMachineSlots: HashMap<RenderContext, StateMachine> = hashMapOf()
```
iujames, line 22: just noting, on the iOS side, the `AppcuesFrameView` actually holds the reference to the `StateMachine`, keeping that one-to-one relationship defined there rather than having to have the two different maps here, ref: https://github.com/appcues/appcues-ios-sdk/blob/b074f690978ccb622bc7c88da8499b613a8002ac/Sources/AppcuesKit/Presentation/UI/AppcuesFrameView.swift#L19-L24

I don't know if that simplifies things here at all, or would cause issue, but wanted to at least note it if you thought it might make things easier to manage.

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
iujames: tricky one. I think there is the potential case where a fatal error occurs and it does _not_ go through EndExperience state - just right to Idle, so we'd want to catch that one, like you have.

There is also the other case where an error can occur but then be recovered - does this pose an issue now?

iujames (approved):
Looking good to me to merge to feature branch whenever convenient, as final details get worked out here or in a later PR.

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
andretortolano: upon testing I found out this is probably not right, an ExperienceEnding should not cancel those jobs too, as new experiences can still be processed.

appcues/src/main/java/com/appcues/RenderContextManager.kt
andretortolano: I may need to have the reference there for what I'm trying to do, so good to know!

PR #447, opened 2023-08-09
Add retry capability for target element matching

iujames:
There were a few complex bits to this, which I'll do my best to explain. The overall idea is this:
1. If a target-element is not found, it should allow some retry - by default after 300, 900, 1800, and 3000ms. If it still is not found after 3s, it will report a StepError and abort the flow (as before)
2. If the element is found, it will continue on and trigger the ExperienceStarted and StepSeen as appropriate
3. If the Activity is removed out from under us (deep link or user navigation) - it will report a new type of StepError `"Activity changed during render of step $index" - and abort the flow (free up the state machine)

To support this, the `@appcues/target-element` trait has an optional array of Ints (milliseconds) for `retryIntervals` that define the retry timeline (defaults as noted above). `AppcuesComposition` is where the real retry logic lives. In the `produceMetadata` function, it handles the `TraitException` and retry loop as necessary. A `TraitException` may optionally define the `retryMilliseconds` now, and this is used in the `TargetElementTrait` for the error types that _may_ be recoverable.

One detail here - when we go into the delay+retry, we must hide the overlay view layer so that the Application is not unresponsive to the user during the delay with that transparent view hanging on top. This is done through a callback, through the view model, and into the AppcuesOverlayViewManager. I need to check if this will have implications on changes being made for Embeds, which change some of how these view managers / presenters work.

The changes in `AppcuesViewModel` are largely around supporting the edge case #3 - making sure that we leave our `StateMachine` in a valid state, regardless of how the the Experience has been presented successfully, or failed.

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/AppcuesTraitException.kt
+++ b/appcues/src/main/java/com/appcues/trait/AppcuesTraitException.kt
@@ -3,4 +3,13 @@ package com.appcues.trait
 /**
  * A type of exception that can occur during the application of an ExperienceTrait.
  */
-public class AppcuesTraitException(message: String) : Exception(message)
+public class AppcuesTraitException(
+    message: String,
+
+    /**
+     *  When set, this value can indicate a number of milliseconds after which re-application
+     *  of traits can be attempted, to try to auto-recover from this error.
+     */
+    public val retryMilliseconds: Int? = null,
+
+) : Exception(message)
--- a/appcues/src/main/java/com/appcues/trait/appcues/TargetElementTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/appcues/TargetElementTrait.kt
@@ -27,6 +27,12 @@ internal class TargetElementTrait(
     private val contentDistance = config.getConfigOrDefault("contentDistanceFromTarget", 0.0)
     private val preferredPosition = config.getConfig<String>("contentPreferredPosition").toPosition()
 
+    @Suppress("MagicNumber")
+    private val retryIntervals = (config.getConfig<List<Int>>("retryIntervals") ?: listOf(300, 600, 900, 1_200)).toMutableList()
+
+    private val retryMilliseconds: Int?
+        get() = if (retryIntervals.isNotEmpty()) retryIntervals.removeFirst() else null
+
     override fun produceMetadata(): Map<String, Any?> {
         val view = viewMatchingSelector()
 
@@ -63,10 +69,16 @@ internal class TargetElementTrait(
 
         val weightedViews = strategy.findMatches(selector)
             // if the result is null (not just empty) - that means the UI layout was not available at all
-            ?: throw AppcuesTraitException("could not read application layout information")
+            ?: throw AppcuesTraitException(
+                message = "could not read application layout information",
+                retryMilliseconds = retryMilliseconds,
+            )
 
         if (weightedViews.isEmpty()) {
-            throw AppcuesTraitException("no view matching selector ${selector.toMap()}")
+            throw AppcuesTraitException(
+                message = "no view matching selector ${selector.toMap()}",
+                retryMilliseconds = retryMilliseconds,
+            )
         }
 
         // views contains an array of Pairs, each item being the view and its integer match value
@@ -99,6 +111,9 @@ internal class TargetElementTrait(
             return maxWeightViews[0]
         }
 
-        throw AppcuesTraitException("multiple non-distinct views (${weightedViews.count()}) matched selector ${selector.toMap()}")
+        throw AppcuesTraitException(
+            message = "multiple non-distinct views (${weightedViews.count()}) matched selector ${selector.toMap()}",
+            retryMilliseconds = retryMilliseconds,
+        )
     }
 }
--- a/appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
+++ b/appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
@@ -60,8 +60,12 @@ internal class AppcuesOverlayViewManager(
         AppcuesActivityMonitor.activity?.removeOverlayView()
     }
 
+    private fun setOverlayVisible(isVisible: Boolean) {
+        AppcuesActivityMonitor.activity?.updateOverlayVisibility(isVisible)
+    }
+
     private fun Activity.addOverlayView(): Boolean {
-        viewModel = AppcuesViewModel(scope, renderContext, ::onCompositionDismiss)
+        viewModel = AppcuesViewModel(scope, renderContext, ::onCompositionDismiss, ::setOverlayVisible)
         gestureListener = ShakeGestureListener(this)
 
         val parentView = getParentView()
@@ -116,6 +120,24 @@ internal class AppcuesOverlayViewManager(
                 viewModel.onFinish()
 
                 it.setAccessibility(View.IMPORTANT_FOR_ACCESSIBILITY_YES)
+            }
+        }
+    }
+
+    private fun Activity.updateOverlayVisibility(isVisible: Boolean) {
+        getParentView().let { parentView ->
+            parentView.post {
+                findViewById<ComposeView>(R.id.appcues_overlay_view)?.let {
+                    it.visibility = if (isVisible) ViewGroup.VISIBLE else ViewGroup.GONE
+                }
+                parentView.setAccessibility(
+                    if (isVisible) {
+                        // when our view is visible, the parent view is removed from the accessibility stack
+                        View.IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS
+                    } else {
+                        View.IMPORTANT_FOR_ACCESSIBILITY_YES
+                    }
+                )
             }
         }
     }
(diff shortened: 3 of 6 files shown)


appcues/src/main/java/com/appcues/trait/AppcuesTraitException.kt
```diff
@@ -3,4 +3,13 @@ package com.appcues.trait
...
  */
-public class AppcuesTraitException(message: String) : Exception(message)
+public class AppcuesTraitException(
+    message: String,
+
+    /**
+     *  When set, this value can indicate a number of milliseconds after which re-application
+     *  of traits can be attempted, to try to auto-recover from this error.
+     */
+    public val retryMilliseconds: Int? = null,
```
iujames, line 13: fyi, there is some discussion on the iOS side over whether we want to expose this value publicly https://github.com/appcues/appcues-ios-sdk/pull/434#discussion_r1287648945

appcues/src/main/java/com/appcues/trait/appcues/TargetElementTrait.kt
```diff
@@ -63,10 +67,16 @@ internal class TargetElementTrait(
...
+                message = "could not read application layout information",
+                retryMilliseconds = retryMilliseconds,
+            )
 
         if (weightedViews.isEmpty()) {
-            throw AppcuesTraitException("no view matching selector ${selector.toMap()}")
+            throw AppcuesTraitException(
+                message = "no view matching selector ${selector.toMap()}",
+                retryMilliseconds = retryMilliseconds,
+            )
```
iujames, line 79: this is the typical selector not found - but the other two items in this class are also considered possibly recoverable as a view loads or animates in.

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
```diff
@@ -60,8 +60,12 @@ internal class AppcuesOverlayViewManager(
         AppcuesActivityMonitor.activity?.removeOverlayView()
     }
 
+    private fun setOverlayVisible(isVisible: Boolean) {
+        AppcuesActivityMonitor.activity?.updateOverlayVisibility(isVisible)
+    }
+
     private fun Activity.addOverlayView(): Boolean {
-        viewModel = AppcuesViewModel(scope, renderContext, ::onCompositionDismiss)
+        viewModel = AppcuesViewModel(scope, renderContext, ::onCompositionDismiss, ::setOverlayVisible)
```
iujames, line 68: this part, and related helper function, may need to evolve to compile correctly in the new embed framework - though embeds will not necessarily have any sort of retry needed, not actually use the hide/show.

appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
```diff
@@ -43,8 +47,32 @@ internal class AppcuesViewModel(
...
-            val presentationComplete: ((ResultOf<Unit, com.appcues.statemachine.Error>) -> Unit),
-        ) : UIState()
+            private val completionHandler: ((ResultOf<Unit, com.appcues.statemachine.Error>) -> Unit),
+        ) : UIState() {
+            // used to track presentation state, to avoid calling the completion multiple times
+            private var presented: Boolean = false
+            private val mutex = Mutex()
+
+            // returns true if the completion handler was invoked, false if not (already presented)
+            suspend fun presentationComplete(result: ResultOf<Unit, com.appcues.statemachine.Error>): Boolean = mutex.withLock {
```
iujames, line 57: note: we now wrap the usage of the completion handler, and use a `Mutex` for thread safety - as we need to ensure this callback only goes back to the state machine once. There is an edge case in the `onActivityChanged` usage where it _could_ be called twice - deep linking away once an experience has already rendered.

appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
```diff
@@ -84,13 +112,40 @@ internal class AppcuesViewModel(
...
 
+    fun onPresentationComplete() {
+        uiState.value.let { state ->
+            if (state is Rendering) {
+                appcuesCoroutineScope.launch {
+                    state.presentationComplete(Success(Unit))
+                }
+            }
+        }
+    }
```
iujames, line 123: this is called from the `AppcuesComposition` now, to make usage in that spot simpler and encapsulate the usage of the `appcuesCoroutineScope` here.

appcues/src/main/java/com/appcues/ui/AppcuesViewModel.kt
```diff
@@ -84,13 +112,40 @@ internal class AppcuesViewModel(
...
+                    // will have already been triggered during first render, and this will do nothing.
+                    val handled = state.presentationComplete(
+                        Failure(
+                            StepError(
+                                experience = state.experience,
+                                stepIndex = state.flatStepIndex,
+                                message = "Activity changed during render of step ${state.flatStepIndex}",
+                            )
+                        )
+                    )
```
iujames, line 142: this is where we handle the new edge case - the Activity was removed during the error handling / retry phase.

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
```diff
@@ -112,16 +109,50 @@ private fun BoxScope.ComposeLastRenderingState(state: Rendering) {
...
+        val previousStepMetaData = remember { mutableStateOf(AppcuesStepMetadata()) }
+
+        LaunchedEffect(metadataSettingTraits) {
+            val actual = produceMetadata(metadataSettingTraits.value, traitExceptionHandler, overlayVisibility)
+            if (actual != null) {
+                stepMetadata.value = AppcuesStepMetadata(previous = previousStepMetaData.value.current, current = actual).also {
+                    previousStepMetaData.value = it
+                }
+            }
+        }
```
iujames, line 152: please review this section carefully around he usage of the `LaunchedEffect` and how it is replacing what was previously in `rememberMetadataSettingsTraits` with the new call to the suspend fun for `produceMetadata(...)`. This allows for the async retry within `produceMetadata` and then the rest of the composition relies on `stepMetadata.value` (L154 below) to continue composing once valid metadata is produced (or fails through exception handler if not).

Mostly, I want to make sure I didn't miss any Compose gotchas around how that previous remember was working, which could cause any unexpected side effects or recomposition issues.

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
```diff
@@ -50,6 +49,7 @@ internal fun AppcuesComposition(
             LocalAppcuesActionDelegate provides DefaultAppcuesActionsDelegate(viewModel),
             LocalAppcuesPaginationDelegate provides AppcuesPagination { viewModel.onPageChanged(it) },
             LocalAppcuesTraitExceptionHandler provides AppcuesTraitExceptionHandler { viewModel.onTraitException(it) },
+            LocalAppcuesOverlayVisibility provides AppcuesOverlayVisibility { viewModel.updateOverlayVisibility(it) },
```
iujames, line 52: we have several things now using the `viewModel` in a sort of abstracted way, for bits and pieces of functionality. We also have the `LocalViewModel` with the entire object available. It seems a bit redundant now to have all these other composition locals that wrap functions of the view model?

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
andretortolano: ok my understanding is that whenever metadataSettingTraits changes (so on a new stepIndex) then try to produce metadata and set as the value, this looks ok change

appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
```diff
@@ -72,10 +72,7 @@ private fun MainSurface() {
             }
 
             LaunchOnShowAnimationCompleted {
-                val currentState = state.value
-                if (currentState is Rendering) {
-                    currentState.presentationComplete.invoke(Success(Unit))
-                }
+                viewModel.onPresentationComplete()
```
andretortolano, line 75: Oh okay nice simplification

appcues/src/main/java/com/appcues/ui/AppcuesOverlayViewManager.kt
```diff
@@ -120,6 +124,24 @@ internal class AppcuesOverlayViewManager(
         }
     }
 
+    private fun Activity.updateOverlayVisibility(isVisible: Boolean) {
+        getParentView().let { parentView ->
+            parentView.post {
+                findViewById<ComposeView>(R.id.appcues_overlay_view)?.let {
+                    it.visibility = if (isVisible) ViewGroup.VISIBLE else ViewGroup.GONE
```
andretortolano, line 131: there is a method View.isVisible [true/false]

iujames: ah yeah I wasn't sure if that set it GONE when false, but looking at the docs, it does - so that should work fine to simplify.

iujames: updated in https://github.com/appcues/appcues-android-sdk/pull/447/commits/885c29baa8fa9d996c65cbc578a33643d4a7ac95

andretortolano (approved):
this looks ok to merge

iujames:
@andretortolano due to the changes in `AppcuesOverlayViewManager`, I think I should rework this on top of the embeds branch.  But for now at least, going to rebase embeds off main prior to this change.

PR #452, opened 2023-08-15
Update session tracking

iujames:
At a high level: remove session events for suspend/resume/reset - only track `session_start`. Update the timeout default to 300 sec (was 1800), and check for timeout after any period of analytics inactivity, regardless of app background/foreground. No longer track background/foreground at all effectively.

`AnalyticsTracker` now completely owns the session management. On any attempt to track any analytics, it will (A) check if any session exists and (B) check if that session is expired. In either case, it will try to restart a new session. This can fail if there is no user identified. Session starts are thus completely based on some initial analytic being tracked (commonly the user identify at start), and the session duration depends on continued tracking with no period of inactivity longer than the timeout. There are no other session start calls made from outside of `AnalyticsTracker` (removed the Appcues.kt usage).

This ended up having some more layers than expected. The session event data was tied into the Debugger UI display (using reset events to refresh some things), and the unit tests had many spots that needed to be revised to fit the new / simpler paradigm.

### DIFF
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -96,8 +96,6 @@ public class Appcues internal constructor(koinScope: Scope) {
     public var navigationHandler: NavigationHandler? by config::navigationHandler
 
     init {
-        sessionMonitor.start()
-
         logcues.info("Appcues SDK $version initialized")
 
         appcuesCoroutineScope.launch {
@@ -167,11 +165,16 @@ public class Appcues internal constructor(koinScope: Scope) {
      * Can be used when the user logs out of your application.
      */
     public fun reset() {
+        // flush any pending analytics for the previous user, prior to reset
+        analyticsTracker.flushPendingActivity()
+
         sessionMonitor.reset()
         storage.userId = ""
         storage.userSignature = null
         storage.isAnonymous = true
         storage.groupId = null
+
+        debuggerManager.reset()
     }
 
     /**
@@ -287,8 +290,9 @@ public class Appcues internal constructor(koinScope: Scope) {
         storage.isAnonymous = isAnonymous
         storage.userSignature = mutableProperties?.remove("appcues:user_id_signature") as? String
         if (userChanged) {
-            // when the identified user changes from last known value, we must start a new session
-            sessionMonitor.start()
+            // when the user changes, reset the session monitor and a new session will
+            // be created the first time analytics are tracked for the new user (the identify)
+            sessionMonitor.reset()
 
             // group info is reset on new user
             storage.groupId = null
--- a/appcues/src/main/java/com/appcues/SessionMonitor.kt
+++ b/appcues/src/main/java/com/appcues/SessionMonitor.kt
@@ -3,11 +3,8 @@ package com.appcues
 import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.ProcessLifecycleOwner
-import com.appcues.analytics.AnalyticsEvent
 import com.appcues.analytics.AnalyticsTracker
 import com.appcues.analytics.SdkMetrics
-import com.appcues.analytics.track
-import com.appcues.logging.Logcues
 import kotlinx.coroutines.launch
 import org.koin.core.component.KoinScopeComponent
 import org.koin.core.component.inject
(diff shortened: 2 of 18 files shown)


appcues/src/main/java/com/appcues/debugger/AppcuesDebuggerManager.kt
```diff
@@ -26,6 +26,7 @@ import kotlinx.coroutines.launch
 import org.koin.core.scope.Scope
 import kotlin.coroutines.CoroutineContext
 
+@Suppress("TooManyFunctions")
```
iujames, line 29: I had to have a way to pass the `reset()` through here, now that no event is triggered. That tipped this class over the 15 function limit. Since its a debugger class, I feel not critical, but don't love suppressing, and don't have a better idea at the moment.

appcues/src/main/java/com/appcues/debugger/DebuggerStatusManager.kt
```diff
@@ -33,6 +33,7 @@ import kotlinx.coroutines.withContext
 import java.util.UUID
 import kotlin.time.Duration.Companion.seconds
 
+@Suppress("TooManyFunctions")
```
iujames, line 36: Similar to the other - the need to pass `reset()` through here added a 15th function.

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -167,11 +165,16 @@ public class Appcues internal constructor(koinScope: Scope) {
      * Can be used when the user logs out of your application.
      */
     public fun reset() {
+        // flush any pending analytics for the previous user, prior to reset
+        analyticsTracker.flushPendingActivity()
```
iujames, line 169: moved this responsibility here, from internal in the `sessionMonitor`, as I'm trying to keep the responsibilities separate as much as possible and not rely on the session reset to side effect and flush analytics.

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -96,8 +96,6 @@ public class Appcues internal constructor(koinScope: Scope) {
     public var navigationHandler: NavigationHandler? by config::navigationHandler
 
     init {
-        sessionMonitor.start()
```
iujames, line 99: not necessary, as the first tracked analytic will start the session as needed.

appcues/src/test/java/com/appcues/SessionMonitorTest.kt
```diff
@@ -28,216 +18,40 @@ internal class SessionMonitorTest : AppcuesScopeTest {
...
         sessionMonitor.reset()
 
         // THEN
-        verify(exactly = 1) { tracker.track(SessionReset, any(), true) }
         assertThat(sessionMonitor.sessionId).isNull()
-        assertThat(sessionMonitor.isActive).isFalse()
-    }
-
-    @Test
-    fun `onStart SHOULD not track an event WHEN onStop was not previously called`() {
```
iujames, line 77: the tests here and below just are not valid any longer - they are testing functionality in the more complicated session tracking that does not exist now.

appcues/src/main/java/com/appcues/analytics/AnalyticsTracker.kt
```diff
@@ -75,6 +75,31 @@ internal class AnalyticsTracker(
...
+                // anything - return false to caller
+                return false
+            }
+        } else {
+            // we have a valid session, update its last activity timestamp to push out the timeout
+            sessionMonitor.updateLastActivity()
+        }
+
+        return true
+    }
```
iujames, line 101: this is the key new spot where everything session related is managed - and called by each tracking usage above in this class. This will either start a new session when needed, or bump out the timestamp for the existing one to delay timeout.

appcues/src/main/java/com/appcues/data/remote/appcues/AppcuesRemoteSource.kt
```diff
@@ -43,8 +41,8 @@ internal class AppcuesRemoteSource(
         experienceId: String,
         userSignature: String?,
     ): ResultOf<ExperienceResponse, RemoteError> =
-        // preview _can_ be personalized, so attempt to use the user info, if a valid session exists
-        if (sessionMonitor.isActive) {
+        // preview _can_ be personalized, so attempt to use the user info, if a valid userId exists
+        if (storage.userId.isNotEmpty()) {
```
iujames, line 45: this is just a simplification and dependency removal - we can do a personalized preview as long as we have the userId

appcues/src/main/java/com/appcues/SessionMonitor.kt
```diff
@@ -43,46 +41,23 @@ internal class SessionMonitor(
...
-        // this is interactive: true since a reset should flush to network immediately (with previous user ID)
-        // and the next session start will be sent in a new request, with the new user ID
-        analyticsTracker.track(AnalyticsEvent.SessionReset, null, true)
-        _sessionId = null
-    }
-
-    fun checkSession(errorMessage: String): Boolean {
-        if (isActive) return true
-
-        logcues.info("No active session - $errorMessage")
```
iujames, line 62: logging these "no active session" messages seemed minimal value IMO. I don't feel strongly though and can re-introduce some logging if desired.

appcues/src/main/java/com/appcues/debugger/AppcuesDebuggerManager.kt
andretortolano: not sure if Applies, there is a DefaultLifetracker that we dont necessarily need to override all methods, maybe could be used here to reduce unused lifecycle methods?

andretortolano (approved):

appcues/src/main/java/com/appcues/debugger/AppcuesDebuggerManager.kt
iujames: It looks like that is for listening to the lifecycle events of a single owner (i.e. Activity) through the Activity's `lifecycle`
```kotlin
lifecycle.addObserver(x)
```

But in this case, we are actually trying to listen to all activities in the application through 
```kotlin
application.registerActivityLifecycleCallbacks(this)
```

So I don't think it can be used in this case.

PR #458, opened 2023-08-18
📝 Migration Docs update

andretortolano:


iujames (commented):
seems ok to me having in a single doc - wondering if we can make consistent with iOS though (currently different docs per version). @mmaatttt do you have a preference?

mmaatttt:
@iujames I don't have a strong preference. I can update iOS to be a single doc if that makes the most sense.

Personally, I've found it convenient having them separate in other tools. Makes it just a little bit easier to find. For example, here's [Capacitor](https://capacitorjs.com/docs). That said, there's really not much that matters to most of our customers in those docs, whereas the Capacitor ones are likely much more high traffic.

Screenshot 2023-08-21 at 9 17 14 AM

andretortolano:
keeping it in a single file keeps a nice history + we dont need to keep updating other references, right?

mmaatttt (approved):

iujames (approved):

docs/VersionMigration.md
```diff
@@ -9,3 +9,34 @@ Updating to this release will not require any code changes for most SDK installa
 ### Removed
 
 - To simplify the public API, `ExperienceAction` and `ExperienceTrait` related classes and functions no longer have public visibility. There are no current use cases for extensibility that require these to be public. This change impacts items in the `com.appcues.action` and `com.appcues.trait` packages.
+
+
+# 1.x to 2.0 Migration Guide
```
iujames, line 14: fwiw, on the iOS side we had a single `#` header and then each version within was a `##` sub-header, kind of liked that.

https://github.com/appcues/appcues-ios-sdk/pull/442/files

mmaatttt: Only did that because the DocC compiler complained, but I agree that it's usually best practice to have a single <h1> in a doc.

PR #459, opened 2023-08-21
✨ Preview for embeds

andretortolano:
Preview logic for embeds.

storing embeds preview in a different dictionary that is no cleared on new screen event calls. also ensuring it is always using a prestine state machine or else we could get into some scenarios where the state machine doesnt have a view and its not dismissing the existing experience before showing the new previewd one.

### DIFF
--- a/appcues/src/main/java/com/appcues/DeepLinkHandler.kt
+++ b/appcues/src/main/java/com/appcues/DeepLinkHandler.kt
@@ -3,11 +3,18 @@ package com.appcues
 import android.app.Activity
 import android.content.Intent
 import android.net.Uri
+import android.util.Log
+import android.widget.Toast
 import com.appcues.data.model.ExperienceTrigger.DeepLink
 import com.appcues.debugger.AppcuesDebuggerManager
 import com.appcues.debugger.DebugMode.Debugger
 import com.appcues.debugger.DebugMode.ScreenCapture
 import com.appcues.ui.ExperienceRenderer
+import com.appcues.ui.ExperienceRenderer.PreviewResponse.ExperienceNotFound
+import com.appcues.ui.ExperienceRenderer.PreviewResponse.Failed
+import com.appcues.ui.ExperienceRenderer.PreviewResponse.PreviewDeferred
+import com.appcues.ui.ExperienceRenderer.PreviewResponse.StateMachineError
+import com.appcues.ui.ExperienceRenderer.PreviewResponse.Success
 import kotlinx.coroutines.launch
 
 internal class DeepLinkHandler(
@@ -39,7 +46,7 @@ internal class DeepLinkHandler(
         return when {
             segments.count() == 2 && segments[0] == "experience_preview" -> {
                 appcuesCoroutineScope.launch {
-                    experienceRenderer.preview(segments[1])
+                    previewExperience(segments[1], activity)
                 }
                 true
             }
@@ -64,6 +71,22 @@ internal class DeepLinkHandler(
                 }
             }
             else -> false
+        }
+    }
+
+    private suspend fun previewExperience(experienceId: String, activity: Activity) {
+        experienceRenderer.preview(experienceId).run {
+            val resources = activity.resources
+
+            Log.i("Logcues", "Preview response $this")
+            when (this) {
+                is Failed -> resources.getString(R.string.appcues_preview_flow_failed)
+                is PreviewDeferred -> resources.getString(R.string.appcues_preview_embed_message, frameId, experience.name)
+                is StateMachineError -> resources.getString(R.string.appcues_preview_flow_failed_reason, experience.name, error.message)
+                is ExperienceNotFound -> resources.getString(R.string.appcues_preview_flow_not_found)
+                // do nothing. previewing experience
+                is Success -> null
+            }?.let { errorMessage -> Toast.makeText(activity, errorMessage, Toast.LENGTH_LONG).show() }
         }
     }
 }
--- a/appcues/src/main/java/com/appcues/data/AppcuesRepository.kt
+++ b/appcues/src/main/java/com/appcues/data/AppcuesRepository.kt
@@ -10,10 +10,12 @@ import com.appcues.data.model.ExperiencePriority
 import com.appcues.data.model.ExperiencePriority.LOW
 import com.appcues.data.model.ExperiencePriority.NORMAL
 import com.appcues.data.model.ExperienceTrigger
+import com.appcues.data.remote.RemoteError
 import com.appcues.data.remote.RemoteError.NetworkError
 import com.appcues.data.remote.appcues.AppcuesRemoteSource
 import com.appcues.data.remote.appcues.request.ActivityRequest
 import com.appcues.logging.Logcues
+import com.appcues.util.ResultOf
 import com.appcues.util.ResultOf.Failure
 import com.appcues.util.ResultOf.Success
 import com.appcues.util.doIfFailure
@@ -35,6 +37,7 @@ internal class AppcuesRepository(
     private val logcues: Logcues,
     private val storage: Storage,
 ) {
+
     private val processingActivity: HashSet<UUID> = hashSetOf()
     private val mutex = Mutex()
 
@@ -50,14 +53,11 @@ internal class AppcuesRepository(
         }
     }
 
-    suspend fun getExperiencePreview(experienceId: String): Experience? = withContext(Dispatchers.IO) {
-        appcuesRemoteSource.getExperiencePreview(experienceId, storage.userSignature).let {
+    suspend fun getExperiencePreview(experienceId: String): ResultOf<Experience, RemoteError> = withContext(Dispatchers.IO) {
+        return@withContext appcuesRemoteSource.getExperiencePreview(experienceId, storage.userSignature).let {
             when (it) {
-                is Success -> experienceMapper.map(it.value, ExperienceTrigger.Preview)
-                is Failure -> {
-                    logcues.info("Experience preview request failed, reason: ${it.reason}")
-                    null
-                }
+                is Success -> Success(experienceMapper.map(it.value, ExperienceTrigger.Preview))
+                is Failure -> it.also { logcues.info("Experience preview request failed, reason: ${it.reason}") }
             }
         }
     }
--- a/appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
+++ b/appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
@@ -14,6 +14,8 @@ import com.appcues.data.model.ExperienceTrigger
 import com.appcues.data.model.ExperienceTrigger.Qualification
 import com.appcues.data.model.RenderContext
 import com.appcues.data.model.RenderContext.Modal
+import com.appcues.data.model.getFrameId
+import com.appcues.data.remote.RemoteError.HttpError
 import com.appcues.statemachine.Action.EndExperience
 import com.appcues.statemachine.Action.StartExperience
 import com.appcues.statemachine.Action.StartStep
@@ -23,6 +25,9 @@ import com.appcues.statemachine.State
 import com.appcues.statemachine.State.Idling
 import com.appcues.statemachine.StateMachine
 import com.appcues.statemachine.StepReference
+import com.appcues.ui.ExperienceRenderer.RenderingResult.NoRenderContext
+import com.appcues.ui.ExperienceRenderer.RenderingResult.StateMachineError
+import com.appcues.ui.ExperienceRenderer.RenderingResult.WontDisplay
 import com.appcues.util.ResultOf
 import com.appcues.util.ResultOf.Failure
 import com.appcues.util.ResultOf.Success
@@ -36,8 +41,14 @@ internal class ExperienceRenderer(
     override val scope: Scope,
 ) : KoinScopeComponent, StateMachineOwning {
 
+    companion object {
+
+        private const val HTTP_CODE_NOT_FOUND = 404
+    }
+
     // Conformance to `StateMachineOwning`.
     override var renderContext: RenderContext? = null
+
     // State machine for `RenderContext.modal`
     override var stateMachine: StateMachine? = null
 
@@ -51,6 +62,7 @@ internal class ExperienceRenderer(
     private val lifecycleTracker by inject<ExperienceLifecycleTracker>()
 
     private val potentiallyRenderableExperiences = hashMapOf<RenderContext, List<Experience>>()
+    private val previewExperiences = hashMapOf<RenderContext, Experience>()
 
     init {
         // sets up the single state machine used for modal experiences (mobile flows, not embeds)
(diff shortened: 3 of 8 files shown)


appcues/src/main/java/com/appcues/DeepLinkHandler.kt
```diff
@@ -39,7 +41,16 @@ internal class DeepLinkHandler(
         return when {
             segments.count() == 2 && segments[0] == "experience_preview" -> {
                 appcuesCoroutineScope.launch {
-                    experienceRenderer.preview(segments[1])
+                    experienceRenderer.preview(segments[1])?.let {
+                        val frameId = if (it.experience.renderContext is RenderContext.Embed)
+                            it.experience.renderContext.frameId else String()
+
+                        Toast.makeText(
```
andretortolano, line 48: should this be a snackbar? we cant re-use the same ones we have in debugger.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -140,12 +140,18 @@ internal class ExperienceRenderer(
...
+    data class PreviewError(val experience: Experience)
+
+    suspend fun preview(experienceId: String): PreviewError? {
         repository.getExperiencePreview(experienceId)?.let {
-            return show(it)
+            renderContextManager.putPreview(it)
+
+            if (show(it)) return null
+
+            return PreviewError(it)
```
iujames, line 151: If this is a preview of a tooltip that is not found, that would also give an error on show, correct? Could we have a situation where the "Please navigate to screen with frame..." error is invalid/confusing? Wondering if we need more targeted error handling specific to embeds and frame not found cases?

appcues/src/main/java/com/appcues/RenderContextManager.kt
```diff
@@ -70,10 +72,25 @@ internal class RenderContextManager(override val scope: Scope) : KoinScopeCompon
...
         stateMachineSlots.values.forEach { it.stop() }
     }
+
+    fun putPreview(experience: Experience) {
+        val renderContext = experience.renderContext
+
+        // check if context is embed
+        if (renderContext is Embed) {
+            // if frame of that embed is null, also remove its state machine from slot
+            stateMachineSlots.remove(renderContext)
```
iujames, line 91: can you say more about this part? the comment makes it sound like it's checking a condition on the frame but it is not actually?

iujames (commented):
I'm thinking we need some way to reset the preview right? Would that only happen if I re-preview the same thing? I'm wondering if a preview should remove from cache after it is rendered once - so you aren't stuck forever (in that app session) with a preview you cannot remove?

appcues/src/main/java/com/appcues/RenderContextManager.kt
andretortolano: so instead of "show" on a existing state machine with an existing experience, we need to remove the state machine and create a new every time, or else the "show" logic will try to dismiss the existing experience, and this might be a problem because the frame might not be visible, thus living the coroutine in a wait state forever.

andretortolano: to reproduce this, lets say we preview frame1, the first time everything is going to work OK, but then if you navigate back to events fragment and trigger the deep link again, then nothing happens. this is because on the second time we would be trying to dismiss an experience that is already being rendered on that state machine, but dismiss is impossible because since the view is not visible, the viewModel does not reply back.

so this is one of the solutions, the other would be to enhance how we dismiss experience to make sure this doesnt happen i guess.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
mmaatttt: FYI for iOS, I've added the embed error, but also handling for network errors and a fallback to a generic message: https://github.com/appcues/appcues-ios-sdk/blob/b96359ca9c02e6f590bbf2cc244a9c0ccd448f9d/Sources/AppcuesKit/Presentation/DeepLinkHandler.swift#L134-L153

appcues/src/main/java/com/appcues/RenderContextManager.kt
andretortolano: the commend line doesnt make sense actually, it was part of a different logic I was trying to put in here, it should say something like

"every time we put a new preview for embed, silently remove existing state machine.

iujames: aha ok i think i follow now thank you. Ran into similar things while reviewing the other code around error handling. Makes sense to consider a preview similar to a re-register of a frame I think, where the new preview will trigger the creation of a new state machine for that context.

I think this can adapt to work on top of the other changes proposed in https://github.com/appcues/appcues-android-sdk/pull/460 fairly easily, but will require some merge handling depending on what goes first.

appcues/src/main/java/com/appcues/DeepLinkHandler.kt
iujames: I think since Android has the native Toast it makes sense to just use it for this utility case, as oppose to iOS where the debugger item was ported over. Porting the debugger item over would prob be fine too, but just extra work. In terms of the end user UX toast/snackbar, maybe best to run that by Sam.

appcues/src/main/res/values/strings.xml
```diff
@@ -2,6 +2,9 @@
 <resources>
     <!-- SDK -->
     <string name="appcues_skippable_trait_dismiss">dismiss</string>
+    <string name="appcues_preview_embed_message">Please navigate to the screen with frame %1$s to preview %2$s.</string>
+    <string name="appcues_preview_flow_not_found">Flow not found.</string>
+    <string name="appcues_preview_preview_failed">Mobile flow preview failed.</string>
```
andretortolano, line 7: using only one "generic" error message instead of two, or should I mirror IOS, maybe I didnt get the difference [here](https://github.com/appcues/appcues-ios-sdk/blob/b96359ca9c02e6f590bbf2cc244a9c0ccd448f9d/Sources/AppcuesKit/Presentation/DeepLinkHandler.swift#L141).

iujames: "Flow not found" would be the 404 case - invalid id
"Mobile flow preview failed" - the network loading was OK, but this case covers anything that causes failed render like invalid element target

the missing one is
"Error loading mobile flow preview." - this would be a network failure for any reason other than 404 - like invalid JSON structure or some other connection issue

appcues/src/main/java/com/appcues/DeepLinkHandler.kt
```diff
@@ -17,6 +23,11 @@ internal class DeepLinkHandler(
     private val debuggerManager: AppcuesDebuggerManager,
 ) {
 
+    companion object {
+
+        private const val NETWORK_ERROR_NOT_FOUND = 400
```
iujames, line 28: 404

appcues/src/main/java/com/appcues/data/AppcuesRepository.kt
```diff
@@ -50,13 +53,16 @@ internal class AppcuesRepository(
         }
     }
 
-    suspend fun getExperiencePreview(experienceId: String): Experience? = withContext(Dispatchers.IO) {
-        appcuesRemoteSource.getExperiencePreview(experienceId, storage.userSignature).let {
+    suspend fun getExperiencePreview(experienceId: String): PreviewResponse = withContext(Dispatchers.IO) {
```
iujames, line 56: Maybe this could return a `ResultOf<Experience, RemoteError>`

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -197,12 +201,26 @@ internal class ExperienceRenderer(
...
+    sealed class PreviewResponse {
+        data class Success(val experience: Experience) : PreviewResponse()
+        data class NetworkError(val throwable: Throwable?) : PreviewResponse()
+        data class HttpError(val code: Int?) : PreviewResponse()
+        data class RenderDeferred(val frameId: String, val name: String) : PreviewResponse()
+    }
 
-        return false
+    suspend fun preview(experienceId: String): PreviewResponse {
+        repository.getExperiencePreview(experienceId).run {
```
iujames, line 212: Then, you could do the mapping to `PreviewResponse` here, since it makes sense that the renderer will need to return that detail back to the caller for the error messages

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -103,7 +106,8 @@ internal class ExperienceRenderer(
         experiment?.let { analyticsTracker.track(it) }
 
         when (stateMachine.handleAction(StartExperience(this))) {
-            is Success -> true
+            // removes possible preview experience from hashMap if it was showed properly
+            is Success -> true.also { previewExperiences.remove(renderContext) }
```
iujames, line 110: it looks like on the iOS side that the preview gets [removed on completion](https://github.com/appcues/appcues-ios-sdk/pull/441/files#diff-e9daefd60bacc377827cb6a2002bcd0fe84cd5b3ec407cd9ab2a6a5dd676a37fR262). That would correspond to L177 the completion handler of the lifecycle tracker in this class, I think.

granted, completion may not happen for embeds very often, right? @mmaatttt makes me wonder if removal here on successful start is better?

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -197,12 +201,26 @@ internal class ExperienceRenderer(
...
+    }
 
-        return false
+    suspend fun preview(experienceId: String): PreviewResponse {
+        repository.getExperiencePreview(experienceId).run {
+            if (this is PreviewResponse.Success) {
+                val renderContext = experience.renderContext
+                previewExperiences[renderContext] = experience
+
+                if (!show(experience) && renderContext is RenderContext.Embed) {
```
iujames, line 217: I think ideally `show()` is updated to return a result with success/failure, and only return the `RenderDeferred` error in the specific case that occurs L94. Then, any other failure would be properly handled. There could be some other reason an embed failed to render, in theory (maybe not today), so `show()` would need to pass that information back.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
andretortolano: I guess my idea is that whenever it StartExperience with success we can basically clean up the preview dictionary right, that would work ok for all

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -197,12 +210,32 @@ internal class ExperienceRenderer(
...
-        repository.getExperiencePreview(experienceId)?.let {
-            return show(it)
-        }
+    sealed class PreviewResponse {
+        object Success : PreviewResponse()
+        data class NetworkError(val throwable: Throwable?) : PreviewResponse()
+        data class HttpError(val code: Int?) : PreviewResponse()
+        data class PreviewDeferred(val experience: Experience, val frameId: String) : PreviewResponse()
+        data class StateMachineError(val experience: Experience, val error: Error) : PreviewResponse()
+    }
```
iujames, line 219: Could we simplify to remove `PreviewResponse`? I'm thinking that `show()` could just return `ResultOf<Unit, RenderingError>` and a preview is just another form of rendering that propagates that error back up. The `RenderingError` could include relevant networking errors that are applicable to preview.

appcues/src/main/java/com/appcues/DeepLinkHandler.kt
```diff
@@ -66,4 +77,23 @@ internal class DeepLinkHandler(
...
+    private suspend fun previewExperience(experienceId: String, activity: Activity) {
+        experienceRenderer.preview(experienceId).run {
+            val resources = activity.resources
+
+            when (this) {
+                is HttpError -> if (code == NETWORK_ERROR_NOT_FOUND) {
+                    resources.getString(R.string.appcues_preview_flow_not_found)
+                } else {
+                    resources.getString(R.string.appcues_preview_flow_failed)
+                }
```
iujames, line 90: Might be best suited to do the 404 check down in the ExperienceRenderer when the networking error is being handled, that way the logic is centralized and not the responsibility of any callers? There could be separate `RenderingError` options for the not found vs other case then, which bubble up to the caller.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -178,16 +192,22 @@ internal class ExperienceRenderer(
     }
 
     suspend fun show(renderContext: RenderContext) {
-        potentiallyRenderableExperiences[renderContext]?.let {
-            attemptToShow(it)
+        arrayListOf<Experience>().run {
```
andretortolano, line 195: this bit was missing, where we mix both lists and attempt to show

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -69,42 +82,43 @@ internal class ExperienceRenderer(
...
+    }
+
     /**
      * returns true/false whether the experience is was started
      */
-    suspend fun show(experience: Experience): Boolean = with(experience) {
-        var canShow = config.interceptor?.canDisplayExperience(experience.id) ?: true
+    suspend fun show(experience: Experience): RenderingResult = with(experience) {
+        // config interceptor determines that this experience should not be displayed
+        if (config.interceptor?.canDisplayExperience(id) == false) return CannotDisplayInterceptor
```
iujames, line 99: this change is not ok, i think - since it will now return early before checking the experiment - meaning the control group analytics are not tracked. I think the expectation is that the experiment analytics are tracked always there, like before

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
iujames: this now differs from iOS implementation here https://github.com/appcues/appcues-ios-sdk/pull/445/files#diff-e9daefd60bacc377827cb6a2002bcd0fe84cd5b3ec407cd9ab2a6a5dd676a37fR86-R89 - where if a preview exists, only that is attempted - not the merge

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
andretortolano: Did not get. a good impl working here, the network errors seems wrong to mix with RenderingResult here

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
andretortolano: you are right, I just reverted that back

samples/kotlin-android-app/build.gradle
```diff
@@ -26,8 +26,8 @@ android {
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
 
-        buildConfigField "String", "APPCUES_ACCOUNT_ID", "\"PLACEHOLDER_ACCOUNT_ID\""
-        buildConfigField "String", "APPCUES_APPLICATION_ID", "\"PLACEHOLDER_APPLICATION_ID\""
+        buildConfigField "String", "APPCUES_ACCOUNT_ID", "\"103523\""
+        buildConfigField "String", "APPCUES_APPLICATION_ID", "\"4b6b796f-1313-440a-b356-ee4b7abe931c\""
```
iujames, line 30: please revert this one and the manifest update

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
iujames: hey @mmaatttt just checking on this one - I think I'd agree - removing a preview as soon as it successfully renders seems appropriate.

iujames (approved):
Looks good! Just the sample app ID revert and ship it

samples/kotlin-android-app/build.gradle
andretortolano: ops

PR #460, opened 2023-08-21
Refactor embed state machine handling to detect frame not found errors

iujames:
This is a change proposed to more closely mirror the iOS implementation and allow us to correctly detect errors for "frame not found" but not trigger other error logging unexpectedly.

## Before
* `RenderContextManager` holds the `potentiallyRenderableExperiences`
* `RenderContextManager` holds the mapping from `RenderContext` to `StateMachine`
* `RenderContextManager` holds parallel mapping of `RenderContext` to `AppcuesFrameView` (weakly)

## After
* `ExperienceRenderer` holds the `potentiallyRenderableExperiences`
* `StateMachineDirectory` holds the reference from `RenderContext` to `StateMachineOwning` (interface)
* `StateMachineOwning` has two implementations - `ExperienceRenderer`, for traditional modal experiences, and `AppcuesFrameView`, for embeds - through the `AppcuesFrameStateMachineOwner`
* `AppcuesFrameStateMachineOwner` is where the weak reference to the frame view resides

This structure allows the `ExperienceRenderer` to control the lifecycle of all state machines, and accurately detect if a given `RenderContext` has been registered or not - and trigger special case errors.

For reference on the iOS side
* [StateMachineDirectory](https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Presentation/ExperienceRendering/StateMachineDirectory.swift)
* [ExperienceRenderer](https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift)
* [AppcuesFrameView](https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Presentation/UI/AppcuesFrameView.swift)


appcues/src/main/java/com/appcues/AppcuesViews.kt
```diff
@@ -29,7 +29,7 @@ public class AppcuesFrameView @JvmOverloads constructor(
         return composeView
     }
 
-    internal fun clearComposition() {
+    internal fun reset() {
```
iujames, line 32: just a renaming to match ios naming conventions

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -66,7 +85,12 @@ internal class ExperienceRenderer(
 
         if (!canShow) return false
 
-        val stateMachine = renderContextManager.getOrCreateStateMachines(renderContext)
+        val stateMachine = stateMachines.getOwner(renderContext)?.stateMachine
+
+        if (stateMachine == null) {
+            analyticsTracker.trackExperienceError(experience, "no render context $renderContext")
```
iujames, line 91: this is the new spot where we track the exact error in question, and avoid propagating other show failures that are already handled elsewhere

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -108,22 +146,41 @@ internal class ExperienceRenderer(
...
+            (owner as? AppcuesFrameStateMachineOwner)?.let {
+                it.frame.get()?.reset()
+                it.stateMachine = null
+            }
+        }
+
+        owner.stateMachine = get<StateMachine>().also {
+            lifecycleTracker.start(it, { potentiallyRenderableExperiences.remove(renderContext) })
+        }
+        stateMachines.setOwner(context, owner)
```
iujames, line 175: this is where we create and register the StateMachine just once, when the frame is registered - rather than on demand - so we can know if a frame should be valid or not

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -27,27 +28,45 @@ import com.appcues.util.ResultOf.Failure
...
 import kotlinx.coroutines.flow.SharedFlow
 import org.koin.core.component.KoinScopeComponent
+import org.koin.core.component.get
 import org.koin.core.component.inject
 import org.koin.core.scope.Scope
 
 internal class ExperienceRenderer(
     override val scope: Scope,
-) : KoinScopeComponent {
+) : KoinScopeComponent, StateMachineOwning {
```
andretortolano, line 37: do we need to have ExperienceRenderer implementing StateMachineOwning? or could wehave a separate Implementation inited ? for instance AppcuesModalStateMachineOwning

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
andretortolano: ok so instead of creating we just get instead, if null then we know the render context is not here. ok

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -27,27 +28,45 @@ import com.appcues.util.ResultOf.Failure
...
     private val sessionMonitor by inject<SessionMonitor>()
     private val config by inject<AppcuesConfig>()
     private val analyticsTracker by inject<AnalyticsTracker>()
-    private val renderContextManager by inject<RenderContextManager>()
+    private val stateMachines by inject<StateMachineDirectory>()
+    private val lifecycleTracker by inject<ExperienceLifecycleTracker>()
+
+    private val potentiallyRenderableExperiences = hashMapOf<RenderContext, List<Experience>>()
+
+    init {
```
andretortolano, line 55: this is the only difference right? we are always init the Modal state machine instead of relying on getOrCreate, and then we can just use "get" during the show and if that returns null it means that the render context is not valid.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -93,12 +117,26 @@ internal class ExperienceRenderer(
     }
 
     suspend fun show(experiences: List<Experience>) {
-        experiences.firstOrNull()?.let {
-            renderContextManager.putExperiences(experiences, it.trigger)
+        val trigger = experiences.firstOrNull()?.trigger
```
andretortolano, line 120: ok this piece is closely mirroring iOS

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -149,11 +206,12 @@ internal class ExperienceRenderer(
     }
 
     fun stop() {
-        renderContextManager.stop()
+        stateMachine?.stop() // modal
```
andretortolano, line 209: do we really need to have a reference of the modal context? cant we just grab it from the StateMachineDirectory?

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
iujames: I was trying to actually not create any new types, but just have `ExperienceRenderer` and `AppcuesFrameView` implement `StateMachineOwning` (similar to iOS side). This is not possible in Kotlin though, as a public type can use an internal Interface, but the properties that it adds must be public - and we cannot have a public StateMachine or RenderContext as those are internal types. This is why the `AppcuesFrameStateMachineOwning` got created - to have a separate internal type for that case. 

I think it would work fine to have a separate `AppcuesModalStateMachineOwning` as well, though I don't think it would actually get used for anything, and it also works to just have `ExperienceRenderer` implement the interface as done here - either way.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
iujames: yep exactly

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
iujames: good question - its already in the directory so it's probably sufficient to just stop all those, will double check that

andretortolano (approved):

PR #461, opened 2023-08-21
Update example app embed screen tracking name

iujames:
`Embed Harness` --> `Embed Container` 

Making this match with [iOS](https://github.com/appcues/appcues-ios-sdk/blob/main/Examples/DeveloperCocoapodsExample/CocoapodsExample/EmbedViewController.swift#L30) to avoid any test confusion

also using `/embed` instead of `/embeds` for the deep link route, for same reason


andretortolano (approved):
nice catch!

PR #463, opened 2023-08-23
Reset any rendering experiences / state machines on SDK reset

iujames:
Similar to the iOS update here https://github.com/appcues/appcues-ios-sdk/pull/446

`StateMachineOwning` has a `reset()` now which will stop any running state machine and dismiss any current experience in view. This also uses a new capability to end the experience without tracking analytics.

Also, small update in the `Appcues.identify()` to call `reset()` when the user ID changes - to ensure the same flow of events happens whether its a logout/reset or just a re-identify with new user - effectively the same as far as Appcues is concerned.

### DIFF
--- a/appcues/src/main/java/com/appcues/statemachine/Action.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/Action.kt
@@ -6,7 +6,11 @@ internal sealed class Action {
     data class StartExperience(val experience: Experience) : Action()
     data class StartStep(val stepReference: StepReference) : Action()
     object RenderStep : Action()
-    data class EndExperience(val markComplete: Boolean, val destroyed: Boolean) : Action()
+    data class EndExperience(
+        val markComplete: Boolean,
+        val destroyed: Boolean,
+        val trackAnalytics: Boolean = true,
+    ) : Action()
     object Reset : Action()
     data class ReportError(val error: Error, val fatal: Boolean) : Action()
 }
--- a/appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
+++ b/appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
@@ -169,24 +169,18 @@ internal class ExperienceRenderer(
 
     suspend fun start(owner: StateMachineOwning, context: RenderContext) {
         // If there's already a frame for the context, reset it back to its unregistered state.
-        val existingOwner = stateMachines.getOwner(context) as? AppcuesFrameStateMachineOwner
-        if (existingOwner != null) {
-            existingOwner.frame.get()?.reset()
-            existingOwner.stateMachine = null
-        }
+        stateMachines.getOwner(context)?.reset()
 
         // If the machine being started is already registered for a different context,
         // reset it back to its unregistered state before potentially showing new content.
-        if (owner.stateMachine != null) {
-            (owner as? AppcuesFrameStateMachineOwner)?.let {
-                it.frame.get()?.reset()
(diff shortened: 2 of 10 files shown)


appcues/src/main/java/com/appcues/statemachine/Action.kt
```diff
@@ -6,7 +6,11 @@ internal sealed class Action {
     data class StartExperience(val experience: Experience) : Action()
     data class StartStep(val stepReference: StepReference) : Action()
     object RenderStep : Action()
-    data class EndExperience(val markComplete: Boolean, val destroyed: Boolean) : Action()
+    data class EndExperience(
+        val markComplete: Boolean,
+        val destroyed: Boolean,
+        val trackAnalytics: Boolean = true,
```
iujames, line 12: this flag can now be used to optionally disable analytics when closing out an experience - for cases like this reset or the existing `stop()` feature. Open to any better ideas, but this was first approach I could think of.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -169,24 +169,18 @@ internal class ExperienceRenderer(
...
-        if (owner.stateMachine != null) {
-            (owner as? AppcuesFrameStateMachineOwner)?.let {
-                it.frame.get()?.reset()
-                it.stateMachine = null
-            }
-        }
+        owner.reset()
 
+        // do we need to unhook any existing listeners to owner.stateMachine, if not null, before
+        // re-assigning a new one here?
```
iujames, line 179: open question here, I'm not sure how to stop the observers on the state flow in the state machines at this point. It seems like we'd need to create and hold references to a coroutine somewhere and be able to cancel them? There does not appear to be any other way to stop a state flow collector. https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-state-flow/

andretortolano: so we either change lifecycleTracker to be an instance per state machine, or we store job references by render context inside of lifecycleTracker, and then call a method like "lifecycleTracker.stop(renderContext)" that would cancel those jobs

iujames: i like the idea of each state machine owning responsibility for its own tracker - will explore that

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -169,24 +165,22 @@ internal class ExperienceRenderer(
...
-            (owner as? AppcuesFrameStateMachineOwner)?.let {
-                it.frame.get()?.reset()
-                it.stateMachine = null
-            }
-        }
+        owner.reset()
+
+        // if the owner already has a state machine, make sure it's lifecycle tracking is stopped before
+        // we replace it with a new one
+        owner.stateMachine?.stopLifecycleTracking()
```
iujames, line 176: @andretortolano the second commit here now handles this - and this line in particular is where we stop any existing listeners when re-creating a state machine for a new frame registration.

`StateMachine` now owns its own `ExperienceLifecycleTracker`, and starts it on init. The tracker is a `factory` in Koin now, with each `StateMachine` getting its own instance. The tracker keeps track of the `Job` references for its observers, and they can be canceled when it is stopped.

andretortolano (approved):
good change around Lifecycle tracker for each state machine

PR #464, opened 2023-08-23
✨ Debugger displaying multiple experiences

andretortolano:
adding debugger ability to read multiple experiences.

suggestion to add the frame id when it exists.

image

### DIFF
--- a/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
+++ b/appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
@@ -7,6 +7,7 @@ import com.appcues.analytics.ExperienceLifecycleEvent.StepInteraction.Interactio
 import com.appcues.analytics.ExperienceLifecycleEvent.StepInteraction.InteractionType.TARGET_TAPPED
 import com.appcues.data.model.Experience
 import com.appcues.data.model.ExperienceStepFormState
+import com.appcues.data.model.getFrameId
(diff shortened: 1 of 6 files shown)


iujames (approved):
nice, i like the idea of showing the frame ID in the debugger. Adding to the actual analytics data seems harmless, it would just be another field that gets put into Snowflake afaik. The only other way I could think to do that would be to scan the StateMachines that are currently in the directory, and look at the current experience in each one to find any that match the ID, then grab the render context / frame ID from there. fyi to @mmaatttt on the topic of adding frame ID, for consistency.

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleEvent.kt
```diff
@@ -85,6 +86,7 @@ internal sealed class ExperienceLifecycleEvent(
             "experienceId" to experience.id.appcuesFormatted(),
             "experienceName" to experience.name,
             "experienceType" to experience.type,
+            "frameId" to experience.renderContext.getFrameId(),
```
mmaatttt, line 89: Is this an empty string for a modal?

iujames: looks like it is. good catch - ideally a null / trimmed out

andretortolano: its null. is that ok?

PR #465, opened 2023-08-24
🐛 Removing FillMaxSize from consumeAllTouch

andretortolano:
Small fix around sizing of AppcuesComposition, specifically targeting embeds


appcues/src/main/java/com/appcues/ui/composables/AppcuesComposition.kt
```diff
@@ -89,9 +88,7 @@ private fun MainSurface() {
 
 @OptIn(ExperimentalComposeUiApi::class)
 private fun Modifier.consumeAllTouchEvents(): Modifier = then(
-    Modifier
-        .fillMaxSize()
-        .pointerInteropFilter { false }
+    Modifier.pointerInteropFilter { false }
```
iujames, line 91: Could just inline this at the callsite since only a single modifier?

iujames (approved):

PR #469, opened 2023-08-25
Add embed related experience renderer tests

iujames:
Wrote some tests to validate the basic behaviors around embed registration / qualification / rendering.

Also, proposing a change here to make `StateMachineOwning` use a `suspend` function for `reset()` - why? I believe I've discovered that if a frame is re-registered rapidly, you could have a situation where the previous state machine usage is still dismissing - an async behavior that could require some dismissal of UI, while the new render is attempting to start. I think this could lead to an edge case of rendering issues where one experience cannot start while the other is ending. I think it is safest to have the caller wait on the `reset()`, rather than having it spawn its own internal coroutine, then continue with the new render attempt.

### DIFF
--- a/appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
@@ -223,19 +223,17 @@ internal class StateMachine(
         }
     }
 
-    fun stop(dismiss: Boolean) {
-        appcuesCoroutineScope.launch {
-            handleAction(
-                EndExperience(
-                    markComplete = false,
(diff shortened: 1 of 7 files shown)


appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
```diff
@@ -223,19 +223,17 @@ internal class StateMachine(
         }
     }
 
-    fun stop(dismiss: Boolean) {
-        appcuesCoroutineScope.launch {
```
iujames, line 227: this is the spot of concern - if this launches it's own coroutine, the caller may continue before this `EndExperience` actually completes

andretortolano (approved):

PR #472, opened 2023-08-28
🐛 Passing activity context instead of Generic context

andretortolano:
small change to pass in the correct Context so that Shot tests can capture screen properly

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/presentation/EmbedViewPresenter.kt
+++ b/appcues/src/main/java/com/appcues/ui/presentation/EmbedViewPresenter.kt
@@ -1,5 +1,6 @@
 package com.appcues.ui.presentation
 
+import android.app.Activity
 import android.view.ViewGroup
 import androidx.compose.ui.platform.ComposeView
 import androidx.core.view.isVisible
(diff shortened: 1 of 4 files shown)


appcues/src/main/java/com/appcues/ui/presentation/EmbedViewPresenter.kt
```diff
@@ -15,7 +16,7 @@ internal class EmbedViewPresenter(
     private val stateMachines: StateMachineDirectory,
 ) : ViewPresenter(scope, renderContext) {
 
-    override fun ViewGroup.setupView(): ComposeView? {
+    override fun ViewGroup.setupView(activity: Activity): ComposeView? {
         return stateMachines.getFrame(renderContext)?.setupComposeView()
```
iujames, line 20: should the `activity` pass through to `setupComposeView()` here? I guess it does not matter, as the `AppcuesFrameView` is already created with a `context` that was used to create the internal `composeView` there.

iujames (approved):

appcues/src/main/java/com/appcues/ui/presentation/EmbedViewPresenter.kt
andretortolano: no need, its using whatever it is coming from the original layout

PR #475, opened 2023-08-30
State machine updates/simplifications to handle rendering issues

iujames:
These are the changes from a deeper review of state machine handling, in particular related to recent embeds work.

* Remove await on presentation callbacks to avoid issues with hung state machine, no longer needed
* Process trait metadata before presenting new container or moving to new steps in container, and handle errors prior to render
* Add remove() to PresentingTraits to ensure that UI can be dismissed in trait metadata error cases
* Simplify StateMachineOwning implementation (for embeds) to remove optionals and handle edge cases

### DIFF
--- a/appcues/src/main/java/com/appcues/statemachine/BeginningStepExt.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/BeginningStepExt.kt
@@ -24,9 +24,12 @@ internal suspend fun BeginningStep.presentContainer(
         produceMetadataWithRetry()
 
         // kick off UI
-        presentingTrait()?.present()
-
-        RenderStep
+        presentingTrait()?.run {
+            present()
+            RenderStep
+        } ?: ReportError(toStepError(AppcuesTraitException("Unable to present step $flatStepIndex")), true)
+        // the exception above would only happen if the step index was not a valid step in the experience and
+        // we could not find the group for that step - should never really happen
     } catch (exception: AppcuesTraitException) {
         ReportError(toStepError(exception), true)
     }
--- a/appcues/src/main/java/com/appcues/statemachine/BeginningStepExt.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/BeginningStepExt.kt
@@ -1,0 +1,81 @@
+package com.appcues.statemachine
+
+import com.appcues.action.ActionProcessor
+import com.appcues.action.ExperienceAction
+import com.appcues.statemachine.Action.RenderStep
+import com.appcues.statemachine.Action.ReportError
+import com.appcues.statemachine.Error.StepError
+import com.appcues.statemachine.State.BeginningStep
+import com.appcues.trait.AppcuesTraitException
+import com.appcues.trait.MetadataSettingTrait
+import com.appcues.trait.PresentingTrait
+import kotlinx.coroutines.delay
+
+internal suspend fun BeginningStep.presentContainer(
+    actionProcessor: ActionProcessor,
+    actions: List<ExperienceAction>,
+): Action {
+    // first, process any pre-step actions that need to be handled before the container is presented.
+    // for example - navigating to another screen in the app
+    actionProcessor.process(actions)
+
+    return try {
+        // this could throw an exception on trait failure, handled below
+        produceMetadataWithRetry()
+
+        // kick off UI
+        presentingTrait()?.present()
+
+        RenderStep
+    } catch (exception: AppcuesTraitException) {
+        ReportError(toStepError(exception), true)
+    }
+}
+
+internal fun BeginningStep.presentingTrait(): PresentingTrait? {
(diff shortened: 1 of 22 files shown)


appcues/src/main/java/com/appcues/statemachine/BeginningStepExt.kt
```diff
@@ -0,0 +1,81 @@
...
+import com.appcues.statemachine.Action.RenderStep
+import com.appcues.statemachine.Action.ReportError
+import com.appcues.statemachine.Error.StepError
+import com.appcues.statemachine.State.BeginningStep
+import com.appcues.trait.AppcuesTraitException
+import com.appcues.trait.MetadataSettingTrait
+import com.appcues.trait.PresentingTrait
+import kotlinx.coroutines.delay
+
+internal suspend fun BeginningStep.presentContainer(
```
andretortolano, line 14: good call moving this away from Transitions.kt

appcues/src/main/java/com/appcues/statemachine/SideEffect.kt
```diff
@@ -1,17 +1,19 @@
...
 
 internal sealed class SideEffect {
     data class ContinuationEffect(val action: Action) : SideEffect()
     data class PresentContainerEffect(
-        val experience: Experience,
-        val containerIndex: Int,
-        val completion: CompletableDeferred<ResultOf<State, Error>>,
-        val actions: List<ExperienceAction>,
+        val presentContainer: suspend (ActionProcessor) -> Action
+//        val experience: Experience,
```
andretortolano, line 12: 👀

appcues/src/main/java/com/appcues/trait/PresentingTrait.kt
```diff
@@ -16,4 +16,9 @@ internal interface PresentingTrait : ExperienceTrait {
      */
     @Throws(AppcuesTraitException::class)
     fun present()
+
+    /**
+     * Remove the container for the applicable group of steps in the Experience.
+     */
+    fun remove()
```
andretortolano, line 23: 👍

appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
```diff
@@ -185,7 +191,7 @@ internal class TooltipTrait(
                                 contentSized.value = true
                             }
                             .styleShadowPath(style, tooltipPath, isSystemInDarkTheme())
-                            .clipToPath(tooltipPath)
+                            .clip(GenericShape { _, _ -> addPath(tooltipPath) })
```
andretortolano, line 194: Ok, was that maybe to solve detekt ? in any case its ok cleanup

appcues/src/main/java/com/appcues/statemachine/SideEffect.kt
iujames: ah yeah will clean up these old things

appcues/src/main/java/com/appcues/trait/appcues/TooltipTrait.kt
iujames: yeah it was just to reduce function count for detekt, inlining a one line function that was just used in one spot

andretortolano (approved):
🚀

appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
```diff
@@ -149,8 +155,10 @@ internal class StateMachine(
     private fun takeValidStateTransitions(state: State, action: Action) = with(Transitions) {
         when {
             // Idling
-            state is Idling && action is StartExperience ->
+            state is Idling && action is StartExperience -> {
+                startLifecycleTracking()
```
iujames, line 159: I noticed that we were never really stopping lifecycle tracking before. So revised this - now we start tracking here when we notice any move from `Idling` to `StartExperience`. Then we stop tracking on L136 on any return to `Idling`.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -163,25 +155,27 @@ internal class ExperienceRenderer(
...
     }
 
-    suspend fun start(owner: StateMachineOwning, context: RenderContext) {
+    suspend fun start(frame: AppcuesFrameView, context: RenderContext) {
         // If there's already a frame for the context, reset it back to its unregistered state.
         stateMachines.getOwner(context)?.reset()
 
-        // If the machine being started is already registered for a different context,
-        // reset it back to its unregistered state before potentially showing new content.
-        owner.reset()
```
iujames, line 172: this was never doing anything, since we were passing a new `owner` in each time, and it never had a state machine.

the new and improved check will actually look for the case of the same `frame` already being registered in a different context, and reset that old context if so - since the new registration should be unique for this frame.

PR #477, opened 2023-09-05
Dependency updates

iujames:
Taking a pass through, post 3.0 release, to update various dependencies and settings. Will need to consider the best time to push these updates out, and test with cross platform toolkits still.

* AGP 8.0.2 -> 8.1.1
* Kotlin 1.8.21 -> 1.9.10
* compileSdk and targetSdk 33 -> 34
* migrate [from kapt to ksp](https://developer.android.com/build/migrate-to-ksp)
* Room 2.4.3 -> 2.5.2 (required for Kotlin 1.9)
* Compose 1.4.3 -> 1.5.0 (compiler 1.4.7 -> 1.5.3)
* Koin 3.4.2 -> 3.5.0
* Coil 2.2.2 -> 2.4.0
* Moshi 1.14.0 -> 1.15.0
* few other misc androidx lib patch updates

removed usage of Accompanist entirely
* the Navigation lib is now rolled into compose navigation in the core
* the WebView lib is [deprecated](https://google.github.io/accompanist/web/), and replaced with a simple port into our code using an AndroidView containing a WebView, [matching what we were using before](https://github.com/google/accompanist/blob/main/web/src/main/java/com/google/accompanist/web/WebView.kt#L234)

Code changes are to fix any API changes or deprecations related to the above - with exception of one bug fix on the web view, will note inline.

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/InAppReviewActivity.kt
+++ b/appcues/src/main/java/com/appcues/ui/InAppReviewActivity.kt
@@ -2,6 +2,8 @@ package com.appcues.ui
 
 import android.content.Context
 import android.content.Intent
+import android.os.Build.VERSION
+import android.os.Build.VERSION_CODES
 import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.os.bundleOf
@@ -40,7 +42,12 @@ internal class InAppReviewActivity : AppCompatActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         // remove enter animation from this activity
-        overridePendingTransition(0, 0)
+        if (VERSION.SDK_INT >= VERSION_CODES.UPSIDE_DOWN_CAKE) {
+            overrideActivityTransition(OVERRIDE_TRANSITION_OPEN, 0, 0)
+        } else {
+            @Suppress("DEPRECATION")
+            overridePendingTransition(0, 0)
+        }
         super.onCreate(savedInstanceState)
 
         val requestCompletion = CompletableDeferred<Boolean>()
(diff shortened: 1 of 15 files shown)


appcues/src/main/java/com/appcues/ui/primitive/EmbedHtmlPrimitive.kt
```diff
@@ -63,15 +99,13 @@ internal class EmbedChromeClient(private val container: ViewGroup) : Accompanist
...
         container.addView(view)
-        container.visibility = View.VISIBLE
     }
 
     override fun onHideCustomView() {
         if (customView == null) {
             return
         }
         customView?.visibility = View.GONE
-        container.visibility = View.GONE
```
iujames, line 74: this is the one bug fix - this is currently breaking the return from fullscreen video in v3.0 - setting the container view GONE - this is just the root view of the decorView now (not a separate overlay view) so we cannot set it GONE.

appcues/src/main/java/com/appcues/ui/InAppReviewActivity.kt
```diff
@@ -40,7 +42,12 @@ internal class InAppReviewActivity : AppCompatActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         // remove enter animation from this activity
-        overridePendingTransition(0, 0)
+        if (VERSION.SDK_INT >= VERSION_CODES.UPSIDE_DOWN_CAKE) {
+            overrideActivityTransition(OVERRIDE_TRANSITION_OPEN, 0, 0)
+        } else {
+            @Suppress("DEPRECATION")
+            overridePendingTransition(0, 0)
```
iujames, line 49: https://developer.android.com/reference/android/app/Activity#overridePendingTransition(int,%20int)

iujames:
Tested this update out in the cross-platform plugins
* [React Native](https://github.com/appcues/appcues-react-native-module/pull/108) - was the most involved, involved updating the example app to use the latest version
* [Flutter](https://github.com/appcues/appcues-flutter-plugin/pull/55) - relatively straightforward, update compile SDK 34 and Kotlin version
* [Ionic](https://github.com/appcues/appcues-capacitor-plugin/pull/114) - relatively straightforward, update compile SDK 34 and Kotlin version

andretortolano (approved):

PR #479, opened 2023-09-08
✨  New Analytics architecture

andretortolano:
Proposing an architectural change to our core SDK  code.

1. Replace AnalyticsTracker with Analytics and Experience.
2. Replace SessionMonitor with SessionService.
3. Replace Storage with SessionSource.
4. Replace ExperienceRenderer with RenderingService.

Analytics class properly queue to AnalyticIntent channel all incoming events, Intents sequentially verified against SessionService, if it goes through this check we map all Intent to Activity that queue up to AnalyticsQueue following the same logic as before.

Experiences class provide a way to directly access experience information (for given RenderContext) or to show and preview experiences directly from the API.

By changing how Analytics internally work we can take down a old [StrictMode warning](https://app.shortcut.com/appcues/story/53440/android-strictmode-on-storage-read) whenever we currently use Storage class, thats because all access to shared preferences is considered IO operation and should never be done on the mainUIThread.

Besides that this change brings some layering of abstraction that ensure clear architectural boundaries between SDK internal modules.


appcues/src/main/java/com/appcues/analytics/Analytics.kt
```diff
@@ -0,0 +1,188 @@
...
+    suspend fun startSession(intent: AnalyticsIntent): Boolean
+
+    // decorates incoming intent and returns a new intent with session/auto props information
+    suspend fun decorateIntent(intent: AnalyticsIntent): AnalyticsIntent
+
+    suspend fun reset()
+}
+
+// new model for Intents available for Analytics
+internal sealed class AnalyticsIntent(open val isInternal: Boolean, val timestamp: Date = Date()) {
```
iujames, line 50: i think this is somewhat similar to the usage in `TrackingUpdate` on the iOS side, fyi https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Data/Analytics/TrackingUpdate.swift

appcues/src/main/java/com/appcues/analytics/Analytics.kt
```diff
@@ -0,0 +1,188 @@
...
+) : IntentProcessor, ApplicationMonitor.Listener {
+
+    companion object {
+
+        const val EVENT_ATTR_SCREEN_TITLE = "screenTitle"
+        const val EVENT_CONTEXT_SCREEN_TITLE = "screen_title"
+    }
+
+    init {
+        intentQueue.setProcessor(this)
```
iujames, line 93: following the flow: this class pushes things into the queue, then the queue calls back here to `process` things as needed based on the logic around batching items.

There is still a circular dependency there it seems, and I am wondering if there is any situation you could get in where `process` ends up spawning multiple coroutines in parallel and not guaranteed to process the queue in order?

would it make sense to have the intentQueue actually be the one managing the coroutine and ensuring things are processed in order? maybe it could pipe things through it's own Channel even to assist (how we did it with ActionProcessor for example)

appcues/src/main/java/com/appcues/data/session/PrefSessionLocalSource.kt
```diff
@@ -0,0 +1,94 @@
...
+) {
+
+    companion object {
+
+        private const val KEY_DEVICE_ID = "appcues.deviceId"
+        private const val KEY_USER_ID = "appcues.userId"
+        private const val KEY_GROUP_ID = "appcues.groupId"
+        private const val KEY_IS_ANONYMOUS = "appcues.isAnonymous"
+        private const val KEY_LAST_CONTENT_SHOWN_AT = "appcues.lastContentShownAt"
+        private const val KEY_USER_SIGNATURE = "appcues.userSignature"
```
iujames, line 23: confirmed these match existing values so no issue with upgrade path 👍

appcues/src/main/java/com/appcues/data/session/PrefSessionLocalSource.kt
```diff
@@ -0,0 +1,94 @@
...
+
+    suspend fun getUserSignature(): String? = withContext(Dispatchers.IO) {
+        return@withContext getSharedPrefs().getString(KEY_USER_SIGNATURE, null)
+    }
+
+    suspend fun setUserSignature(userSignature: String?) = withContext(Dispatchers.IO) {
+        getSharedPrefs().edit().putString(KEY_USER_SIGNATURE, userSignature).apply()
+    }
+
+    suspend fun reset() = withContext(Dispatchers.IO) {
```
iujames, line 86: maybe should clear `KEY_LAST_CONTENT_SHOWN_AT` as well now that I'm thinking about it - that represents the last content the current user has seen

appcues/src/main/java/com/appcues/analytics/Analytics.kt
```diff
@@ -0,0 +1,188 @@
...
+
+        ApplicationMonitor.subscribe(this)
```
andretortolano, line 95: ApplicationMonitor here is simplifying an existing logic on SessionMonitor that clears analytics when we know the application lifecycle calls onStopped()

appcues/src/main/java/com/appcues/qualification/DefaultQualificationService.kt
```diff
@@ -0,0 +1,241 @@
...
+) : QualificationService {
+
+    private val processingActivity: HashSet<UUID> = hashSetOf()
+    private val mutex = Mutex()
+
+    override suspend fun qualify(intents: List<AnalyticsIntent>): QualificationResult? {
+        val activityRequest = intents.mergeToActivityRequest()
+
+        // FIXME trying to recall reason why old call site flushPendingActivity could accept null as date.
+        SdkMetrics.tracked(activityRequest.requestId, Date())
```
iujames, line 52: if i recall, it was because _some_ network calls we know are not qualifying - just emptying the queue after 10 sec, like background flow analytics, and we didn't want to track metrics for those?

appcues/src/main/java/com/appcues/analytics/Analytics.kt
```diff
@@ -0,0 +1,188 @@
...
+
+    fun anonymous() {
+        intentQueue.flushAndProcess(Anonymous)
+    }
+
+    fun identify(userId: String, properties: Map<String, Any>? = null) {
+        intentQueue.flushAndProcess(Identify(userId, properties))
+    }
+
+    fun updateProfile(properties: Map<String, Any>? = null, isInternal: Boolean) {
```
andretortolano, line 110: adding updateProfile is an idea that make explicitly for places like: UpdateProfileAction we would need to add it to the API.

appcues/src/main/java/com/appcues/qualification/DefaultQualificationService.kt
```diff
@@ -0,0 +1,241 @@
...
+
+        return ActivityRequest(
+            accountId = config.accountId,
+            // at this point it should never be null
+            userId = sessionLocalSource.getUserId() ?: String(),
+            profileUpdate = profileUpdate,
+            groupId = sessionLocalSource.getGroupId(),
+            groupUpdate = groupUpdate,
+            events = events,
+            userSignature = sessionLocalSource.getUserSignature(),
```
iujames, line 92: the usage of `sessionLocalSource` here makes me possibly uneasy - what if the session data changed from the moment when this event was tracked to the moment here where we are sending it - is that possible? Would it be possible to capture that session data into the AnalyticsIntent earlier? at the moment of initial processing? that way each analytics item will be an atomic set of data that we know matches truly to the time it was captured.

appcues/src/main/java/com/appcues/session/DefaultSessionService.kt
```diff
@@ -0,0 +1,187 @@
...
+import java.util.Date
+import java.util.UUID
+import java.util.concurrent.TimeUnit
+
+internal class DefaultSessionService(
+    private val config: AppcuesConfig,
+    private val sessionLocalSource: PrefSessionLocalSource,
+    private val contextResources: ContextResources,
+    private val sessionRandomizer: SessionRandomizer,
+) : SessionService {
```
iujames, line 27: this is effectively merging the session management with the auto property decoration, is that accurate? If so, I think it might be nice still to separate those responsibilities.

iujames (commented):
I think this is good clean up! It makes sense to take a second look at the responsibilities here and see how we can handle the analytics pipeline more cleanly.

key things after the initial work will be to clean up the old/dead code as well as revise tests to make sure we don't have any unexpected regressions, since this is so core to the system

appcues/src/main/java/com/appcues/analytics/Analytics.kt
iujames: Would not need in the public API right? the UpdateProfileAction execute is already a suspend fun, so it could get the user ID from storage just like it does now, but safely?

appcues/src/main/java/com/appcues/analytics/Analytics.kt
andretortolano: I think this somewhat better than just a circular dependency, its a callback within a limited scope.

process could be suspend fun and we could launch coroutine scope before calling it for sure, I think we can explore this idea and ensure whenever we say Flush its going in on the same order

appcues/src/main/java/com/appcues/data/session/PrefSessionLocalSource.kt
andretortolano: yes I mainly tried to recreate what we were clearing on Appcues.reset, but good to revise.

appcues/src/main/java/com/appcues/qualification/DefaultQualificationService.kt
andretortolano: ok I guess in that case it would be interesting to change how SdkMetrics work in a way that we figure out if we should track activityRequest or not, without depending on the timestamp value, then we can change this call to track(activityRequest) and not worry about loose logic around timestamp being null or not

appcues/src/main/java/com/appcues/qualification/DefaultQualificationService.kt
andretortolano: I think if we make sure that when we call "process" we know it will wait until its done to call the next then we make sure there is no way information will change, right?

appcues/src/main/java/com/appcues/session/DefaultSessionService.kt
andretortolano: this class is a combination of SessionMonitor and AutoPropertyDecorator, yes. I think the sessionProperty has room here, the applicationProperties and autoProperties could be separate in a way, how would you do that?

appcues/src/main/java/com/appcues/analytics/Analytics.kt
andretortolano: yes, I think it could also import Analytics and call this directly.

andretortolano: this is a side idea, that maybe this could be beneficial for the customer to have this option to update the current user profile without the needing to store the current user somewhere else. good utility method maybe

iujames: perhaps, I'd rather not expand the public API though unless there is very clear demand for it. I'd assume the customer app already knows and stores the user information somewhere for it's own purposes so probably not a big deal. And maintains consistency with web and all the other mobile platforms to keep our identify call behavior as-is.

appcues/src/main/java/com/appcues/analytics/Analytics.kt
iujames: yeah, key thing is that everything is sent out in the desired order, however that happens.

appcues/src/main/java/com/appcues/qualification/DefaultQualificationService.kt
iujames: yeah i'm not 100% sure on previous usage either, need to investigate that one more - but surely it can change if we can find a more appropriate way to do it.

appcues/src/main/java/com/appcues/qualification/DefaultQualificationService.kt
iujames: yeah probably so - I'd feel most comfortable if the values were immediately stuck into the AnalyticsIntent, so that we are guaranteed that nothing downstream could affect the user/signature/group information.

andretortolano: thats one of the pieces that surely needs to change 😭 , because we queue Intents on the main thread which makes sense, but we cant decorate it while doing so.

appcues/src/main/java/com/appcues/analytics/AnalyticsIntentQueue.kt
```diff
@@ -0,0 +1,121 @@
...
+import com.appcues.analytics.AnalyticsIntentQueue.QueueAction.FLUSH_THEN_PROCESS
+import com.appcues.analytics.AnalyticsIntentQueue.QueueAction.QUEUE
+import com.appcues.analytics.AnalyticsIntentQueue.QueueAction.QUEUE_THEN_FLUSH
+import java.util.Timer
+import java.util.TimerTask
+import kotlin.concurrent.schedule
+
+internal class AnalyticsIntentQueue(private val scheduler: QueueScheduler) {
+
+    enum class QueueAction {
```
andretortolano, line 13: Enum helps to push to the channel the proper "QueueAction" for each Intent.

appcues/src/main/java/com/appcues/analytics/AnalyticsIntentQueue.kt
```diff
@@ -0,0 +1,121 @@
...
+            QUEUE -> queue(intent)
+            QUEUE_THEN_FLUSH -> queueThenFlush(intent)
+            FLUSH_THEN_PROCESS -> flushThenProcess(intent)
+        }
+    }
+
+    fun flush() {
+        // to be called when any pending activity should immediately be flushed to cache, and network if possible
+        // i.e. app going to background / being killed
+        synchronized(this) {
```
andretortolano, line 56: still keeping the synchronized blocks here but I'm not sure they are needed, maybe just as a safeguard but probably not required since everything that is coming to this class comes from the same coroutine thread.

appcues/src/main/java/com/appcues/qualification/DefaultQualificationService.kt
```diff
@@ -0,0 +1,235 @@
...
+        SdkMetrics.tracked(activityRequest.requestId, Date())
+
+        return trackActivity(activityRequest).also {
+            // We can check here if there are experiences or else we just cancel the SdkMetrics track
+            if (it == null || it.experiences.isEmpty())
+                SdkMetrics.remove(activityRequest.requestId)
+        }
+    }
+
+    private fun List<AnalyticsActivity>.mergeToActivityRequest(): ActivityRequest? {
```
andretortolano, line 58: note to self: need to double check if this "merge" is correct

appcues/src/main/java/com/appcues/session/DefaultSessionService.kt
```diff
@@ -0,0 +1,176 @@
...
+            }
+        }
+
+        return this
+    }
+
+    private fun Map<String, Any>?.sanitized(): Map<String, Any> {
+        return mutableMapOf<String, Any>().apply {
+            putAll(this@sanitized ?: hashMapOf())
+            remove(KEY_USER_SIGNATURE)
```
andretortolano, line 110: its very convenient that we "access" and also "clear" the user signature here in SessionService. dont need to about this property anywhere else.

appcues/src/main/java/com/appcues/qualification/DefaultQualificationService.kt
andretortolano: need to figure out another way to filter out unwanted Activity requests  from SdkMetrics, or if we dont care we could just "track" all of them since its not a big lift, we just report those that come back with qualified experiences that can show == true anyway.

appcues/src/main/java/com/appcues/qualification/DefaultQualificationService.kt
andretortolano: its probably better now!

appcues/src/main/java/com/appcues/session/DefaultSessionService.kt
andretortolano: broke this down into two classes again, thanks for the feedback

appcues/src/main/java/com/appcues/analytics/Analytics.kt
```diff
@@ -0,0 +1,236 @@
...
+    ) : AnalyticsIntent(isInternal, properties)
+
+    data class Screen(
+        override val isInternal: Boolean,
+        val title: String,
+        override val properties: Map<String, Any>?
+    ) : AnalyticsIntent(isInternal, properties)
+}
+
+internal data class AnalyticsActivity(
```
andretortolano, line 107: this looks similar to ActivityRequest but its supposed to be an improvement over it, at the time when we "map" intents to AnalyticsActivity  we know we always have ONE event instead of a list that we can send on ActivityRequest. this way its easier to map from one to another, and read it through analyticsFlow

appcues/src/main/java/com/appcues/analytics/Analytics.kt
```diff
@@ -0,0 +1,236 @@
...
+    val groupId: String?,
+    val profileProperties: Map<String, Any>?,
+    val groupProperties: Map<String, Any>?,
+    val eventName: String?,
+    val eventAttributes: Map<String, Any>?,
+    val eventContext: Map<String, Any>?,
+    val userSignature: String?,
+)
+
+internal class Analytics(
```
andretortolano, line 122: note about interfaces used here.

they are architectural bounds, we may them or not in the future, but the idea is that this "core" piece of our project could be isolated in a module, and we could have different modules providing the implementation on those dependencies.

appcues/src/main/java/com/appcues/analytics/DefaultActivityBuilder.kt
```diff
@@ -0,0 +1,148 @@
...
+import com.appcues.analytics.AnalyticsIntent.Anonymous
+import com.appcues.analytics.AnalyticsIntent.Event
+import com.appcues.analytics.AnalyticsIntent.Identify
+import com.appcues.analytics.AnalyticsIntent.Screen
+import com.appcues.analytics.AnalyticsIntent.UpdateGroup
+import com.appcues.analytics.AnalyticsIntent.UpdateProfile
+import com.appcues.util.ContextResources
+import java.util.Date
+
+internal class DefaultActivityBuilder(
```
andretortolano, line 21: class is responsible for mapping AnalyticsIntent to AnalyticsActivity, this means simplifying from Anonymous, Identify and UpdateProfile to the AnalyticType.IDENTIFY for instance. + adding the necessary properties, attributes and context values

appcues/src/main/java/com/appcues/analytics/DefaultActivityBuilder.kt
```diff
@@ -0,0 +1,148 @@
...
+            is Screen -> mutableMapOf<String, Any>().apply {
+                // screen events got extra context key that defines screen title
+                put(SCREEN_CONTEXT_TITLE, intent.title)
+                putAll(contextProperties)
+            }
+            else -> null
+        }
+    }
+
+    private fun AnalyticsIntent.toAnalyticType(): AnalyticType {
```
andretortolano, line 138: this method is a clear demonstration of whats happening here.

appcues/src/main/java/com/appcues/debugger/DebuggerRecentEventsManager.kt
```diff
@@ -98,6 +131,77 @@ internal class DebuggerRecentEventsManager(
...
+                        title = contextResources.getString(R.string.appcues_debugger_event_details_properties_title),
+                        properties = activity.groupProperties?.toSortedList(),
+                    )
+                )
+            )
+        )
+        lastEventId++
+    }
+
+    private fun onActivityEvent(activity: AnalyticsActivity) {
```
andretortolano, line 152: comparing this to the "old" method that gets an ActivityRequest as input, this one is straight forward since we know that every activity = an EventItem, whereas before we had to iterate the events list (that always contained ONE element at this moment)

appcues/src/main/java/com/appcues/debugger/DebuggerStatusManager.kt
```diff
@@ -69,6 +70,45 @@ internal class DebuggerStatusManager(
         connectToAppcues(false)
     }
 
+    suspend fun onActivity(activity: AnalyticsActivity) = withContext(Dispatchers.IO) {
```
andretortolano, line 73: same here, one activity is equal to possibly one event.

appcues/src/main/java/com/appcues/statemachine/StateMachineFactory.kt
```diff
@@ -0,0 +1,17 @@
+package com.appcues.statemachine
+
+import com.appcues.AppcuesConfig
+import com.appcues.AppcuesCoroutineScope
+import com.appcues.action.ActionProcessor
+import com.appcues.analytics.ExperienceLifecycleTracker
+import com.appcues.statemachine.State.Idling
+
+internal class StateMachineFactory(
```
andretortolano, line 9: instead of relying on Koin its much more readable that we have a factory that produces state machines.

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -122,7 +121,12 @@ public class Appcues internal constructor(koinScope: Scope) {
      * @param properties Optional properties that provide additional context about the user.
      */
     public fun identify(userId: String, properties: Map<String, Any>? = null) {
-        identify(false, userId, properties)
+        if (userId.isEmpty()) {
```
andretortolano, line 124: not sure if I could validate user id further down the line,  the return with log here could be placed somewhere else.

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -166,21 +163,7 @@ public class Appcues internal constructor(koinScope: Scope) {
...
-        storage.isAnonymous = true
-        storage.groupId = null
-
-        debuggerManager.reset()
-
-        appcuesCoroutineScope.launch {
-            // stopping running experiences runs async, as it may require dismissal of UI.
-            experienceRenderer.resetAll()
-        }
+        analytics.reset()
```
andretortolano, line 166: need to double check if reset kept same behavior at the end

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -65,8 +56,6 @@ internal class ExperienceLifecycleTracker(
                 when (it) {
                     is RenderingStep -> {
                         if (it.isFirst) {
-                            // update this value for auto-properties
-                            storage.lastContentShownAt = Date()
```
andretortolano, line 69: this was out of place, now we can easily update this property when ExperienceStarted gets "pre-processed" by SessionService [here](https://github.com/appcues/appcues-android-sdk/pull/479/files#diff-54273f483b977ea6b839f9f1119798f237f0214e1725c22497a6df34933f324aR102)

appcues/src/main/java/com/appcues/data/model/ExperienceTrigger.kt
```diff
@@ -3,9 +3,17 @@ package com.appcues.data.model
...
+        companion object {
+
+            const val REASON_SCREEN_VIEW = "screen_view"
+        }
+    }
+
     data class ExperienceCompletionAction(val fromExperienceId: UUID?) : ExperienceTrigger()
-    data class LaunchExperienceAction(val fromExperienceId: UUID?) : ExperienceTrigger()
+
+    // val fromExperienceId: UUID was not being used
```
andretortolano, line 15: TODO: go back to this and re-add fromExperienceId to minimize change

PR #481, opened 2023-09-11
🐛 guarding onSizeChange to only apply when size is not the same as old size

andretortolano:
comparing width and height (as Int) to make sure we are not updating size on every decimal size update.

unsure the root cause of the issue, its probably because we may have an unintentional loop in this composition, where we need the containerSize and contentSize to build the tooltipSettings, and tooltipSettings needs to know the content size too.

maybe worth revisiting this later to see if we can simplify how we build tooltips, but for now I think we can limit the update on containerSize to only update when its a relevant value


iujames (approved):
seems like a good check to avoid unnecessary updates - this resolves that one unexpected case we saw on the pixel 5?

andretortolano:
it does, this is freds flow running on the faulty emulator now is properly displaying

image

PR #482, opened 2023-09-11
✨ Sending new property session_id on every ActivityRequest

andretortolano:
As requested [here](https://appcues.slack.com/archives/C04MYUL758B/p1694113053183879). adding new property session_id to every ActivityRequest.

most of the changes impacted unit test since ActivityRequest is an object used in many places, the core change is on AnalyticsTracker and SessionMonitor, where we changed `checkSession: Boolean` to `getSession(): UUID? `
and `startSession(): Boolean` to `startSession(): UUID?`

updated example of response including new session_id:

```json

{
	"request_id": "c592f943-8b38-42b0-a3b5-cdfaa060c56d",
	"profile_update": {
		"_deviceType": "phone",
		"_deviceModel": "Google sdk_gphone64_arm64",
		"_sessionRandomizer": 19,
		"_appVersion": "1.0",
		"_lastContentShownAt": 1694446894789,
		"_appBuild": "1",
		"_sessionId": "d473c99e-e810-4308-98dd-9e7ec2b30f4a",
		"_userAgent": "Dalvik/2.1.0 (Linux; U; Android 13; sdk_gphone64_arm64 Build/TE1A.220922.031)",
		"userId": "default-5555",
		"_operatingSystem": "android",
		"_sdkName": "appcues-android",
		"_osVersion": "33",
		"_appName": "Appcues Example",
		"_lastBrowserLanguage": "en",
		"_appId": "4b6b796f-1313-440a-b356-ee4b7abe931c",
		"_isAnonymous": false,
		"_sessionPageviews": 0,
		"_localId": "6231d407-fb8f-43e2-a177-75e2e4fa28b4",
		"_updatedAt": 1694447738282,
		"_bundlePackageId": "com.appcues.samples.kotlin",
		"_sdkVersion": "3.0.2"
	},
	"user_id": "default-5555",
	"account_id": "103523",
	"session_id": "d473c99e-e810-4308-98dd-9e7ec2b30f4a",
	"group_id": null
}
```

### DIFF
--- a/appcues/src/main/java/com/appcues/analytics/ActivityRequestBuilder.kt
+++ b/appcues/src/main/java/com/appcues/analytics/ActivityRequestBuilder.kt
@@ -5,6 +5,7 @@ import com.appcues.Storage
 import com.appcues.analytics.AnalyticsEvent.ScreenView
 import com.appcues.data.remote.appcues.request.ActivityRequest
 import com.appcues.data.remote.appcues.request.EventRequest
+import java.util.UUID
 
 internal class ActivityRequestBuilder(
     private val config: AppcuesConfig,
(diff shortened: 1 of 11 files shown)


appcues/src/main/java/com/appcues/analytics/ActivityRequestBuilder.kt
```diff
@@ -18,25 +19,27 @@ internal class ActivityRequestBuilder(
         const val SCREEN_TITLE_CONTEXT = "screen_title"
     }
 
-    fun identify(properties: Map<String, Any>? = null) = decorator.decorateIdentify(
+    fun identify(sessionId: UUID, properties: Map<String, Any>? = null) = decorator.decorateIdentify(
```
iujames, line 22: would it work for `ActivityRequestBuilder` to have `SessionMonitor` injected? then it could just get `sessionId` directly rather than having to pass it in at all callsites?

The network object `ActivityRequest` could have this value optional, since technically that value in `SessionMonitor` is optional - but we know at this point it will always be set, due to the other checks that run before activity is tracked.

iujames (commented):
Does it make sense to do this _after_ the other analytics changes? Seems like that will have some conflicts here perhaps.

appcues/src/main/java/com/appcues/analytics/ActivityRequestBuilder.kt
andretortolano: since all the ActivityRequest should have sessionId I think its better if we enforce it with a non-null UUID.

assuming that it means that we should be getting sessionId from a place were we know its never null (AnalyticsTracker), even though this is more verbose and account for more change it explicitly say that session id is a non null required variable we need to pass on, we need to have a session in order to call the activity endpoint

iujames (approved):

PR #483, opened 2023-09-11
Add slideout transition handling to dialog modal

iujames:
This PR adds support for the slideout display style in our dialog modal. The key aspects of this are
* Can have a specific width specified, not always full width
* Can be positioned in any horizontal or vertical alignment combination
* Can include a slide in from edge transition animation (and matching exit)

There were some structural changes in `DialogModal` to support this, and I'll try to call those out specifically. But there should not be any visual difference in a standard dialog modal using the existing "dialog" style and the default fade transition. Will confirm this with UI tests as well.

https://github.com/appcues/appcues-android-sdk/assets/19266448/12af2426-f45c-44bf-90a2-c8c8b102eb71

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
+++ b/appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
@@ -28,11 +28,13 @@ import com.appcues.trait.AppcuesContentAnimatedVisibility
 import com.appcues.ui.extensions.getBoxAlignment
 import com.appcues.ui.extensions.getPaddings
 import com.appcues.ui.extensions.modalStyle
-import com.appcues.ui.modal.TransitionEdge.BOTTOM
-import com.appcues.ui.modal.TransitionEdge.CENTER
-import com.appcues.ui.modal.TransitionEdge.LEADING
-import com.appcues.ui.modal.TransitionEdge.TOP
-import com.appcues.ui.modal.TransitionEdge.TRAILING
+import com.appcues.ui.modal.DialogTransition.FADE
+import com.appcues.ui.modal.DialogTransition.SLIDE
+import com.appcues.ui.modal.SlideTransitionEdge.BOTTOM
+import com.appcues.ui.modal.SlideTransitionEdge.CENTER
+import com.appcues.ui.modal.SlideTransitionEdge.LEADING
+import com.appcues.ui.modal.SlideTransitionEdge.TOP
+import com.appcues.ui.modal.SlideTransitionEdge.TRAILING
 import com.appcues.ui.utils.AppcuesWindowInfo
 import com.appcues.ui.utils.AppcuesWindowInfo.ScreenType.COMPACT
 import com.appcues.ui.utils.AppcuesWindowInfo.ScreenType.EXPANDED
@@ -46,15 +48,11 @@ private val MAX_HEIGHT_MEDIUM_DP = 800.dp
 private val MAX_HEIGHT_EXPANDED_DP = 900.dp
 private const val SCREEN_PADDING = 0.05
 
-private val SCREEN_VERTICAL_PADDING = 24.dp
+internal enum class DialogTransition {
+    FADE, SLIDE
+}
 
-internal data class DialogTransition(
-    val type: String,
-    val slideInEdge: String?,
-    val slideOutEdge: String?,
-)
-
-internal enum class TransitionEdge {
+internal enum class SlideTransitionEdge {
     LEADING, TRAILING, TOP, BOTTOM, CENTER
 }
 
@@ -71,18 +69,33 @@ internal fun DialogModal(
     val maxWidth = maxWidthDerivedOf(windowInfo)
     val maxHeight = maxHeightDerivedOf(windowInfo)
 
-    // horizontal padding is a scale factor based on device width
-    // vertical padding is a fixed constant
-    val configuration = LocalConfiguration.current
-    val dialogHorizontalPadding = (configuration.screenWidthDp * SCREEN_PADDING).dp
+    // fixed with is used for slideout style dialogs
+    val isFixedWidth = (style?.width ?: -1.0) > 0.0
+
+    val dialogHorizontalPadding = if (isFixedWidth) {
+        // fixed value for fixed width slideouts
+        12.dp
+    } else {
+        // normal dialog modals use horizontal padding wih a scale factor based on device width
+        val configuration = LocalConfiguration.current
+        (configuration.screenWidthDp * SCREEN_PADDING).dp
+    }
+
+    val dialogVerticalPadding = if (isFixedWidth) {
+        // fixed value for fixed width slideouts
+        12.dp
+    } else {
+        // normal dialog modals use standard 24.dp vertical padding
+        24.dp
+    }
 
     val horizontalPaddingPx = with(density) { dialogHorizontalPadding.roundToPx() }
-    val verticalPaddingPx = with(density) { SCREEN_VERTICAL_PADDING.roundToPx() }
+    val verticalPaddingPx = with(density) { dialogVerticalPadding.roundToPx() }
 
     Box(
         modifier = Modifier
             .fillMaxSize()
-            .padding(horizontal = dialogHorizontalPadding, vertical = SCREEN_VERTICAL_PADDING)
+            .padding(horizontal = dialogHorizontalPadding, vertical = dialogVerticalPadding)
     ) {
         AppcuesContentAnimatedVisibility(
             modifier = Modifier.align(style.getBoxAlignment()),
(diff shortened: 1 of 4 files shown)


appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
```diff
@@ -33,48 +43,119 @@ private val MAX_WIDTH_EXPANDED_DP = 560.dp
...
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(horizontal = SCREEN_HORIZONTAL_PADDING, vertical = SCREEN_VERTICAL_PADDING)
     ) {
-        Box(modifier = Modifier.fillMaxSize()) {
+        AppcuesContentAnimatedVisibility(
+            modifier = Modifier.align(style.getBoxAlignment()),
+            enter = transition.toEnterTransition(style, horizontalPaddingPx, verticalPaddingPx),
+            exit = transition.toExitTransition(style, horizontalPaddingPx, verticalPaddingPx),
+        ) {
```
iujames, line 85: The animation needs to go here - inside the `Box`. This is because we don't want to animate the entire box (mostly whitespace around the dialog), but rather, the modal contents. This doesn't matter much on `fade` animations, but on `slide` animations its the difference between sliding the entire screen in from an edge (undesired) versus sliding just the content box in from the edge (desired)

appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
```diff
@@ -33,48 +43,119 @@ private val MAX_WIDTH_EXPANDED_DP = 560.dp
 private val MAX_HEIGHT_COMPACT_DP = Dp.Unspecified
 private val MAX_HEIGHT_MEDIUM_DP = 800.dp
 private val MAX_HEIGHT_EXPANDED_DP = 900.dp
-private const val SCREEN_PADDING = 0.05
+
+private val SCREEN_HORIZONTAL_PADDING = 12.dp
+private val SCREEN_VERTICAL_PADDING = 24.dp
```
iujames, line 48: This is up for debate, but in tooltips we have this standard padding defined, and in modals we had a screen % padding factor being applied. This seemed inconsistent to me, and on the vertical padding in particular it could lead to slideouts seeming to be a fair bit further away from the top or bottom edge than expected. This also would affect top or bottom positioned dialog modals, which similarly seemed a bit too far from the edge, IMO. It is a matter of visual design though, and open to whatever product and design think is best if there are concerns. It felt best to be consistent with tooltips.

appcues/src/main/java/com/appcues/ui/modal/SharedModal.kt
```diff
@@ -31,6 +40,42 @@ internal fun dialogExitTransition(): ExitTransition {
...
+    }
+}
+
+internal fun slideOutExitTransition(target: TransitionEdge, horizontalPadding: Int, verticalPadding: Int): ExitTransition {
+    val slide = slideOut(
+        targetOffset = { target.toOffset(it, horizontalPadding, verticalPadding) },
+        animationSpec = tween(durationMillis = 300)
+    )
+
+    return if (target == CENTER) {
```
iujames, line 62: this is the special case "edge" `CENTER` - used when the slideout is positioned at center+center rather than truly on an edge. In this case, we do a fade in/out as well as a transition up by half the content height.

appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
```diff
@@ -33,48 +43,119 @@ private val MAX_WIDTH_EXPANDED_DP = 560.dp
...
-    AppcuesContentAnimatedVisibility(
-        enter = dialogEnterTransition(),
-        exit = dialogExitTransition(),
+    val horizontalPaddingPx = with(density) { SCREEN_HORIZONTAL_PADDING.roundToPx() }
+    val verticalPaddingPx = with(density) { SCREEN_VERTICAL_PADDING.roundToPx() }
+
+    Box(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(horizontal = SCREEN_HORIZONTAL_PADDING, vertical = SCREEN_VERTICAL_PADDING)
```
iujames, line 79: also an important note - the padding is being applied here, at the root of the composition, outside of the animation block, rather than inside on the surface as before (L63 in older code). It seemed like this was required to give the animation the proper area to transition things. It might have also been possible before to have a very large shadow clipped off by the container box from what I was seeing? Need to double check that, but I think this should be safe.

appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
andretortolano: SCREEN_PADDING was one of first things we did here (this is old code), it is ok to change from that to dp paddings, only question regarding tablet and landscape, which i recall was the reason why I went with % in the first place, but I guess since then we did specific layout for those cases as well. 
it should be well covered.

appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
```diff
@@ -33,48 +43,119 @@ private val MAX_WIDTH_EXPANDED_DP = 560.dp
 private val MAX_HEIGHT_COMPACT_DP = Dp.Unspecified
 private val MAX_HEIGHT_MEDIUM_DP = 800.dp
 private val MAX_HEIGHT_EXPANDED_DP = 900.dp
-private const val SCREEN_PADDING = 0.05
+
+private val SCREEN_HORIZONTAL_PADDING = 12.dp
+private val SCREEN_VERTICAL_PADDING = 24.dp
+
+internal data class DialogTransition(
```
andretortolano, line 50: up to you if it looks more organized this way, but since all information is coming from style you could build this object inside of DialogModal since this object does not matter too much to the other presentation styles.

appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
andretortolano: in any case you need Box to be outside the AnimatedVisibility

appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
iujames: right, I believe those are covered by the max width that gets set, but I will so review those to test - good call.

I did run the UI tests, and things work out fine in general - but there will be several image updates where the dialog gets updated to 12.dp margin left/right instead of something slightly larger than that in the previous code.

appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
iujames: right now, this information is coming in through `ModalTrait`, since it is in the trait config (outside of style) and needs to be passed through into the `DialogModal` composition. So that is why I made this extra data class to capture and pass the info through here. Maybe I am missing another point on simplification though?

appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
andretortolano: im ok with that.

appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
andretortolano: you are right I for some reason thought this was coming from Style.

appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
iujames: I am second guessing this change now and think I'll put it back the way it was. On the 390dp width test device, for example, the padding on the sides is now 12 instead of about 20. It just looks like it's a little to horizontally stretched for a dialog modal. Maybe we could go with fixed vertical padding but keep horizontal as it was before.

andretortolano: do you have a visual example of that?

iujames: here is an example on pixel 6

|  before (and current)  | experimental 12 dp margin |
| ---- | ---- |
| Screenshot 2023-09-12 at 9 29 26 AM | Screenshot 2023-09-12 at 10 30 36 AM |

iujames: Here is the "downside" to that - on a larger screen, a slideout along an edge appears a bit further from that edge than I would expect.  So maybe we should have different rules for normal dialogs (horizontal center) versus items with leading/trailing edge alignment?

Screenshot 2023-09-12 at 10 40 36 AM

andretortolano: maybe its just a matter of figuring out a good value for horizontal margins there, maybe 12 is good for tooltips but not good for modals, how is IOS dealing with this? is this one of those settings coming from the SwiftUI Package?

andretortolano: maybe instead of 12. maybe 16 or 20, 24  may look the same on regular devices, and then should be good for mini modals too.

PR #486, opened 2023-09-14
remove Compose preview functions

iujames:
Proposed change to (A) not include the preview code in our release package and (B) make it easy to exclude these from test coverage work being done.


andretortolano:
one of the benefits of @Preview is that you can live check changes, mostly used for developing time.

 I'm seriously considering removing all @Preview instead of moving away from these files, if we ever need to come back and change a component we can we-do the preview if we need to for developing purposes.
 
 at some point we tried to explore the idea of using the preview function for screenshot testing, that did not work as we would like, and we basically left it out there
 
 what do you think?

iujames:
@andretortolano I would not be opposed to just removing the Previews at this point. Given that we have an entire test library of component JSON and snapshots, that is really serving the purpose of validating all the UI behavior and these Previews do not really serve a meaningful purpose for current development work. I'll go ahead and remove them then and update this PR.

iujames:
@andretortolano updated this PR to just remove the previews

andretortolano (approved):
🚀

andretortolano:
suggest changing the PR title before merging!

PR #491, opened 2023-09-19
Fix reporting of missing render context errors

iujames:
Discovered this bug while testing some other analytics changes in 3.1. We had an issue with how we were reporting experience errors for the "no render context found" case, related to embeds.

1. (less severe) It was not including any of the experience properties like normal lifecycle events
2. (primary issue) it can potentially end up reporting the same error over and over each 10 sec interval as it is flushed and a new (empty) qualify response comes back

The key thing was that we were missing this guard that iOS has: https://github.com/appcues/appcues-ios-sdk/blob/e8308edbf160cd7b3b6bd2e70982a8e08f110f26/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateObserver.swift#L118 -- this ensures that a recoverable error is only tracked once - when it initially gets processed and the error ID is null. If the error ID is not null, that means it was already sent and we should just ignore and wait until if/when it recovers.

We now have the similar check in the function used in AnalyticsTrackerExt here.

### DIFF
--- a/appcues/src/main/java/com/appcues/analytics/AnalyticsTrackerExt.kt
+++ b/appcues/src/main/java/com/appcues/analytics/AnalyticsTrackerExt.kt
@@ -1,7 +1,9 @@
 package com.appcues.analytics
 
+import com.appcues.analytics.ExperienceLifecycleEvent.ExperienceError
 import com.appcues.data.model.Experience
 import com.appcues.data.model.Experiment
+import com.appcues.statemachine.Error
 import com.appcues.util.appcuesFormatted
 import java.util.UUID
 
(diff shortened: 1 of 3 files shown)


appcues/src/main/java/com/appcues/analytics/AnalyticsTrackerExt.kt
```diff
@@ -23,19 +25,19 @@ internal fun AnalyticsTracker.track(experiment: Experiment) {
     )
 }
 
-internal fun AnalyticsTracker.trackExperienceError(experience: Experience, message: String) {
-    val errorId = UUID.randomUUID()
+internal fun AnalyticsTracker.trackRecoverableExperienceError(experience: Experience, message: String) {
+    // only track the recoverable error once
+    if (experience.renderErrorId != null) return
```
iujames, line 30: the key check is here - don't track if it's already been tracked.

without this, it will keep re-tracking every 10 secs as the previous error gets back a qualify response, which then triggers looking through the cache and retrying the embeds, and failing again (loop)

andretortolano (approved):

PR #493, opened 2023-09-19
Add locale name and id in analytics

iujames:
Quick update here to capture a `context` object in the experience response that will contain the locale id and name, in the case of localized content, when this is returned from the server in the future. This info is then returned verbatim in the experience analytics to denote which version of an experience the user saw.

Per Andy, the response property at the root of the experience will look like 
```js
"context": {
    "locale_id": "the-id",
    "locale_name": "the-name"
}
```

and the name and id value is "default" when no localization matched


iujames:
confirmed these values are now coming through from the backend as expected (via early staging testing)

Screenshot 2023-09-20 at 8 50 54 AM

andretortolano (approved):

PR #494, opened 2023-09-20
Fix Debugger UI to LTR layout direction

iujames:
Detail from testing RTL languages like Arabic - our debugger is fixed to English language and should fix itself to LTR layout regardless of the system/app settings.


andretortolano (approved):
simple change, fix to LTR on debugger 👍

PR #495, opened 2023-09-20
Improved trait error handling

iujames:
This update is meant to address two scenarios:
1. A modal trait has a presentationStyle that is unknown
2. There is no valid presenting trait in the experience

Both of these scenarios currently result in sub-optimal behavior:
1. leaves the backdrop overlay on the screen with no content, in a potentially "hung" state
2. throws an exception that is not handled until it hits the root coroutine handler, aborting the entire response processing flow

In both cases, no trackable errors are reported to the backend, and if there are other valid experiences in the response, lower in the priority list, they get skipped instead of attempted.

This change addresses these by (1) supporting traits throwing AppcuesTraitException instances during init, if config is invalid and (2) handling these exceptions in the proper spot to track events and allow fallback experiences to try to render.

Updated handling with errors tracked
| error 1 | error 2 |
| ---- | ---- |
| Screenshot 2023-09-20 at 1 06 37 PM | Screenshot 2023-09-20 at 1 07 13 PM |

### DIFF
--- a/appcues/src/main/java/com/appcues/data/AppcuesRepository.kt
+++ b/appcues/src/main/java/com/appcues/data/AppcuesRepository.kt
@@ -15,7 +15,12 @@ import com.appcues.data.remote.RemoteError
 import com.appcues.data.remote.RemoteError.NetworkError
 import com.appcues.data.remote.appcues.AppcuesRemoteSource
 import com.appcues.data.remote.appcues.request.ActivityRequest
+import com.appcues.data.remote.appcues.response.QualifyResponse
+import com.appcues.data.remote.appcues.response.experience.ExperienceResponse
+import com.appcues.data.remote.appcues.response.experience.FailedExperienceResponse
+import com.appcues.data.remote.appcues.response.experience.LossyExperienceResponse
 import com.appcues.logging.Logcues
+import com.appcues.trait.AppcuesTraitException
 import com.appcues.util.ResultOf
 import com.appcues.util.ResultOf.Failure
 import com.appcues.util.ResultOf.Success
@@ -153,12 +158,11 @@ internal class AppcuesRepository(
             val qualifyResult = appcuesRemoteSource.qualify(activity.userId, activity.userSignature, activity.requestId, activity.data)
 
             qualifyResult.doIfSuccess { response ->
-                val priority: ExperiencePriority = if (response.qualificationReason == "screen_view") LOW else NORMAL
                 val trigger = ExperienceTrigger.Qualification(response.qualificationReason)
                 qualificationResult = QualificationResult(
                     trigger = trigger,
-                    experiences = response.experiences.map {
-                        experienceMapper.mapDecoded(it, trigger, priority, response.experiments, activity.requestId)
+                    experiences = response.experiences.mapNotNull {
+                        mapExperience(activity.requestId, response, trigger, it)
                     }
                 )
             }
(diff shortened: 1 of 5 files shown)


appcues/src/main/java/com/appcues/data/AppcuesRepository.kt
```diff
@@ -208,4 +212,34 @@ internal class AppcuesRepository(
             post(queue, current)
         }
     }
+
+    private fun mapExperience(
```
iujames, line 216: this function was moved out from L161 to it's own function, since it got larger and having all this inline would cause detekt issues as well.

the key aspect is the try/catch now, where it handles AppcuesTraitException during mapping, and returns back the failed experience so that it can properly track the error - the actual error message is inserted on L236 from the exception message

appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
```diff
@@ -150,11 +151,10 @@ internal class ExperienceMapper(
     }
 
     private fun List<ExperienceTrait>.getExperiencePresentingTraitOrThrow(): PresentingTrait {
-        return filterIsInstance<PresentingTrait>().firstOrNull() ?: throw AppcuesPresentingTraitNotFound()
+        return filterIsInstance<PresentingTrait>().firstOrNull()
+            ?: throw AppcuesTraitException("Presenting capability trait required")
```
iujames, line 155: getting rid of `AppcuesPresentingTraitNotFound` and consolidating as just another type of `AppcuesTraitException` here.

appcues/src/main/java/com/appcues/data/mapper/trait/TraitsMapper.kt
```diff
@@ -14,7 +15,16 @@ internal class TraitsMapper(
...
-                    add(factory.invoke(it.first.config, it.second, renderContext))
+                    try {
+                        add(factory.invoke(it.first.config, it.second, renderContext))
+                    } catch (ex: InstanceCreationException) {
+                        // since Traits are loaded through Koin, we catch any issue
+                        // with Koin instance creation here, then hopefully get the underlying
+                        // AppcuesTraitException as the cause and throw that instead. This is
+                        // because higher level mapping code handles the trait exceptions and
+                        // reports proper ExperienceErrors
+                        throw ex.cause ?: ex
```
iujames, line 26: explained in the comments, but this is what supports ModalTrait (or any other) throwing an exception if it's config is invalid

appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
```diff
@@ -77,8 +78,11 @@ internal class StateMachine(
         lifecycleTracker.start(this, { onEndedExperience?.invoke(it) })
     }
 
+    @OptIn(ExperimentalCoroutinesApi::class)
     private fun stopLifecycleTracking() {
         lifecycleTracker.stop()
+        _stateFlow.resetReplayCache()
+        _errorFlow.resetReplayCache()
```
iujames, line 85: this is the semi-related fix, so that when multiple experiences are attempted due to failure/fallback - the subsequent attempts dont re-emit the errors from the previous try, due to the SharedFlow having a retry: 1 (avoiding changing that for now to hopefully not introduce any unexpected regressions elsewhere)

andretortolano (approved):

appcues/src/main/java/com/appcues/trait/appcues/ModalTrait.kt
```diff
@@ -23,17 +28,21 @@ import org.koin.core.scope.Scope
...
     }
 
-    // should we have a default presentation style?
-    private val presentationStyle = config.getConfigOrDefault("presentationStyle", "full")
+    private enum class PresentationStyle {
+        DIALOG, FULL, HALF_SHEET, SHEET
+    }
+
+    // this can throw on init, if invalid presentation style - not one of the known types
+    private val presentationStyle = config.getConfig<String?>("presentationStyle").toPresentationStyle()
```
iujames, line 45: this is the key check now, preventing the error 1 from the description -- it will throw right away (L94) and be caught and handled with error event

PR #498, opened 2023-09-22
✅ AppcuesViewModel Tests

andretortolano:
Was able to put together some valid tests for AppcuesViewModel

when testing state the idea here is that during setup() we observe and put all states in a list.

then when testing we can check the list size and type check elements

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/presentation/AppcuesViewModel.kt
+++ b/appcues/src/main/java/com/appcues/ui/presentation/AppcuesViewModel.kt
@@ -21,15 +21,14 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.launch
-import org.koin.core.component.KoinScopeComponent
-import org.koin.core.component.inject
-import org.koin.core.scope.Scope
 
 internal class AppcuesViewModel(
-    override val scope: Scope,
     private val renderContext: RenderContext,
+    private val coroutineScope: AppcuesCoroutineScope,
+    private val experienceRenderer: ExperienceRenderer,
+    private val actionProcessor: ActionProcessor,
     private val onDismiss: () -> Unit,
-) : ViewModel(), KoinScopeComponent {
+) : ViewModel() {
 
     sealed class UIState {
         object Idle : UIState()
@@ -44,10 +43,6 @@ internal class AppcuesViewModel(
 
         data class Dismissing(val continueAction: () -> Unit) : UIState()
     }
-
-    private val appcuesCoroutineScope by inject<AppcuesCoroutineScope>()
-    private val experienceRenderer by inject<ExperienceRenderer>()
-    private val actionProcessor by inject<ActionProcessor>()
 
     private val _uiState = MutableStateFlow<UIState>(Idle)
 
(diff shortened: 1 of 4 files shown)


appcues/src/test/java/com/appcues/ui/presentation/AppcuesViewModelTest.kt
```diff
@@ -0,0 +1,277 @@
...
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.TestRule
+
+@OptIn(ExperimentalCoroutinesApi::class)
+internal class AppcuesViewModelTest {
+
+    @get:Rule
+    val rule: TestRule = InstantTaskExecutorRule()
```
andretortolano, line 52: testImplementation 'androidx.arch.core:core-testing:2.2.0'

appcues/src/test/java/com/appcues/ui/presentation/AppcuesViewModelTest.kt
```diff
@@ -0,0 +1,277 @@
...
+
+    private lateinit var viewModel: AppcuesViewModel
+
+    private val uiStates = arrayListOf<UIState>()
+
+    private lateinit var stateJob: Job
+
+    @Before
+    fun setup() {
+        Dispatchers.setMain(UnconfinedTestDispatcher())
```
andretortolano, line 75: setting the main dispatcher to UnconfinedTestDispatcher allows us to use viewModelScope from ViewModel() (we do it on init block)

appcues/src/main/java/com/appcues/ui/presentation/AppcuesViewModel.kt
```diff
@@ -21,15 +21,14 @@ import kotlinx.coroutines.flow.MutableStateFlow
...
-import org.koin.core.scope.Scope
 
 internal class AppcuesViewModel(
-    override val scope: Scope,
     private val renderContext: RenderContext,
+    private val coroutineScope: AppcuesCoroutineScope,
+    private val experienceRenderer: ExperienceRenderer,
+    private val actionProcessor: ActionProcessor,
     private val onDismiss: () -> Unit,
-) : ViewModel(), KoinScopeComponent {
```
andretortolano, line 32: changed injection approach to constructor base to simplify test, decouple from the KOIN framework

appcues/src/test/java/com/appcues/ui/presentation/AppcuesViewModelTest.kt
```diff
@@ -0,0 +1,311 @@
...
+            every { groupLookup } returns hashMapOf(3 to 0)
+            every { stepIndexLookup } returns hashMapOf(0 to 1234)
+            every { stepContainers } returns listOf(mockStepContainer)
+        }
+        val isFirst = true
+        val mockMetadata: Map<String, Any?> = hashMapOf()
+        // WHEN (flatStepIndex is 2 here)
+        experienceStates.emit(BeginningStep(mockExperience, 0, isFirst, mockMetadata))
+        // THEN
+        assertThat(uiStates).hasSize(1)
```
iujames, line 137: for my own understanding, where does uiStates get reset between each test?

iujames (approved):
great to test all this since the VM is such a core part of rendering, nice job!

appcues/src/test/java/com/appcues/ui/presentation/AppcuesViewModelTest.kt
andretortolano: yes, each run the list is a new one

appcues/src/test/java/com/appcues/ui/presentation/AppcuesViewModelTest.kt
```diff
@@ -0,0 +1,311 @@
...
+    @Before
+    fun setup() {
+        Dispatchers.setMain(UnconfinedTestDispatcher())
+
+        viewModel = AppcuesViewModel(renderContext, coroutineScope, experienceRenderer, actionProcessor, onDismiss).apply {
+            stateJob = viewModelScope.launch { uiState.collect { uiStates.add(it) } }
+        }
+
+        // clear getStateFlow from history
+        excludeRecords { experienceRenderer.getStateFlow(renderContext) }
```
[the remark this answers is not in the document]
andretortolano: this is excluding the already triggered event from getStateFlow, helps with tests that check for experienceRenderer wasNot called

on each test we have a new experience renderer. before each test run its created an instance of this class and the @Before method is called, followed by the @Test and then the @After, so every run is a new ViewModel nonetheless

PR #499, opened 2023-09-25
♻️  StateMachine transition improvements

andretortolano:
This is a followup PR to ViewModel improvements.

1. AppcuesViewModel will call Dismiss if getStateFlow is null
2. When we receive EndingStep + UiDismissed we cancel the collect job instead of running a if (uiState == Dismissing) during collect
3. calling onUiDismissed from composition (viewModel.onDismissed(onUiDismissed) to avoid checking for dismissing state, since it was already being checked by the composition.
4. ViewPresenter has viewModel as nullable var, so we can unset it whenever onCompositionDismiss
5. Removed unused `LaunchOnShowAnimationCompleted`
6. Turns out we can simplify how AwaitEffect is currently, given that we dont need to pass in AppcuesCoroutineScope (any suspend thread can call complete()) and also we can have it all inside AwaitEffect class, since we dont actually need the result from the await.
7. update to handleInternalAction to support new AwaitEffect more efficiently.

### DIFF
--- a/appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
@@ -90,55 +90,55 @@ internal class StateMachine(
     }
 
     private suspend fun handleActionInternal(action: Action): ResultOf<State, Error> {
-        val transition = _state.take(action)
-        val state = transition.state
-        val sideEffect = transition.sideEffect
+        return _state.take(action).run {
 
-        if (state != null) {
-            // update current state
-            _state = state
+            if (state != null) {
+                // transition to new state
+                _state = state
 
-            if (transition.emitStateChange) {
                 // emit state change to all listeners via flow
-                _stateFlow.emit(state)
+                if (emitStateChange) _stateFlow.emit(state)
+            }
+
+            sideEffect.launch().also {
+                if (_state == Idling) {
+                    stopLifecycleTracking()
+                }
             }
         }
+    }
 
-        return if (sideEffect != null) {
-            when (sideEffect) {
-                is ContinuationEffect -> {
-                    // recursive call on continuations to get the final return value
-                    handleActionInternal(sideEffect.action)
-                }
-
-                is ReportErrorEffect -> {
-                    _errorFlow.emit(sideEffect.error)
-                    // return a failure if this call to `handleAction` ended with a reported error
-                    Failure(sideEffect.error)
-                }
-
-                is PresentContainerEffect -> {
-                    // this will execute the presentation, asynchronously, and return
-                    // the subsequent action to take - either continue to Rendering, or report Error
-                    handleActionInternal(sideEffect.presentContainer(actionProcessor))
-                }
-
-                is AwaitEffect -> {
-                    sideEffect.completion.await()
-                }
-
-                is ProcessActions -> {
-                    actionProcessor.process(sideEffect.actions)
-                    Success(_state)
-                }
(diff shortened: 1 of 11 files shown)


appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
```diff
@@ -66,8 +66,14 @@ internal class StateMachine(
         appcuesCoroutineScope.launch {
             stateFlow.collect {
                 when (it) {
-                    is BeginningExperience -> config.experienceListener?.experienceStarted(it.experience.id)
-                    is EndingExperience -> config.experienceListener?.experienceFinished(it.experience.id)
+                    is BeginningExperience -> {
+                        startLifecycleTracking()
```
andretortolano, line 70: wondering if instead of start/stop lifecycle tracking on transitions if we can do it here when we collect state.

appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
```diff
@@ -90,59 +96,55 @@ internal class StateMachine(
...
-        }.also {
-            if (_state == Idling) {
-                stopLifecycleTracking()
-            }
+            else -> Success(_state)
         }
     }
 
-    private fun State.take(action: Action): Transition {
+    private fun State.takeTransition(action: Action): Transition {
```
iujames, line 147: I prefer the previous naming, as the `State` is taking the `Action`, and producing the `Transition`

appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
```diff
@@ -90,59 +96,55 @@ internal class StateMachine(
...
-            // update current state
-            _state = state
+            if (state != null) {
+                // transition to new state
+                _state = state
 
-            if (transition.emitStateChange) {
                 // emit state change to all listeners via flow
-                _stateFlow.emit(state)
+                if (emitStateChange) _stateFlow.emit(state)
```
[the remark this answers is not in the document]
andretortolano: getState could get the last state from the shared flow ( can dig into that) and then we have only one source of truth which is the SharedFlow.

yes. I saw that we either fail and return Idling or we do null so that state is unchanged, but emiting the same state again does nothing since SharedFlow distinguishUntilChanged by nature, (surely need testing too)

appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
andretortolano: its a valid concern, but come to think of it now, maybe we can centralize somehow. will keep this in mind for possible next PR

appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
andretortolano: forgot to take out before pushing it

appcues/src/main/java/com/appcues/ui/presentation/ViewPresenter.kt
```diff
@@ -104,15 +103,12 @@ internal abstract class ViewPresenter(
 
         gestureListener?.clearListener()
         gestureListener = null
+        viewModel = null
```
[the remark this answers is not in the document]
andretortolano: yeah, I feel like its better if we clear its instance when we get out of the experience.

iujames:
@andretortolano nice work! this looks all good to go right? Want to merge this one after we do 3.1.2 as well?

iujames (approved):
nice simplifications!

PR #501, opened 2023-09-26
Add support for experience test snapshots

iujames:
These are the changes in debug only test helper to support experience snapshot tests in https://github.com/appcues/appcues-mobile-experience-spec/pull/223

plus one error handling case with WebViews, since accessing WebView.settings [is not supported in Paparazzi tests](https://github.com/cashapp/paparazzi/issues/1058)

### DIFF
--- a/appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
+++ b/appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
@@ -6,6 +6,7 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import coil.ImageLoader
@@ -16,6 +17,7 @@ import com.appcues.data.mapper.experience.ExperienceMapper
 import com.appcues.data.mapper.step.primitives.mapPrimitive
 import com.appcues.data.model.ExperienceStepFormState
 import com.appcues.data.model.ExperienceTrigger
+import com.appcues.data.remote.appcues.response.QualifyResponse
 import com.appcues.data.remote.appcues.response.experience.ExperienceResponse
 import com.appcues.data.remote.appcues.response.step.primitive.PrimitiveResponse
 import com.appcues.data.remote.appcues.response.step.primitive.PrimitiveResponse.StackPrimitiveResponse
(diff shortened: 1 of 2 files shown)


appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
```diff
@@ -125,6 +127,76 @@ public fun ComposeContainer(context: Context, stepContentJson: List<String>?, tr
     }
 }
 
+// This helper supports testing any experience JSON content, rendering the given group and step index.
+@Composable
+public fun ComposeContainer(
```
iujames, line 132: this is _mostly_ like the trait version, but it handles some special cases for experience tests

appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
```diff
@@ -125,6 +127,76 @@ public fun ComposeContainer(context: Context, stepContentJson: List<String>?, tr
...
+            stepContainerResponse.copy(
+                children = stepContainerResponse.children.map { stepResponse ->
+                    stepResponse.copy(
+                        actions = emptyMap()
+                    )
+                },
+                actions = emptyMap()
+            )
+        },
+    )
```
iujames, line 160: this is the simplest way to just not worry about Actions in the mapping, since they often refer to internal services and end up trying to resolve dependencies that can rely on the Android context, and that is not available in snapshot tests

appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
```diff
@@ -125,6 +127,76 @@ public fun ComposeContainer(context: Context, stepContentJson: List<String>?, tr
...
+                ComposeContainer(container, stepIndex, metadata)
+            }
+
+            if (animated) {
+                val compositionState = LocalExperienceCompositionState.current
+                LaunchedEffect(Unit) {
+                    compositionState.isContentVisible.targetState = true
+                    compositionState.isBackdropVisible.targetState = true
+                }
+            }
```
iujames, line 195: this allows us to run some tests with full normal animation support, if needed - some experience JSON test do need this (not many)

appcues/src/main/java/com/appcues/ui/primitive/EmbedHtmlPrimitive.kt
```diff
@@ -73,9 +73,16 @@ internal fun EmbedHtmlPrimitive.Compose(modifier: Modifier) {
...
+                    @Suppress("SwallowedException")
+                    try {
+                        with(settings) {
+                            javaScriptEnabled = true
+                            mediaPlaybackRequiresUserGesture = false
+                        }
+                    } catch (ex: NoSuchMethodError) {
+                        // in some UI tests, WebView.settings throws a
+                        // NoSuchMethodError, so we're ignoring here for tests. This should
+                        // not impact normal usage at all.
```
iujames, line 85: this is to support using WebView in snapshot compositions where it is not really supported in general

andretortolano (approved):
Nice way to get experience test snapshots working

PR #502, opened 2023-09-26
🐛 Text primitive not resizing properly

andretortolano:
After updating dependency compose 1.5.1 it seems that an existing behavior of Basic Text changed, this led to a bug where our TextPrimitive was not getting resized if it was reusing a Text node from previous composition.

To fix it we ensure that we are never to reuse composition nodes for TextPrimitives.


iujames (commented):
this seems a fair enough workaround for this issue, but I'd like to dig a little bit more into the root cause and understand for sure if this is an issue in Compose (it seems like it!) and see if there is any other option to avoid extra compositions

iujames:
propose closing this if we use the approach in #504 with the `key(..) {... }` composable

PR #504, opened 2023-09-28
three layout fixes

iujames:
three commits here for three specific fixes that came up recently

1. using `key(id) { ... }` around each step composition, to prevent node re-use causing an issue with our usage of IntrinsicSize.Min on row heights - the original issue in the linked ticket
2. fix the resizing logic on our `text` primitives to resize the span font sizes, not just the root style font size, since now all text content is actually within the spans
3. fix the vertical scroll handling for each individual step - moving the `rememberScrollState()` into the step composition, rather than the outer container composition, to restore expected behavior

### DIFF
--- a/appcues/src/main/java/com/appcues/trait/ContentWrappingTrait.kt
+++ b/appcues/src/main/java/com/appcues/trait/ContentWrappingTrait.kt
@@ -22,6 +22,11 @@ internal interface ContentWrappingTrait : ExperienceTrait {
      */
     @Composable
     fun WrapContent(
-        content: @Composable (modifier: Modifier, containerPadding: PaddingValues, safeAreaInsets: PaddingValues) -> Unit
+        content: @Composable (
+            modifier: Modifier,
+            containerPadding: PaddingValues,
+            safeAreaInsets: PaddingValues,
+            hasVerticalScroll: Boolean,
+        ) -> Unit
     )
 }
--- a/appcues/src/main/java/com/appcues/ui/composables/StepComposition.kt
+++ b/appcues/src/main/java/com/appcues/ui/composables/StepComposition.kt
@@ -5,12 +5,15 @@ import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.verticalScroll
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.key
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.composed
 import androidx.compose.ui.platform.LocalDensity
 import com.appcues.data.model.Step
 import com.appcues.trait.StepDecoratingTrait.StepDecoratingType
@@ -23,7 +26,8 @@ internal fun Step.ComposeStep(
     modifier: Modifier = Modifier,
     containerPadding: PaddingValues,
     safeAreaInsets: PaddingValues,
-    parent: BoxScope
+    parent: BoxScope,
+    hasVerticalScroll: Boolean,
 ) {
     key(id) {
         CompositionLocalProvider(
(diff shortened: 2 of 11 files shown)


appcues/src/main/java/com/appcues/trait/ContentWrappingTrait.kt
```diff
@@ -22,6 +22,11 @@ internal interface ContentWrappingTrait : ExperienceTrait {
      */
     @Composable
     fun WrapContent(
-        content: @Composable (modifier: Modifier, containerPadding: PaddingValues, safeAreaInsets: PaddingValues) -> Unit
+        content: @Composable (
+            modifier: Modifier,
+            containerPadding: PaddingValues,
+            safeAreaInsets: PaddingValues,
+            hasVerticalScroll: Boolean,
```
iujames, line 29: The only change here is the new `hasVerticalScroll` flag, which is passed in from the wrapping trait (modal, tooltip, embed) and propagated all the way through into the StepComposition now. That is the spot where `rememberScrollState()` is now called, giving each step its proper scroll state.

appcues/src/main/java/com/appcues/ui/composables/StepComposition.kt
```diff
@@ -22,23 +26,26 @@ internal fun Step.ComposeStep(
...
+    hasVerticalScroll: Boolean,
 ) {
-    CompositionLocalProvider(
-        LocalAppcuesActions provides actions,
-        LocalExperienceStepFormStateDelegate provides formState
-    ) {
-        // used to get the padding values from step decorating trait and apply to the Column
-        val density = LocalDensity.current
-        val stickyContentPadding = remember(this) { StickyContentPadding(density) }
+    key(id) {
```
iujames, line 32: this line is the fix for the item in the ticket - preventing the compose node re-use between steps that could lead to unexpected text sizing (due to the row using IntrinsicSize.Min for height)

appcues/src/main/java/com/appcues/ui/primitive/TextPrimitive.kt
```diff
@@ -26,27 +27,35 @@ internal fun TextPrimitive.Compose(modifier: Modifier) {
...
-        text = spans.toAnnotatedString(context, isDark),
-        style = resizedStyle,
-        overflow = TextOverflow.Ellipsis,
-        onTextLayout = { textLayoutResult ->
-            // this will attempt to auto scale down single line / char text, handling cases
-            // like emoji that are sized too large to fit in side by side layouts
-            if (textLayoutResult.lineCount == 1) {
-                if (textLayoutResult.isLineEllipsized(0)) {
-                    resizedStyle = resizedStyle.copy(fontSize = resizedStyle.fontSize * TEXT_SCALE_REDUCTION_INTERVAL)
+    key(resizedStyle) {
```
iujames, line 33: this `key` usage I'm not sure about.

The issue here is that IF the text actually does get resized down (L47-51 below), then the outer row size also needs to resize down to the new height. Since our Row has already set it's height to IntrinsicSize.Min during the initial layout, the row stays the height of the original font size and ends up causing padding to show visually above and below the actual Text content.

Somehow, we need to force the row to reset this intrinsic height whenever this downscale happens, and the `key` here will do so. However, it is also probably causing unnecessary usage for the vast majority of cases where no resizing is happening. It does work, functionally, tested locally and snapshot tests.

iujames: here is a visual example of the issue, with the item on the left showing what happens without this key - as the row is not resized down to match the new text size. The item on the right is with this fix.

| before | after |
| --- | --- | 
| Screenshot 2023-09-28 at 11 23 14 AM | Screenshot 2023-09-28 at 11 24 14 AM |

andretortolano (approved):
Ok with all theses changes, probably pending on revisiting TextPrimitive later

appcues/src/main/java/com/appcues/ui/primitive/TextPrimitive.kt
iujames: it does cause the text to redraw, yes, but the problem is that the row has already set its height to IntrinsicSize.Min, using the initial intrinsic size of the text from the original font size, which was larger. As you scale own the font, that intrinsic size also scales down, but the Compose layout system does not update it - it is only set once at the initial render. It is similar to the "deleting blue items" use case described here https://medium.com/@lepicekmichal/compose-intrinsic-redraw-bug-2cdecef0f96c

PR #505, opened 2023-09-28
Add support for custom font asset path in config

iujames:
This is to support custom font loading in React Native Expo projects, on Android. The new config option for `fontAssetPath` will allow specifying something other than the default (`/assets/fonts`), since Expo apps put custom fonts in the `/assets` root.

* update the Debugger list of fonts to look in both the default `/assets/fonts` and any custom path provided
* update the runtime styling for text to look for custom fonts in both the default and any custom path

The majority of the changes here are propagating a new CompositionLocal for AppcuesConfig down through the spots that eventually need it for text styling.

### DIFF
--- a/appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
+++ b/appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
@@ -28,6 +28,7 @@ import com.appcues.ui.composables.AppcuesActionsDelegate
 import com.appcues.ui.composables.ComposeContainer
 import com.appcues.ui.composables.ExperienceCompositionState
 import com.appcues.ui.composables.LocalAppcuesActionDelegate
+import com.appcues.ui.composables.LocalAppcuesConfig
 import com.appcues.ui.composables.LocalExperienceCompositionState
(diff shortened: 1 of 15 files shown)


appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
```diff
@@ -47,6 +48,7 @@ public fun ComposeContent(json: String, imageLoader: ImageLoader) {
         CompositionLocalProvider(
             LocalImageLoader provides imageLoader,
             LocalLogcues provides Logcues(LoggingLevel.DEBUG),
+            LocalAppcuesConfig provides AppcuesConfig("test-account", "test-app"),
```
iujames, line 51: The updates in this file ensure a non-null AppcuesConfig is available in the compositions used in snapshot testing. There really is no use case right now for passing these values in, so I just used a default config with hardcoded test account/app (which is not used in the UI layer). If we ever needed to pass in custom config from the snapshot tests, we could update these functions to take the config as a param from tests.

andretortolano (approved):

PR #506, opened 2023-09-29
Revert dependency updates for React Native compatibility

iujames:
This change allows us to use Kotlin 1.8.22, which is compatible with React Native.

After discovering that React Native likely won't support Kotlin 1.9 fully until 0.73 (upcoming in ~1 month?), this downgrade will put us back in a more compatible state for all possible use cases. This means we cannot use a Compose compiler > 1.4.8. https://developer.android.com/jetpack/androidx/releases/compose-kotlin#pre-release_kotlin_compatibility -- however, we can remain on latest Compose, now 1.5.2.


andretortolano (approved):

appcues/build.gradle
```diff
@@ -12,9 +12,9 @@ apply from: 'codecov.gradle'
 apply from: 'dokka.gradle'
 
 ext.room_version = '2.5.2'
-ext.compose_version = '1.5.1'
-ext.compose_compiler_version = '1.5.3'
-ext.koin_version = '3.5.0'
+ext.compose_version = '1.5.2'
+ext.compose_compiler_version = '1.4.8'
+ext.koin_version = '3.4.3'
```
iujames, line 17: note: Koin 3.5.0 depends on Kotlin 1.9 so that cannot be updated at this time either

PR #507, opened 2023-09-30
♻️ StateMachine refactor and cleanup

andretortolano:
Revisiting our experience state machine structure after a long time to see if there is some room for improvements, also trying to figure out testable pieces (unit test), the focus of this change is around deleting the Transitions and changing the StateMachine implementation so that:

1. State is now an interface, each State implementation needs to handle actions, returning null means that this action is not valid for this state 
```kotlin
internal interface State {
    fun take(action: Action): Transition?
}
```

2. SideEffect is now an interface, and instead of returning ResultOf<State, Error> sideEffects can produce a nextAction, returning non-null will re-invoke handleActionInternal(newAction) which is basically what was happening before but it is now simpler.
```kotlin
internal interface SideEffect {
    suspend fun launch(processor: ActionProcessor): Action?
}
```

so for instance ContinuationEffect(action) returns  action, while AwaitEffect(action) returns the action after await is satisfied. 

3. Transition takes in optional `error: Error?`
```kotlin
internal class Transition(
    val newState: State,
    val error: Error?,
    val sideEffect: SideEffect?
)
```

5. StateMachine [handleInternal](https://github.com/appcues/appcues-android-sdk/pull/507/files#diff-ca58c45f7b044e398e74d1cb446616170d1e5a4b98736272e4870a55fda2bcb9R43) now runs the following
    * take action into transition
    * update current state/emit
    * check if we have any error, if so we emit and return Failure(error)
    * check if we have any sideEffect, if so we apply and re-invoke handleInternalAction(next), else we Success(state)

With these core changes we have some other small changes too, the most notable one is that ViewModel now listen to [RenderingStep instead of BeginningStep](https://github.com/appcues/appcues-android-sdk/pull/507/files#diff-b9abd8b6c7de410a474fba095627a7cd3eb1a3b812ecba70f42822560eafaf7aR65) this is because instead of relying on State extension functions to produce the metadata we are not moving to `BeginningStep` with [PresentationEffect](https://github.com/appcues/appcues-android-sdk/pull/507/files#diff-81e73d2e2473b16a3ba86867cc76ec415c9e7664ecc319333c11752929411c6cR15) that does all the work and produce the `RenderStep` action with valid metadata or else it ReportError

Listening to BeginingStep was important before because we used to callback a "rendering completion" which is no longer done, by the time the ViewModel is collecting state we assume the experience is showing already, so there is not practical difference to this change aside from better organizing the state machine flow.

```
BeginningExperienceState
|--> next(RenderingStepState, PresentationEffect)
    |--> produceMetadataWithRetry()
    |--> presentingTrait.present()
    |--> return action RenderStep(metadata)
```
                  
```
EndingStepState        
|--> next(RenderingStepState(index), PresentationEffect(index))
    |--> produceMetadataWithRetry() or produceMetadata()
    |--> presentingTrait.present()
    |--> return action RenderStep(metadata)         
``` 

other changes:
1. change: `StepReference` package to model, since its not a state machine specific class
2. change: `config.experienceListener` calls to ExperienceLifecycleTracker to avoid creating extra collects + taking responsibility from state machine
6. add: `getNavigationActions`/`areStepsFromDifferentGroup`/`getPresentingTrait`/ `isValidStepIndex` to Experience class
7. change: `RenderStep` action now carries metadata to ViewModel
8. delete: `BeginningStepExt` and moved all methods to other places (State, SideEffect)
9. Added: Bunch of unit tests!


appcues/src/main/java/com/appcues/util/ResultOf.kt
```diff
@@ -5,13 +5,13 @@ internal sealed class ResultOf<out Success, out Failure> {
     data class Failure<out Failure>(val reason: Failure) : ResultOf<Nothing, Failure>()
 }
 
-internal inline fun <reified Success, reified Failure> ResultOf<Success, Failure>.doIfSuccess(callback: (value: Success) -> Unit) {
+internal inline fun <reified Success, reified Failure> ResultOf<Success, Failure>.doIfSuccess(callback: (value: Success) -> Unit) = apply {
```
andretortolano, line 8: revert changes.

appcues/src/main/java/com/appcues/data/model/Experience.kt
```diff
@@ -48,4 +54,34 @@ internal data class Experience(
...
+    @OptIn(ExperimentalContracts::class)
+    fun isValidStepIndex(flatStepIndex: Int?): Boolean {
+        contract {
+            returns(true) implies (flatStepIndex != null)
+        }
+
+        return flatStepIndex != null && flatStepIndex >= 0 && flatStepIndex < flatSteps.size
+    }
+
+    fun getMetadataSettingTraits(flatStepIndex: Int): List<MetadataSettingTrait> {
```
andretortolano, line 83: before it this is what we were basically doing to get the step trait, am i missing something here? we can just get the step on the steplist by flatStepIndex i believe

```kotlin
private fun BeginningStep.metadataSettingsTraits(): List<MetadataSettingTrait> {
    return with(experience) {
        // find the container index
        val containerId = groupLookup[flatStepIndex]
        // find the step index in relation to the container
        val stepIndexInContainer = stepIndexLookup[flatStepIndex]
        // if both are valid ids we return Rendering else null
        if (containerId != null && stepIndexInContainer != null) {
            val container = stepContainers[containerId]
            val step = container.steps[stepIndexInContainer]
            // this will throw if metadata fails
            step.metadataSettingTraits
        } else {
            // should be impossible at this point, but will cover with an exception as well
            throw AppcuesTraitException("Invalid step index $flatStepIndex")
        }
    }
}
```

appcues/src/main/java/com/appcues/data/model/StepContainer.kt
```diff
@@ -12,4 +12,9 @@ internal data class StepContainer(
     val presentingTrait: PresentingTrait,
     val contentHolderTrait: ContentHolderTrait,
     val contentWrappingTrait: ContentWrappingTrait,
-)
+) {
+
+    init {
+        steps.forEach { it.parent = this }
```
andretortolano, line 18: setting "parent" reference on each step makes it easier to reference its container when getting presenting trait.

another idea is to actually pass in the presentingTrait to all children, but setting the parent reference might be useful for other things in the future if we need.

appcues/src/main/java/com/appcues/data/model/Experience.kt
```diff
@@ -48,4 +54,37 @@ internal data class Experience(
...
+    fun getNavigationActions(stepContainerIndex: Int): List<ExperienceAction> {
+        val stepGroup = stepContainers.getOrNull(stepContainerIndex) ?: return arrayListOf()
+        return stepGroup.actions[stepGroup.id]?.filter { it.on == NAVIGATE }?.map { it.experienceAction } ?: emptyList()
+    }
+
+    fun areStepsFromDifferentGroup(stepIndexOne: Int, stepIndexTwo: Int): Boolean {
+        return groupLookup[stepIndexOne] != groupLookup[stepIndexTwo]
+    }
+
+    fun getPresentingTrait(flatStepIndex: Int): PresentingTrait {
```
andretortolano, line 69: the old implementation had a fundamental problem, it is not fully testable since we know internally both lists are coupled together, but whenever accessing a list (stepContainers) we need to test what would happen if we get IndexOutOfBounds, or replace it by getOrNull to ensure all cases are covered.

so changing this code seems logical as we should avoid the possibility of untestable state in general

```kotlin
internal fun BeginningStep.presentingTrait(): PresentingTrait? {
    return experience.groupLookup[flatStepIndex]?.let { experience.stepContainers[it].presentingTrait }
}
```

appcues/src/main/java/com/appcues/data/model/Experience.kt
iujames: Yeah, I think this old code got ported over from somewhere else, and probably not simplified - but I agree, just using the `flatSteps` and the `flatStepIndex` seems like it should work

appcues/src/main/java/com/appcues/data/model/Experience.kt
iujames: previously, did we have error handling for the null return value for `PresentingTrait?` and throwing an AppcuesTraitException somewhere? Want to make sure we don't change the error handling behavior, as this is now possibly throwing a different exception type.

appcues/src/main/java/com/appcues/data/model/StepContainer.kt
andretortolano: scratch that, its so much easier to have presenting trait in the step, also leaving an lateinit var variable not tested from within the class is a bad idea.

appcues/src/main/java/com/appcues/data/model/Experience.kt
andretortolano: reusing bit of code since the same exception is used inside Experience

andretortolano: actually took another approach here were we dont need to change AppcuesTraitException to "open"

andretortolano:
After I was done with implementing a the new structure, I followed up with creating tests for each state and sideEffect that we have, and that led me into changing some of the existing code, during every turn I kept checking existing tests and accommodating it accordingly.

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -83,20 +88,20 @@ internal class ExperienceLifecycleTracker(
 
     private fun State.track() {
         when (this) {
-            is RenderingStep -> {
-                if (isFirst) {
```
andretortolano, line 87: BeginningExperience and EndExperience are now working properly as "ExperienceStarted" and "ExperienceEnded"

appcues/src/main/java/com/appcues/data/model/Step.kt
```diff
@@ -3,12 +3,14 @@ package com.appcues.data.model
...
 import com.appcues.trait.ContainerDecoratingTrait
 import com.appcues.trait.MetadataSettingTrait
+import com.appcues.trait.PresentingTrait
 import com.appcues.trait.StepDecoratingTrait
 import java.util.UUID
 
 internal data class Step(
     val id: UUID,
     val content: ExperiencePrimitive,
+    val presentingTrait: PresentingTrait,
```
andretortolano, line 13: every step contains presentingTrait to make it easier to reference it later (same way we have contaienrDecoratingTrait here)

appcues/src/main/java/com/appcues/statemachine/Action.kt
```diff
@@ -1,16 +1,19 @@
 package com.appcues.statemachine
 
 import com.appcues.data.model.Experience
+import com.appcues.data.model.StepReference
 
 internal sealed class Action {
     data class StartExperience(val experience: Experience) : Action()
-    data class StartStep(val stepReference: StepReference) : Action()
-    object RenderStep : Action()
+    data class MoveToStep(val stepReference: StepReference) : Action()
```
andretortolano, line 8: added this MoveToStep action to Separate concern from StartStep, which was being used by both RenderingStepState and EndingStepState.

now RenderingStepState takes MoveToStep and decides if it should go to EndingStepState(StartStep) this way we can pass in more valid information thru StartStep like nextFlatStepIndex and nextStepContainerIndex, both of which needs to be valid when we get there. this way we
 1. avoid getting locked into EndingStepState with possibly an invalid container step (which should never happen)
 2. if any error when getting those values, we can keep the current RenderingStepState and log the error, meaning user will continue to see the experience and can find another way to get out of it (which should also never happen but we safe check)

appcues/src/main/java/com/appcues/statemachine/Action.kt
```diff
@@ -1,16 +1,19 @@
...
+    data class StartStep(val nextFlatStepIndex: Int, val nextStepContainerIndex: Int) : Action()
+    data class RenderStep(val metadata: Map<String, Any?>) : Action()
```
andretortolano, line 10: RenderStep has the produced metadata, this gives us better control of the states, instead of doing an elaborate work during BeginningStepState we can get to Beginning and invoke PresentationEffect, which will do all of this for us and return RenderStep(metadata), isolating the produceMetada, PresentingTrait.present() and retries

appcues/src/main/java/com/appcues/statemachine/Error.kt
```diff
@@ -10,8 +9,6 @@ internal sealed class Error(open val message: String, errorId: UUID? = null) {
 
     object ExperienceAlreadyActive : Error("Experience already active")
 
-    data class RenderContextNotActive(val renderContext: RenderContext) : Error("RenderContext $renderContext is not active.")
```
andretortolano, line 13: this error is was being "used" by ExperienceRenderer, but not really used.

appcues/src/main/java/com/appcues/statemachine/effects/AwaitDismissEffect.kt
```diff
@@ -0,0 +1,21 @@
+package com.appcues.statemachine.effects
+
+import com.appcues.action.ActionProcessor
+import com.appcues.statemachine.Action
+import com.appcues.statemachine.SideEffect
+import kotlinx.coroutines.CompletableDeferred
+import org.jetbrains.annotations.VisibleForTesting
+
+internal data class AwaitDismissEffect(private val action: Action) : SideEffect {
```
andretortolano, line 9: Renaming AwaitEffect to AwaitDismissEffect since its being used in a particular place (when moving to EndingStep and waiting for UI to dismiss), better naming to help replace a boolean we had describing when we shouldPresent the new container, we know that every time we have this Effect it also means we want to present the next.

appcues/src/main/java/com/appcues/statemachine/states/BeginningExperienceState.kt
```diff
@@ -0,0 +1,29 @@
...
+
+internal data class BeginningExperienceState(val experience: Experience) : State {
+
+    override val currentExperience: Experience
+        get() = experience
+    override val currentStepIndex: Int?
+        get() = null
+
+    override fun take(action: Action): Transition? {
+        return if (action is StartExperience) {
```
andretortolano, line 18: beginning experience takes StartExperience (same from IdlingState) since it will always start at Step 0, we can then use StartStep for a more specific use case later.

appcues/src/main/java/com/appcues/statemachine/states/BeginningStepState.kt
```diff
@@ -0,0 +1,28 @@
...
+        get() = flatStepIndex
+
+    override fun take(action: Action): Transition? {
+        return if (action is RenderStep) {
+            toRenderingStep(action)
+        } else null
+    }
+
+    private fun toRenderingStep(action: RenderStep): Transition {
+        return next(RenderingStepState(experience, flatStepIndex, action.metadata))
```
andretortolano, line 26: Takes RenderStep(metadata) anda sets it to RenderingStepState, which is the new state the viewModel listens to to know the step it should render.

appcues/src/main/java/com/appcues/statemachine/states/EndingStepState.kt
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private fun toEndingExperience(action: EndExperience): Transition {
+        return next(
+            state = EndingExperienceState(experience, flatStepIndex, action.markComplete, action.trackAnalytics),
+            sideEffect = ContinuationEffect(Reset)
+        )
+    }
+
+    private fun toBeginningStep(action: StartStep): Transition {
```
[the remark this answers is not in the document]
andretortolano: gives us a more straightforward State, and we dont need to run the same checks/or throw error again

appcues/src/main/java/com/appcues/trait/AppcuesTraitException.kt
```diff
@@ -4,12 +4,11 @@ package com.appcues.trait
  * A type of exception that can occur during the application of an ExperienceTrait.
  */
 internal class AppcuesTraitException(
-    message: String,
+    override val message: String,
```
andretortolano, line 7: will undo this

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -229,9 +229,8 @@ internal class ExperienceRenderer(override val scope: AppcuesScope) : AppcuesCom
         }
     }
 
-    suspend fun dismiss(renderContext: RenderContext, markComplete: Boolean, destroyed: Boolean): ResultOf<State, Error> {
-        return stateMachines.getOwner(renderContext)?.stateMachine?.handleAction(EndExperience(markComplete, destroyed))
-            ?: Failure(RenderContextNotActive(renderContext))
```
andretortolano, line 234: Failure was not being used at all, RenderContextNotActive here doesnt make too much sense, we have other similar methods in this class that calls to handleAction, and neither returns this error

appcues/src/test/java/com/appcues/statemachine/effects/PresentationEffectTest.kt
```diff
@@ -0,0 +1,221 @@
...
+import io.mockk.coVerifySequence
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.verify
+import io.mockk.verifySequence
+import kotlinx.coroutines.test.advanceTimeBy
+import kotlinx.coroutines.test.runTest
+import org.junit.Test
+
+internal class PresentationEffectTest {
```
andretortolano, line 26: PresentationEffect is the complex effect, there is a bunch of tests that represent things that was previously being done by BeginningStepState

appcues/src/test/java/com/appcues/statemachine/states/BeginningExperienceStateTest.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.appcues.statemachine.State
+import com.appcues.statemachine.effects.PresentationEffect
+import com.google.common.truth.Truth.assertThat
+import io.mockk.mockk
+import org.junit.Test
+
+internal class BeginningExperienceStateTest {
+
+    @Test
+    fun `BeginningExperienceState SHOULD be instance of STATE`() {
```
andretortolano, line 14: checking to see if State is instance o State and what are the values on currentExperience and currentStepIndex (same test for each StateTest)

appcues/src/test/java/com/appcues/statemachine/states/BeginningExperienceStateTest.kt
```diff
@@ -0,0 +1,53 @@
...
+        val experience = mockk<Experience>()
+        val state = BeginningExperienceState(experience)
+        // THEN
+        assertThat(state).isInstanceOf(State::class.java)
+        assertThat(state.currentExperience).isEqualTo(experience)
+        assertThat(state.currentStepIndex).isNull()
+    }
+
+    @Test
+    fun `take SHOULD ignore listed actions`() {
```
andretortolano, line 25: check which actions we are not processing (similar test for all other states test)

appcues/src/test/java/com/appcues/statemachine/states/BeginningExperienceStateTest.kt
```diff
@@ -0,0 +1,53 @@
...
+                MockActions.RenderStep,
+                MockActions.EndExperience,
+                MockActions.Reset,
+                MockActions.ReportError,
+            )
+        )
+    }
+
+    @Test
+    fun `take StartExperience SHOULD transition to BeginningStepState`() {
```
andretortolano, line 43: and this is basically how we should test States, when it takes a valid Action, assert the new State and possible error or sideEffect

appcues/src/test/java/com/appcues/statemachine/states/StatesTestingHelper.kt
```diff
@@ -0,0 +1,57 @@
...
+
+internal fun Transition?.assertState(state: State) {
+    if (this != null) {
+        assertThat(newState).isEqualTo(state)
+    } else {
+        assert(false) { "transition should not be null" }
+    }
+}
+
+internal fun Transition?.assertEffect(effect: SideEffect?) {
```
andretortolano, line 30: when we are asserting side effect error should be null

appcues/src/test/java/com/appcues/statemachine/states/StatesTestingHelper.kt
```diff
@@ -0,0 +1,57 @@
...
+    if (this != null) {
+        assertThat(error).isNull()
+        assertThat(sideEffect).isEqualTo(effect)
+    } else {
+        assert(false) { "transition should not be null" }
+    }
+}
+
+internal fun Transition?.assertError(error: Error) {
```
andretortolano, line 39: when asserting error, side effect should be null (it will be ignored anyways but just in case we are validating here)

iujames (commented):
Impressive restructuring! I reviewed all the code changes in pretty good detail - did not review all the details of the tests, but can give initial comments/questions here first.

I'm not opposed to the overall approach here, just some things I'm noting below, as I try to understand it all. Key things to me are (1) ensuring no regressions in behavior or analytics - and given the level of testing here now that brings much more confidence and (2) maintaining logical consistency with how we think about the [experience state machine definition](https://whimsical.com/mobile-experience-states-Cb9WqPWhfUjb7tin4t5LYf) as much as possible (I think the `MoveToStep` action is the only material change there to mull over).

I really like the approach of moving logic out of the giant "Transitions" file and splitting responsibility into each State for handling it's own valid transitions.

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
andretortolano: so this was inconsistent with experienceStarted from config.experienceListener ?, I thought for some reason this was old code because both experience_started analytics and the experienceListener should be reporting at the same time. or thats at least what I thought

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -83,20 +88,20 @@ internal class ExperienceLifecycleTracker(
...
+            is BeginningExperienceState -> {
+                // update this value for auto-properties
+                storage.lastContentShownAt = Date()
+                trackLifecycleEvent(ExperienceStarted(experience), SdkMetrics.trackRender(experience.requestId))
+            }
+            is RenderingStepState -> {
                 trackLifecycleEvent(StepSeen(experience, flatStepIndex))
             }
-            is EndingStep -> {
+            is EndingStepState -> {
```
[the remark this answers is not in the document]
andretortolano: we can revert the naming, but since they are not part of a sealed class in ths impl this is the reason why I re-named to include State at the end

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -116,8 +121,7 @@ internal class ExperienceLifecycleTracker(
         when (this) {
             is Error.ExperienceError -> trackLifecycleEvent(ExperienceError(this))
             is Error.StepError -> trackLifecycleEvent(StepError(this))
-            is ExperienceAlreadyActive -> Unit
-            is RenderContextNotActive -> Unit
+            else -> Unit
```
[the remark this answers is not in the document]
andretortolano: good call

appcues/src/main/java/com/appcues/data/model/Step.kt
andretortolano: gotta think of a better way of referencing the presenting trait that is easier to understand and testable then. because the way it was before we had to consult a list to find the containerId, and then use the containerId to get the presentingTrait from ANOTHER list, there was always a missing branch that was impossible to test because it was impossible to reach, it usually means that we can simplify code.

appcues/src/main/java/com/appcues/statemachine/Action.kt
andretortolano: I drafted what I think it would look like now.

https://whimsical.com/mobile-experience-states-Cb9WqPWhfUjb7tin4t5LYf

the MoveToStep(stepReference) is useful because it saves us the need to re-check the nextIndex and container, currently we have basically duplicated code in both transitions, and we have noted in commends that from EndingStep to BeginningStep some of those checks are safe guards that SHOULD NEVER happen.

appcues/src/main/java/com/appcues/statemachine/states/BeginningStepState.kt
andretortolano: I believe that this is true for a while now. we dont rely on the UI to know we are actually rendering step, so basically the RenderStep action is only possible if we can produceMetada for that step via PresentationEffect, when we go from BeginningToRendering it means that presentingTrait was already called, and we are emiting RenderingStep to the ViewModel, if for some reason the presentingTrait fails we never get to move to RenderingStep. so no analytics is reported.

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
andretortolano: I drew up the this PR state machine design [here](https://whimsical.com/mobile-experience-states-Cb9WqPWhfUjb7tin4t5LYf) and you are right, this is actually changing behavior

It would be nice if we could use BeginningExperience for something though, or else it just seems to be an extra step that does nothing.

andretortolano: upon investigation I discovered that the only time we actually report a fatal error is when the PresentationEffect fails to create metadata setting trait, or present(), lets say we are in the first step, this happens BEFORE we get to RenderingStep which is Good. 

but if customers are listening to config.experienceListener, they will receive a experienceStarted, when in fact we never did. 

probably the flag isFirst is needed here, wouldnt that be the case for both the analytics and the listener?

appcues/src/main/java/com/appcues/statemachine/states/RenderingStepState.kt
```diff
@@ -0,0 +1,87 @@
...
+        return isLastStep && isContinuingToNextStep
+    }
+
+    private fun toEndingStep(action: MoveToStep): Transition {
+        val nextStepIndex = action.stepReference.getIndex(experience, flatStepIndex)
+            ?: return keep(StepError(experience, flatStepIndex, "Step at ${action.stepReference} does not exist"))
+
+        // should never happen but in case next step group lookup returns null we handle it here
+        val stepContainerIndex = experience.groupLookup[nextStepIndex]
+            ?: return keep(StepError(experience, flatStepIndex, "StepContainer for nextStepIndex $nextStepIndex not found"))
```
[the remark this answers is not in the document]
andretortolano: so usually what i've noticed was a lot of "this should ever happen" checks that had no "next state", this was the case for [this](https://github.com/appcues/appcues-android-sdk/blob/main/appcues/src/main/java/com/appcues/statemachine/Transitions.kt#L176C9-L176C24) check too.

and on current implementation when state is null we basically keep it in the same state, the only transition here that changed state to Idling is an exit failure when experience has no steps [here](https://github.com/appcues/appcues-android-sdk/blob/main/appcues/src/main/java/com/appcues/statemachine/Transitions.kt#L49). another "extra check" that is not really needed since we can check flatSteps size when we get to Idling.

appcues/src/main/java/com/appcues/data/model/Step.kt
andretortolano: I think thats the argument that the state machine doesnt know about "container", it only knows about steps, I guess this is the reason why it feels easier to put it inside the step....

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -129,19 +133,18 @@ internal class ExperienceLifecycleTracker(
...
-            is EndingExperience -> experience.published
-            is EndingStep -> experience.published
-            is Idling -> false
-            is RenderingStep -> experience.published
+            is BeginningExperienceState -> experience.published
+            is BeginningStepState -> experience.published
+            is EndingExperienceState -> experience.published
+            is EndingStepState -> experience.published
+            is RenderingStepState -> experience.published
+            else -> false
```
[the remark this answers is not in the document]
andretortolano: here it wont work the same becaue State is an interface and the when clause only force remaining branches on Sealed classes and enums.

appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
```diff
@@ -138,14 +138,13 @@ internal class ExperienceMapper(
...
-
+        val presentingTrait = mappedTraits.getExperiencePresentingTraitOrThrow()
         // this is where we will use groups to organize steps into stepContainers
         // also merge all necessary traits for each step
         StepContainer(
             id = stepContainerResponse.id,
-            steps = stepContainerResponse.children.map { step -> stepMapper.map(renderContext, step, mergedTraits) },
+            steps = stepContainerResponse.children.map { step -> stepMapper.map(renderContext, step, presentingTrait, mergedTraits) },
             actions = actionsMapper.map(renderContext, stepContainerResponse.actions),
-            presentingTrait = mappedTraits.getExperiencePresentingTraitOrThrow(),
```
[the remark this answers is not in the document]
andretortolano: the step is what is actually being presented as i noted above, the existence of StepContainer for the state machine is irrelevant, we can skip it and only consult flatSteps when going back and forth on steps.

the only reason we do use StepContainer is when we want to check if a step belongs in the same container.

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -229,9 +229,8 @@ internal class ExperienceRenderer(override val scope: AppcuesScope) : AppcuesCom
...
+    suspend fun dismiss(renderContext: RenderContext, markComplete: Boolean, destroyed: Boolean) {
```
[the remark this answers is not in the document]
andretortolano: yes

appcues/src/main/java/com/appcues/statemachine/states/BeginningStepState.kt
iujames: ok, that makes sense - thanks for clarifying.

appcues/src/main/java/com/appcues/data/mapper/experience/ExperienceMapper.kt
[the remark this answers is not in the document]
iujames: gotcha, yeah it seems not a huge deal - logically the presentation is part of the container, but i can understand how from the code perspective it simplifies things to have access to it from the step. It would only ever really be used by the first step in the container, so it is a little unexpected that say steps 2, 3, and so on have a presenting trait, but it's an implementation detail, fair enough.

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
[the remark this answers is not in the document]
iujames: would it work to use sealed classes here?

appcues/src/main/java/com/appcues/statemachine/Action.kt
iujames: thanks for the updated diagram, that helps me understand - feels like an implementation detail we can live with being slightly different on Android then, in exchange for the code simplification. The overall state flow diagram is still generally matching, which is the key thing I think.

iujames (approved):

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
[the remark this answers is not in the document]
andretortolano: unfortunately sealed classes need to be defined together, that would make for a Huge State file, thats the main reason I moved to interface, give clear abstraction.

appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
```diff
@@ -1,214 +1,77 @@
...
+    initialState: State = IdlingState,
+    private val onEndedExperience: (Experience) -> Unit = {},
 ) {
 
     private val _stateFlow = MutableSharedFlow<State>(1)
     val stateFlow: SharedFlow<State>
         get() = _stateFlow
 
-    private val _errorFlow = MutableSharedFlow<Error>(1)
+    private val _errorFlow = MutableSharedFlow<Error>()
```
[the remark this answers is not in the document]
andretortolano: we can keep an eye for it but ideally we dont need to keep the last error since we are emitting errors BEFORE we stop the experienceLifetracking, so this is so when we start observing again we dont get old value (like we had before where we had to call resetReplayCache()

appcues/src/main/java/com/appcues/statemachine/State.kt
```diff
@@ -1,54 +1,19 @@
...
-        val onUiDismissed: (() -> Unit)?,
-    ) : State()
-
-    data class EndingExperience(
-        val experience: Experience,
-        val flatStepIndex: Int,
-        val markComplete: Boolean,
-        val trackAnalytics: Boolean, // to disable analytics on force stop
-    ) : State()
+internal interface State {
```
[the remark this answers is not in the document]
andretortolano: yes but we dont need it before we had that huge method that would deal with all possible states at once, now we call State.take and this abstracts the implementation that is actually handling the action.

PR #510, opened 2023-10-03
🔧 removing codecov.gradle

andretortolano:


appcues/build.gradle
```diff
@@ -3,12 +3,16 @@ plugins {
     id 'kotlin-android'
     id 'kotlin-kapt'
     id 'org.jetbrains.dokka'
+    id 'jacoco'
+}
+
+jacoco {
+    toolVersion = "0.8.9"
 }
```
iujames, line 11: I'm not sure if this part is required, it is also set in the spec repo and I think the plugin version defined in the project level build gradle defines it? Maybe harmless, but could explore removing also.

iujames:
just want to check and see how the final report comes out on this one from the spec repo tests

iujames (approved):

PR #512, opened 2023-10-03
Run android sample release build and detekt jobs in CI

iujames:
stacks on #510 

Runs detekt and the sample app release build on branch validation in CI, in a job in this repo while the spec repo tests run in parallel.


andretortolano (approved):
yes!

PR #513, opened 2023-10-03
♻️  Removing unused file ExperienceTestHelper

andretortolano:


iujames (commented):
I believe we can also remove the imageLoader here, no longer used: https://github.com/appcues/appcues-android-sdk/blob/main/appcues/src/main/java/com/appcues/AppcuesConfig.kt#L99-L100

Basically, can revert much of this one? https://github.com/appcues/appcues-android-sdk/pull/485

Then, will need the corresponding changes from this one reverted https://github.com/appcues/appcues-mobile-experience-spec/pull/215 (at least the TestHostApp.kt usage spot)

iujames:
@andretortolano ok if we close this one? I suggest we leave as-is for now, as this still has use in the UI tests we're now running with Roborazzi.

Issue #517, opened 2023-10-18
Completely disable the SDK


nelletto:
We use a configuration system through which we have decided, for the moment, to completely turn off Appcues. However, we noticed that the initialization of a part of the SDK still occurs via `AppcuesInitializer`. Is it possible to have a way to completely deactivate the SDK programmatically (therefore without using `tools:node="remove"`) so that we can activate or deactivate it when we need it?

iujames:
Hi @nelletto - I believe to support this use case, we would need to add a couple of functions in the public API of our SDK to `start(context)` and `stop(context)` these services. A proposed approach is in https://github.com/appcues/appcues-android-sdk/pull/518 , if you would like to give any feedback. Do you think this would address your use case?

iujames:
Hi @nelletto , this update has been merged and a new version `3.1.3` has been published to Maven Central. I'll go ahead and close this issue now. Thanks for your feedback here!

PR #518, opened 2023-10-18
Add support for programmatic control of InitializationProvider services

iujames:
Provides programmatic access to start and stop SDK functionality that is currently initialized by the AppcuesInitializer (an `androidx.startup.Initializer`). It would not be expected to need this access in a typical install, which will continue to work the same, by default. But this would support more advanced control as requested in #517


andretortolano (approved):
Looks good

appcues/src/main/java/com/appcues/AppcuesProviderService.kt
```diff
@@ -0,0 +1,52 @@
...
+     * The [context] parameter should be the currently active Activity, if possible, to ensure
+     * that the Appcues SDK has this Activity available to present content.
+     *
+     * @param context The Android Context used by the host application.
+     */
+    public fun start(context: Context) {
+        // stop any existing services first before starting new
+        stop(context)
+
+        val application = context.applicationContext as Application
```
nelletto, line 32: First of all, thank you for accepting our request. In our experience we have had crashes due to application context null, for example when initializing a library in `onAttachBaseContext`. personally I would suggest doing this check and logging the exception, what do you think?

iujames: Good call - added a check for this in the commit [here](https://github.com/appcues/appcues-android-sdk/pull/518#:~:text=Resolve%20conversation-,%F0%9F%91%8C%20Add%20check%20for%20null%20application%20context,-f71d276). Would that work ok?

nelletto: perfect 👌

nelletto: it's perfect as in pull request, for us the important thing was only to avoid the crash and have a log entry :)

PR #519, opened 2023-10-19
Do not dismiss and re-render an experience with the same instance from cache

iujames:
Related to a recent customer investigation on the iOS side. Scenario:
* embed is triggered on session start (event trigger) and placed in cache
* later gets rendered ok
* another analytic occurs (i.e. identify, not a screen)
* no new experiences, and no cache clear, cache is re-processed
* the experience in cache tries to re-render by dismissing the current experience and re-starting --> not desired

We only want to dismiss the current experience if (1) it is a new experience from the server of normal priority (not a screen view) or (2) it is an updated version of the current experience from server (new instance id)

Also, rearranged the flow of checks a bit in this show method - we should check for invalid frame, interceptor blocking, or this instance check _before_ triggering an experiment_entered analytic for a control group, to match iOS.


andretortolano (approved):
Like the addition of tests here.

PR #521, opened 2023-10-23
✨ Add Overlay to highlight selectable elements

andretortolano:
As part of this [ticket](https://app.shortcut.com/appcues/story/58343/add-target-overlay-and-help-text-to-the-screen-capture-confirm-screen). we need to include a highlighting overlay for selectable elements during screen capture.

this is how it looks after this change:

image


iujames:
it looks like the test issue is totally unrelated and is that flakey analytics test, sigh. Maybe we need to take those out of CI for now.

iujames (approved):
nice!

PR #523, opened 2023-10-23
✨ Add Extra texts regarding captured screen and link to docs

andretortolano:
Completing the task to match iOS when capturing screen

when we get to the confirmation dialog we show extra information and give a link to docs where users can troubleshoot in case they are not finding what they are trying to target.

|Some selectors|No selectors|
|-|-|
|||


iujames (approved):
Nice, I'll get some tests to cover this in the work I'm doing currently.

I think it looks like a tad bit much padding around the "Tap to view..." link, personally (also compared to iOS), but maybe this is ok and matching the material tap target height guidelines.

PR #525, opened 2023-10-31
Fix element target visibility check

iujames:
This is a defect discovered while doing other tooltip testing. We were previously not checking the return value of [getGlobalVisibleRect](https://developer.android.com/reference/android/view/View#getGlobalVisibleRect(android.graphics.Rect)). However, a false value returned means none of the view is visible, and "If the method returns false, the contents of r are undefined." This means we were previously allowing some undefined rect values to flow through as target element rects, incorrectly.

This could lead to undesired behavior like this:

| incorrect target | target button is actually behind bottom nav |
| --- | --- |
|  Screenshot 2023-10-30 at 4 40 51 PM | Screenshot 2023-10-30 at 4 41 51 PM |

This change updates to actually check the return value and ignore the view if no valid Rect is found. There is another check to confirm that the screen bounds contains the rect, which was already in place - this may be redundant really, based on the documentation for getGlobalVisibleRect, but I was not sure and did not want to cause any regressions from previous behavior, so leaving that as-is for now.

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
+++ b/appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
@@ -104,17 +104,20 @@ private fun View.asCaptureView(screenBounds: Rect): ViewElement? {
     val displayMetrics = context.resources.displayMetrics
     val density = displayMetrics.density
 
-    // this is the position of the view relative to the entire screen
-    val actualPosition = Rect()
-    getGlobalVisibleRect(actualPosition)
+    // the coordinates of the non-clipped area of this view in the coordinate space of the view's root view
+    val globalVisibleRect = Rect()
 
-    // if the view is not currently in the screenshot image (scrolled away), ignore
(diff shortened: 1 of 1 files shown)


appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
```diff
@@ -104,17 +104,20 @@ private fun View.asCaptureView(screenBounds: Rect): ViewElement? {
...
+
+    if (
+        // ignore the Appcues SDK content that has been injected into the view hierarchy
+        this.isAppcuesView() ||
+        // if getGlobalVisibleRect returns false, that indicates that none of the view is
+        // visible within the root view, and we will not include it in our capture
+        getGlobalVisibleRect(globalVisibleRect).not() ||
+        // if the view is not currently in the screenshot image (scrolled away), ignore
+        // (this is possibly a redundant check to item above, but keeping for now)
+        Rect.intersects(globalVisibleRect, screenBounds).not()
```
andretortolano, line 118: better be safe than sorry

andretortolano (approved):
nice catch!

PR #528, opened 2023-11-01
Add state machine capabilities for FailingState and Retry action

iujames:
This is some foundational work that will support future research on tooltip error/retry handling. I wanted to put up a preliminary PR checkpoint for feedback.

When an error occurs in the PresentationEffect now, it will use ReportError (as before) but also provide a retry SideEffect object as part of the ReportError Action. This SideEffect is whatever should be retried, if a recover attempt is to be made - in this case, the PresentationEffect again.

The StateMachine then has the ability to handle ReportError by transitioning to a FailingState, which captures the current state (to return to) and the SideEffect to retry, to try again later. FailingState can respond to 
* Retry - try the effect again
* StartExperience - which means something new has come into the machine and it should give up (transition back to Idling from the error), and continue on with the new StartExperience as normal.
* EndExperience - edge case if a new event triggered experience comes in and it needs to dismiss any active experience

The concept of "fatal error" is removed - as it was only used in once spot - the PresentationEffect, and this is now a StepError that supports retry. StepError now tracks the renderErrorId in the experience (like we had for embed errors) and the related capability to have a StepRecovered analytic event tracked if they later get retried and make it to StepSeen.

* The first commit here has the core of the new logic
* The second commit contains new tests for all the new logic around state machine transitions and related analytics

### DIFF
--- a/appcues/src/main/java/com/appcues/statemachine/Error.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/Error.kt
@@ -16,5 +16,10 @@ internal sealed class Error(open val message: String) {
     object ExperienceAlreadyActive : Error("Experience already active")
 
     data class ExperienceError(val experience: Experience, override val message: String) : Error(message)
-    data class StepError(val experience: Experience, val stepIndex: Int, override val message: String) : Error(message)
+    data class StepError(
+        val experience: Experience,
+        val stepIndex: Int,
+        override val message: String,
+        var recoverable: Boolean = false,
+    ) : Error(message)
 }
--- a/appcues/src/main/java/com/appcues/statemachine/Error.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/Error.kt
@@ -3,17 +3,18 @@ package com.appcues.statemachine
 import com.appcues.data.model.Experience
 import java.util.UUID
 
-internal sealed class Error(open val message: String, errorId: UUID? = null) {
+internal sealed class Error(open val message: String) {
+    private val _id = UUID.randomUUID()
 
-    val id: UUID = errorId ?: UUID.randomUUID()
+    // allows for override of error ID with a value driven by the experience.renderErrorId
+    // for matching error/recovery IDs on render retry
+    var errorId: UUID? = null
+
+    val id: UUID
+        get() = errorId ?: _id
 
     object ExperienceAlreadyActive : Error("Experience already active")
 
-    data class ExperienceError(
-        val experience: Experience,
-        override val message: String,
-        val errorId: UUID? = null
-    ) : Error(message, errorId)
-
+    data class ExperienceError(val experience: Experience, override val message: String) : Error(message)
     data class StepError(val experience: Experience, val stepIndex: Int, override val message: String) : Error(message)
 }
--- a/appcues/src/main/java/com/appcues/statemachine/Error.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/Error.kt
@@ -20,6 +20,6 @@ internal sealed class Error(open val message: String) {
         val experience: Experience,
         val stepIndex: Int,
         override val message: String,
-        var recoverable: Boolean = false,
+        val recoverable: Boolean = false,
     ) : Error(message)
 }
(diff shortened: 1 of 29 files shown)


appcues/src/main/java/com/appcues/statemachine/SideEffect.kt
```diff
@@ -4,5 +4,5 @@ import com.appcues.action.ActionProcessor
 
 internal interface SideEffect {
 
-    suspend fun launch(processor: ActionProcessor): Action?
+    suspend fun launch(duringTransition: Transition, processor: ActionProcessor): Action?
```
andretortolano, line 7: I like the idea but I think there should be a way for us to infer the transition (current state + side effect) without needing to pass it in the launch. maybe we can have ReportError(shouldRetry = true) and if this is the next action (in the state machine take logic) then we can move to FailingState(Transition(currentState, sideEffect)) to avoid passing this "duringTransition" to every sideEffect.

the idea of this launch methods to provide a way for "outside dependencies" like action processor to be used by side effects.

iujames: @andretortolano this is all updated now to remove this extra `duringTransition` param, like we chatted about. Thanks for the ideas!

appcues/src/main/java/com/appcues/statemachine/StateMachine.kt
```diff
@@ -24,11 +26,16 @@ internal class StateMachine(
 ) {
 
     private val _stateFlow = MutableSharedFlow<State>(1)
-    val stateFlow: SharedFlow<State>
-        get() = _stateFlow
+    val stateFlow: Flow<State>
+        // important that we use .distinctUntilChanged explicitly here, as it is not enabled
+        // by default on SharedFlows (only on StateFlows) - duplicate states can sent
+        // during error transitions, and we don't want those to actually be emitted to listeners
+        get() = _stateFlow.distinctUntilChanged()
```
iujames, line 33: @andretortolano fyi bundled this fix in here as well

it is quite confusing that we have something named "state flow" that is really a SharedFlow 🙃

appcues/src/main/java/com/appcues/statemachine/Error.kt
```diff
@@ -3,17 +3,18 @@ package com.appcues.statemachine
...
 import java.util.UUID
 
-internal sealed class Error(open val message: String, errorId: UUID? = null) {
+internal sealed class Error(open val message: String) {
+    private val _id = UUID.randomUUID()
 
-    val id: UUID = errorId ?: UUID.randomUUID()
+    // allows for override of error ID with a value driven by the experience.renderErrorId
+    // for matching error/recovery IDs on render retry
+    var errorId: UUID? = null
```
andretortolano, line 11: i guess we could also do val id:: UUID = UUID.randomUUID() ?

andretortolano: all errors could have a UUID wouldnt hurt. for simplicity

iujames: yeah all errors do have an ID, it is just that sometimes we need to override the ID with a "render error ID" that matches that value from the experience that gets set. That way, we can create another recovery event with a matching error ID - to tie the error to the recovery correctly.

appcues/src/main/java/com/appcues/statemachine/states/FailingState.kt
```diff
@@ -0,0 +1,35 @@
...
+    override val currentExperience: Experience?
+        get() = targetState.currentExperience
+    override val currentStepIndex: Int?
+        get() = targetState.currentStepIndex
+
+    override fun take(action: Action): Transition? {
+        return when (action) {
+            is Retry -> next(targetState, retryEffect)
+            // a new start means recovery never happened, move to Idling and continue with the new start attempt
+            is StartExperience -> Transition(IdlingState, null, ContinuationEffect(action))
```
andretortolano, line 28: next(IdlingState, ContinuatationEffect(action))

andretortolano (approved):
left some minor comments, but it looks good now, thanks!

PR #530, opened 2023-11-03
🐛 Back press not being handled on Experiences

andretortolano:
related to https://app.shortcut.com/appcues/story/42518/non-skippable-experience-can-be-skipped-by-device-back-button-or-gesture noticed that since we moved away from the Activity approach, experiences are not handling the back press properly, this change takes into consideration a possible feature "Block" users from dismissing experience by not including skippable traits into the step, but also includes a fix to make so experiences showing can properly handle the native back press


iujames (approved):
looks good! makes sense to me to allow the native back button/gesture on Android work like a backdrop tap and dismiss, when the current step has been marked to support skipping the experience.

PR #531, opened 2023-11-03
Add retry handling on tooltip render when ui layout updates are detected

iujames:
Stacks on error/retry handling in #528 

Creating a draft here for feedback on approach.

The bulk of the work is being done in the `ModalStateMachineOwner` now - split out from the main ExperienceRenderer, as this is Modal render context specific logic. There are a variety of options available to observe changes in the view tree, using a ViewTreeObserver. This starts out by watching scroll, global layout changes, and draw updates - detecting when the UI is idling by waiting for a 1 sec period after all updates have concluded, before attempting a retry.

The retry is controlled by listening for StepErrors that are marked recoverable. Only selected errors are such type, like "no target element found" and "multiple matching targets" - things that _could_ be resolved by UI updates. Each time a retry is triggered, all the listeners stop, and if another error occurs, then they re-attach and keep trying. If a screen view is processed - retry handling also stops (permanently for that failed experience).

This would support a tooltip step failing when it was the first step in a group - new container presentation, as well as a subsequent step in the group - where the container would be removed, then re-presented on the desired step, if it recovers.

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
+++ b/appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
@@ -115,7 +115,7 @@ private fun View.asCaptureView(screenBounds: Rect): ViewElement? {
         getGlobalVisibleRect(globalVisibleRect).not() ||
         // if the view is not currently in the screenshot image (scrolled away), ignore
         // (this is possibly a redundant check to item above, but keeping for now)
-        Rect.intersects(globalVisibleRect, screenBounds).not()
+        screenBounds.contains(globalVisibleRect).not()
     ) {
         // if any of these conditions failed, this view is not captured
         return null
@@ -151,8 +151,9 @@ private fun View.asCaptureView(screenBounds: Rect): ViewElement? {
             val semanticsOwnerField = androidComposeViewClass.getDeclaredField("semanticsOwner")
                 .apply { isAccessible = true } // make private filed accessible
             val semanticsOwner = semanticsOwnerField.get(this) as SemanticsOwner
-            val composeChildren = listOf(semanticsOwner.rootSemanticsNode.asCaptureView(context))
-            children.addAll(composeChildren)
+            semanticsOwner.rootSemanticsNode.asCaptureView(context, screenBounds)?.let {
+                children.add(it)
+            }
         } catch (_: Exception) {
             // Catching and swallowing exceptions here with the Compose view handling in case
             // something changes in the future that breaks the expected structure being accessed
@@ -161,7 +162,7 @@ private fun View.asCaptureView(screenBounds: Rect): ViewElement? {
         }
     }
 
-    return selector().let {
+    return selector(globalVisibleRect).let {
         ViewElement(
             x = globalVisibleRect.left.toDp(density),
             y = globalVisibleRect.top.toDp(density),
@@ -170,12 +171,26 @@ private fun View.asCaptureView(screenBounds: Rect): ViewElement? {
             displayName = it?.displayName,
             selector = it,
             type = it?.type ?: this::class.java.simpleName,
-            children = if (children.isEmpty()) null else children,
+            children = children.ifEmpty { null },
         )
     }
 }
 
-private fun View.selector(): AndroidViewSelector? {
+private fun View.isFullyVisible(globalVisibleRect: Rect, tolerance: Double = 0.1): Boolean {
+    val visibleHeight = globalVisibleRect.height().toDouble()
+    val actualHeight = measuredHeight.toDouble()
+    val visibleWidth = globalVisibleRect.width().toDouble()
+    val actualWidth = measuredWidth.toDouble()
+
+    val heightVisible = visibleHeight > actualHeight * (1.0 - tolerance)
+    val widthVisible = visibleWidth > actualWidth * (1.0 - tolerance)
+    return heightVisible && widthVisible
+}
+
+private fun View.selector(globalVisibleRect: Rect): AndroidViewSelector? {
+    // if the entire view is not visible (within tolerance) exclude selector for targeting
+    if (isFullyVisible(globalVisibleRect).not()) return null
+
     return AndroidViewSelector(
         properties = mapOf(
             SELECTOR_CONTENT_DESCRIPTION to contentDescription?.toString(),
@@ -195,16 +210,17 @@ private fun View.extractResourceName(): String? {
     }
 }
 
-private fun SemanticsNode.asCaptureView(context: Context): ViewElement {
+private fun SemanticsNode.asCaptureView(context: Context, screenBounds: Rect): ViewElement? {
     val displayMetrics = context.resources.displayMetrics
     val density = displayMetrics.density
 
-    var childElements: List<ViewElement>? = children.map {
-        it.asCaptureView(context)
+    // if the view is not currently in the screenshot image (scrolled away), ignore
+    if (screenBounds.contains(unclippedGlobalBounds).not()) {
+        return null
     }
 
-    if (childElements?.isEmpty() == true) {
-        childElements = null
+    val childElements: List<ViewElement> = children.mapNotNull {
+        it.asCaptureView(context, screenBounds)
     }
 
     return selector().let {
@@ -216,7 +232,7 @@ private fun SemanticsNode.asCaptureView(context: Context): ViewElement {
             displayName = it?.displayName,
             selector = it,
             type = it?.type ?: "Composable #$id",
-            children = childElements
+            children = childElements.ifEmpty { null },
         )
     }
 }
--- a/appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
+++ b/appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
@@ -115,7 +115,7 @@ private fun View.asCaptureView(screenBounds: Rect): ViewElement? {
         getGlobalVisibleRect(globalVisibleRect).not() ||
         // if the view is not currently in the screenshot image (scrolled away), ignore
         // (this is possibly a redundant check to item above, but keeping for now)
-        screenBounds.contains(globalVisibleRect).not()
+        Rect.intersects(globalVisibleRect, screenBounds).not()
     ) {
         // if any of these conditions failed, this view is not captured
         return null
--- a/appcues/src/main/java/com/appcues/statemachine/Error.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/Error.kt
@@ -16,5 +16,10 @@ internal sealed class Error(open val message: String) {
     object ExperienceAlreadyActive : Error("Experience already active")
 
     data class ExperienceError(val experience: Experience, override val message: String) : Error(message)
-    data class StepError(val experience: Experience, val stepIndex: Int, override val message: String) : Error(message)
+    data class StepError(
+        val experience: Experience,
+        val stepIndex: Int,
+        override val message: String,
+        var recoverable: Boolean = false,
+    ) : Error(message)
 }
--- a/appcues/src/main/java/com/appcues/statemachine/Error.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/Error.kt
@@ -20,6 +20,6 @@ internal sealed class Error(open val message: String) {
         val experience: Experience,
         val stepIndex: Int,
         override val message: String,
-        var recoverable: Boolean = false,
+        val recoverable: Boolean = false,
     ) : Error(message)
 }
--- a/appcues/src/main/java/com/appcues/statemachine/effects/PresentationEffect.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/effects/PresentationEffect.kt
@@ -16,6 +16,7 @@ internal data class PresentationEffect(
     private val flatStepIndex: Int,
     private val stepContainerIndex: Int,
     private val shouldPresent: Boolean,
+    private val isRecovering: Boolean = false,
 ) : SideEffect {
 
     override suspend fun launch(processor: ActionProcessor): Action {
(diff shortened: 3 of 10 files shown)


appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
```diff
@@ -195,16 +210,17 @@ private fun View.extractResourceName(): String? {
...
-private fun SemanticsNode.asCaptureView(context: Context): ViewElement {
+private fun SemanticsNode.asCaptureView(context: Context, screenBounds: Rect): ViewElement? {
     val displayMetrics = context.resources.displayMetrics
     val density = displayMetrics.density
 
-    var childElements: List<ViewElement>? = children.map {
-        it.asCaptureView(context)
+    // if the view is not currently in the screenshot image (scrolled away), ignore
+    if (screenBounds.contains(unclippedGlobalBounds).not()) {
+        return null
```
iujames, line 219: we were not previously checking Compose views to see if they were in the screen bounds - this fixes that portion. I plan to pull this out into a separate PR on main, as it should be fixed regardless of where this retry work ends up.

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
```diff
@@ -170,12 +171,26 @@ private fun View.asCaptureView(screenBounds: Rect): ViewElement? {
...
+    val actualWidth = measuredWidth.toDouble()
+
+    val heightVisible = visibleHeight > actualHeight * (1.0 - tolerance)
+    val widthVisible = visibleWidth > actualWidth * (1.0 - tolerance)
+    return heightVisible && widthVisible
+}
+
+private fun View.selector(globalVisibleRect: Rect): AndroidViewSelector? {
+    // if the entire view is not visible (within tolerance) exclude selector for targeting
+    if (isFullyVisible(globalVisibleRect).not()) return null
```
iujames, line 192: this new check ensures that 90% of the height and width of a view will be visible on screen before any tooltip can point to it - to handle things scrolling into the view.

the Compose side didn't really need such a check (i think) as we are already constructing the view rect and checking if it is within screen bounds (new check on L218) - need to test/confirm behavior on this part.

appcues/src/main/java/com/appcues/ui/ModalStateMachineOwner.kt
```diff
@@ -0,0 +1,167 @@
...
+            }
+        }
+    }
+
+    private fun attemptRetry() {
+        onUiIdle {
+            // cancel any current touch (drag) as our overlay is about to drop on top and we don't want things moving behind it
+            AppcuesActivityMonitor.activity?.getParentView()?.dispatchTouchEvent(
+                MotionEvent.obtain(0L, 0L, MotionEvent.ACTION_CANCEL, 0f, 0f, 0)
+            )
```
iujames, line 78: this needs to happen _before_ the retry, since it might cause pager views to "snap" into place, and the tooltip needs to be positioned based on the final position of the view, not the position prior to that auto-snap animation.

appcues/src/main/java/com/appcues/ui/ModalStateMachineOwner.kt
```diff
@@ -0,0 +1,167 @@
...
+            coroutineScope.launch {
+                // stop listening for updates, will re-attach if another error occurs
+                stopRetryHandling()
+                stateMachine.handleAction(Retry)
```
iujames, line 82: this is the key line that initiates the retry

appcues/src/main/java/com/appcues/ui/ModalStateMachineOwner.kt
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    private class ViewTreeUpdateHandler(private var action: () -> Unit) :
+        OnScrollChangedListener,
+        OnGlobalLayoutListener,
+        OnTouchModeChangeListener,
+        OnDrawListener,
+        OnGlobalFocusChangeListener,
+        OnWindowAttachListener,
+        OnWindowFocusChangeListener
```
iujames, line 106: not all of these are used (they were used for logging an experiementation) - this will be cleaned up on any final pass over this code to trim down to the set of listeners needed. At this point it appears that scroll change, draw, and global layout listeners will suffice - and watching those until they become Idle gives us an indicator that we can re-query the UI for missing elements

appcues/src/main/java/com/appcues/ui/ModalStateMachineOwner.kt
```diff
@@ -0,0 +1,167 @@
...
+    private var viewTreeObserver: ViewTreeObserver? = null
+
+    init {
+        // set up observer on the Modal state machine errors to initiate retry
+        coroutineScope.launch(Dispatchers.IO) {
+            stateMachine.errorFlow.collect { error ->
+                // if a recoverable step error is observed, begin attempting retry
+                if (error is StepError && error.recoverable) {
+                    startRetryHandling()
+                }
```
iujames, line 50: we don't want to have retry handling going on all the time in the customer app - as rarely as possible - so this allows it to ONLY happen in special cases of recoverable error, and it will stop (on L81) as soon as a retry is attempted (until another error hits or the screen changes)

appcues/src/main/java/com/appcues/statemachine/Error.kt
```diff
@@ -16,5 +16,10 @@ internal sealed class Error(open val message: String) {
     object ExperienceAlreadyActive : Error("Experience already active")
 
     data class ExperienceError(val experience: Experience, override val message: String) : Error(message)
-    data class StepError(val experience: Experience, val stepIndex: Int, override val message: String) : Error(message)
+    data class StepError(
+        val experience: Experience,
+        val stepIndex: Int,
+        override val message: String,
+        var recoverable: Boolean = false,
```
andretortolano, line 23: can be `val`

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
```diff
@@ -115,7 +115,7 @@ private fun View.asCaptureView(screenBounds: Rect): ViewElement? {
         getGlobalVisibleRect(globalVisibleRect).not() ||
         // if the view is not currently in the screenshot image (scrolled away), ignore
         // (this is possibly a redundant check to item above, but keeping for now)
-        Rect.intersects(globalVisibleRect, screenBounds).not()
+        screenBounds.contains(globalVisibleRect).not()
```
andretortolano, line 118: can you expand on this change?

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
```diff
@@ -170,12 +171,26 @@ private fun View.asCaptureView(screenBounds: Rect): ViewElement? {
...
             selector = it,
             type = it?.type ?: this::class.java.simpleName,
-            children = if (children.isEmpty()) null else children,
+            children = children.ifEmpty { null },
         )
     }
 }
 
-private fun View.selector(): AndroidViewSelector? {
+private fun View.isFullyVisible(globalVisibleRect: Rect, tolerance: Double = 0.1): Boolean {
```
andretortolano, line 179: good isolated method

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
andretortolano: oh I guess our check works for the first ComposeView, but the subsequent ones (inside the SemanticsNode) can be placed outside, so this extra check makes sense, so in this case should this be similar to isFullyVisible? (with tolerance maybe)

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
```diff
@@ -195,16 +210,17 @@ private fun View.extractResourceName(): String? {
...
+        return null
     }
 
-    if (childElements?.isEmpty() == true) {
-        childElements = null
+    val childElements: List<ViewElement> = children.mapNotNull {
+        it.asCaptureView(context, screenBounds)
     }
 
     return selector().let {
```
andretortolano, line 226: wondering if there is a way to make both `isFullyVisible()` in a place that is similar in relation

appcues/src/main/java/com/appcues/statemachine/effects/PresentationEffect.kt
```diff
@@ -16,6 +16,7 @@ internal data class PresentationEffect(
     private val flatStepIndex: Int,
     private val stepContainerIndex: Int,
     private val shouldPresent: Boolean,
+    private val isRecovering: Boolean = false,
```
andretortolano, line 19: oh I like that, clear data structure for recovering flow.

appcues/src/main/java/com/appcues/trait/AppcuesTraitException.kt
```diff
@@ -11,4 +11,11 @@ internal class AppcuesTraitException(
      *  of traits can be attempted, to try to auto-recover from this error.
      */
     val retryMilliseconds: Int? = null,
+
+    /**
+     * When true, this means that the issue may be recoverable at a later time, for instance
+     * if a target element is not found, but future layout updates to scroll other content
+     * into view could resolve the target.
+     */
+    val recoverable: Boolean = false
```
andretortolano, line 20: ok following along there is a new  property for exception to describe if its recoverable.

appcues/src/main/java/com/appcues/trait/appcues/TargetElementTrait.kt
```diff
@@ -114,6 +115,7 @@ internal class TargetElementTrait(
         throw AppcuesTraitException(
             message = "multiple non-distinct views (${weightedViews.count()}) matched selector ${selector.toMap()}",
             retryMilliseconds = retryMilliseconds,
+            recoverable = true,
```
andretortolano, line 118: then on TargetElementTrait we mark as recoverable, so not all Exceptions can be recovered

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -55,6 +55,8 @@ internal class ExperienceRenderer(override val scope: AppcuesScope) : AppcuesCom
     private val potentiallyRenderableExperiences = hashMapOf<RenderContext, List<Experience>>()
     private val previewExperiences = hashMapOf<RenderContext, Experience>()
 
+    private val modalStateMachineOwner = ModalStateMachineOwner(get(::onExperienceEnded), get())
```
andretortolano, line 58: so instead of extracting the modalStateMachineOwner from the stateMachines object, would it make more sense to have some logic around the stateMachiens.cleanup() that could call "stopRetry"? maybe each StatemachineOwner can have their on cleanup() method, so things that are particular to each StateMachineOwner implementation can be handled internally

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -135,6 +137,9 @@ internal class ExperienceRenderer(override val scope: AppcuesScope) : AppcuesCom
             // clear list in case this was a screen_view qualification
             potentiallyRenderableExperiences.clear()
             stateMachines.cleanup()
+
+            // and stop any Modal retry processing ongoing when screen changes
+            modalStateMachineOwner.stopRetryHandling()
```
andretortolano, line 142: following on previous comment, we could remove this.

appcues/src/main/java/com/appcues/ui/StateMachineDirectory.kt
```diff
@@ -26,15 +25,6 @@ internal class AppcuesFrameStateMachineOwner(
...
-internal class ModalStateMachineOwner(override val stateMachine: StateMachine) : StateMachineOwning {
-
-    override val renderContext: RenderContext = Modal
-
-    override suspend fun reset() {
-        stateMachine.stop(true)
-    }
-}
-
 internal class StateMachineDirectory {
```
andretortolano, line 28: add to cleanup() to something like 

```kotlin
fun cleanup() {
   stateMachines.value.cleanup()

   stateMachines.value.removeAll { it.isInvalid() }
}
```

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
andretortolano: maybe we can evolve this later, but the only thing I dont like about this entire appraoch is that the recoverable properties in Exception and StepError seems to be directly tied to basically waiting them to scroll to retry, so I'm trying to imagine if in the future we have other type of recoverable tasks (might never happen), how would the StateMachineOwning be able to differ one from the other, since they would both be StepError && recoverable

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
iujames: I think it is probably not changing anything, in reality. I was trying to change the logic here to say: the screen bounds has to contain the entire view rect -- but this `globalVisibleRect` is just the visible rect - the part that is in the screen bounds, so this is always the same result as checking if it intersects -- and in reality, this is probably all redundant since we now check the return value on `getGlobalVisibleRect` and that indicates something is not on the screen (if false)

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
iujames: do you mean check the Compose views for isFullyVisible?

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
iujames: yep, would certainly need to evolve if there are different types of recovery. Maybe some enum of recovery types, and this type is "UI changed" -- i.e. check again when you detect the UI changed in some meaningful way, rather than just one generic `recoverable` boolean.

appcues/src/main/java/com/appcues/ui/StateMachineDirectory.kt
iujames: I updated this in the latest commit if you want to take a look and see what you think. I think it did clean things up nicely. I changed `cleanup()` to `onScreenChange()` to more closely reflect what it was used for. I did that because I found it confusing to have that interface have both `reset()` and `cleanup()` functions that really had two different meanings, but sounded very similar and generic.

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
andretortolano: yes, right now we are only checking that for View class correct?

andretortolano (approved):
the approach looks good, considering we only want to recover for a single type of error today

PR #532, opened 2023-11-06
Update Proguard rules to protect reflection usage

iujames:
Fix for customer issue where a failure to process Compose layouts in release builds was observed.


appcues/build.gradle
```diff
@@ -1,8 +1,8 @@
 plugins {
     id 'com.android.library'
     id 'kotlin-android'
-    id 'kotlin-kapt'
     id 'org.jetbrains.dokka'
+    id 'kotlin-kapt'
```
iujames, line 5: I noticed during test builds that this warning was showing up
```
> Task :appcues:kaptReleaseKotlin
warning: Kapt support in Moshi Kotlin Code Gen is deprecated and will be removed in 2.0. Please migrate to KSP. https://github.com/square/moshi#codegen
warning: The following options were not recognized by any processor: '[room.schemaLocation, kapt.kotlin.generated]'
```

I don't think it actually caused anything stop working, but I'm not sure. I looked into a bit and found this solution https://stackoverflow.com/a/73853553
```
Move the id "kotlin-kapt" to the bottom of plugins{} in build.gradle Module-level.
```
it makes little sense to me still, but I suppose this is more correct?

appcues/consumer-rules.pro
```diff
@@ -1,12 +1,19 @@
--keep class com.appcues.data.remote.**.response.** { *; }
+# generated Moshi adapaters
+-keep class com.appcues.**.*JsonAdapter { *; }
```
iujames, line 2: This is revised to be more generic, to cover the existing `data.remote` usage, as well as the root level com.appcues.ViewElementJsonAdapter - does this seem correct?

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
```diff
@@ -153,11 +154,12 @@ private fun View.asCaptureView(screenBounds: Rect): ViewElement? {
             val semanticsOwner = semanticsOwnerField.get(this) as SemanticsOwner
             val composeChildren = listOf(semanticsOwner.rootSemanticsNode.asCaptureView(context))
             children.addAll(composeChildren)
-        } catch (_: Exception) {
+        } catch (ex: Exception) {
             // Catching and swallowing exceptions here with the Compose view handling in case
             // something changes in the future that breaks the expected structure being accessed
             // through reflection here. If anything goes wrong within this block, prefer to continue
             // processing the remainder of the view tree as best we can.
+            Log.e("Appcues", "error processing Compose layout, ${ex.message}")
```
iujames, line 162: There is really not a clear access to Logcues at this point, and the targeting strategy is kept in a static spot in `Appcues.elementTargeting`. We could add a one-off log message here to provide some feedback in the case of an error like what we've seen recently. I can also remove this if undesired to use Logcat directly like this, outside of our normal logging structure.

appcues/consumer-rules.pro
andretortolano: yea it looks correct

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
andretortolano: i see no other way to log possible erros here, unless we make the strategy non-static

andretortolano (approved):

PR #534, opened 2023-11-07
Ensure that compose elements are within screen bounds

iujames:
Previously, we were not confirming that Compose view rects intersected the bounds of the visible screen, which is corrected here. Also, we did not have any policy around how much of a target element (Android View or Composable) needed to be visible, to render targeted content (i.e. have a selector). Now, we have a policy in place that at least the center point of the view must be visible.

This example illustrates the handling before and after. This is a view with a horizontal scrolling composable section at the bottom, as well as a horizontal pager. Prior to this change, as elements are laid out off-screen in preparation for a UI gesture to bring them into view, we were recognizing them in the view hierarchy and generating selectors. This led to snapshots where tooltips could be targeted at elements that were mostly or completely off the screen. Now, only the elements that have their center point within the visible viewport are eligible for a target element experience.

| before | after |
| --- | --- |
|  Screenshot 2023-11-07 at 12 29 54 PM | Screenshot 2023-11-07 at 12 26 44 PM |


andretortolano (approved):
👍

PR #535, opened 2023-11-09
Batch group updates with user identification when called right after identify

iujames:
This is a similar update to what was recently done in the web sdk in https://github.com/appcues/javascript-sdk/pull/613

* on `identify(id, props)` or the `appcues:session_started` event - a 50ms timer is started to allow analytics that occur immediately after, to be batched together into the same activity payload to the API.
* the expected use case would be a `group(id, props)` called immediately after `identify(id, props)` resulting in a combined payload with the new user and the current group - this prevents any chance of state group information being used during the resulting qualification on the server for the new user update or session start qualification
* all other analytics are unaffected, unless they also happen within that 50ms window (in which case they would go in this immediate batch as well)

This is an example of what this result would look like, for a new user landing on a login page, signing in, and then triggering the resulting user profile update, the session_started event, and an immediate group update to identify that user's current group and props.
Screenshot 2023-11-08 at 3 27 28 PM

This is implemented by adding a new priority queue to our `AnalyticsQueueProcessor` - the 50ms queue. We still have the 10 sec background batching queue for flow events as well. If a new priority update comes in (identify or session_start) it will flush anything in the queue existing, and start this new 50ms queue to gather up all immediately subsequent updates.

Couple of edge cases:
* There is already a 50ms queue going for another user - could happen if you made multiple `identify(id)` calls for different user ids, in rapid succession. This will result in the existing priority queue being immediately flushed, stopping any existing timer, before processing the new one.
* There is a 50ms queue running while an otherwise immediate event is triggered (a screen or track event) - this will result in the normally immediate event having some delay of < 50ms, as it is merged into this queue (hopefully rare).

### DIFF
--- a/appcues/src/main/java/com/appcues/analytics/AnalyticsQueueProcessor.kt
+++ b/appcues/src/main/java/com/appcues/analytics/AnalyticsQueueProcessor.kt
@@ -17,7 +17,7 @@ internal class AnalyticsQueueProcessor(
     private val repository: AppcuesRepository,
     private val experienceRenderer: ExperienceRenderer,
     // this is the background analytics processing queue - 10 sec batch
-    private val analyticsQueueScheduler: QueueScheduler,
+    private val backgroundQueueScheduler: QueueScheduler,
     // this is the immediate processing queue - 50ms batch to group identify with immediate subsequent updates
     private val priorityQueueScheduler: QueueScheduler,
 ) {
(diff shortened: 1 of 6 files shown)


andretortolano (approved):
I agree with the approach, at the same time its hard to keep track of the queue methods and what they are used for, so good to add extra comments there for context.

appcues/src/main/java/com/appcues/analytics/AnalyticsQueueProcessor.kt
```diff
@@ -16,22 +16,29 @@ internal class AnalyticsQueueProcessor(
     private val appcuesCoroutineScope: AppcuesCoroutineScope,
     private val repository: AppcuesRepository,
     private val experienceRenderer: ExperienceRenderer,
+    // this is the background analytics processing queue - 10 sec batch
     private val analyticsQueueScheduler: QueueScheduler,
+    // this is the immediate processing queue - 50ms batch to group identify with immediate subsequent updates
+    private val priorityQueueScheduler: QueueScheduler,
```
iujames, line 22: @andretortolano it was simpler for review to not rename old things, but I think I might take another pass over just to rename, not change any logic. As I think having the old stuff named `analyticsQueueScheduler` and `pendingActivity` is not really ideal now, maybe something like `backgroundQueueScheduler` and `backgroundQueue` -- I just want to make it more clear why there are two paths of things here and which is background vs priority.

PR #538, opened 2023-11-15
✨ Added group identification row in debugger

andretortolano:
adding extra row in the debugger to identify Group same way we identify userId


iujames (approved):
nice add!

PR #539, opened 2023-11-17
🐛 LaunchExperience should launch different coroutine and unlock processing actions

andretortolano:
During regression testing I found an issue where a post complete action to LaunchExperience would lead to StateMachine getting into a deadlock situation, this happens because LaunchExperience is running on the same suspend function as the handleInternalAction and it gets to a point where both are waiting on each other to move on.

this can be tested by tapping on event2 and invoking [this](https://studio.appcues.com/mobile/flows/bf387041-bf99-43b7-b21f-da76bb56340c/settings) experience

now LaunchExperience launches a new coroutine which unlocks the ActionProcessor.process call, and unlocks the whole transition

### DIFF
--- a/appcues/src/main/java/com/appcues/action/ActionProcessor.kt
+++ b/appcues/src/main/java/com/appcues/action/ActionProcessor.kt
@@ -31,8 +31,7 @@ internal class ActionProcessor(override val scope: AppcuesScope) : AppcuesCompon
         }
     }
 
-    // This is used for internal actions, like navigation actions processed during step transitions, or
-    // post flow completion actions.
+    // This is used for internal actions, like navigation actions processed during step transitions.
     suspend fun process(actions: List<ExperienceAction>) {
         if (actions.isEmpty()) return
 
@@ -42,6 +41,13 @@ internal class ActionProcessor(override val scope: AppcuesScope) : AppcuesCompon
         // make this a suspend function and wait on them, we cannot place them in the queue or
         // suspend the currently executing action at all, or the queue would become deadlocked.
         transformQueue(actions).toMutableList().forEach { it.execute() }
+    }
+
+    // This is used for post flow actions, ensures we never get to a deadlock state with the current transition
+    // being processed by the state machine. Mainly because of LaunchExperienceAction when we have another
+    // experience we would want to trigger when current flow completes
+    suspend fun processPostFlowActions(actions: List<ExperienceAction>) {
+        transformQueue(actions).forEach { actionQueue.send(it) }
     }
 
     // This version is used by the viewModel to process interactive actions from user input - button taps.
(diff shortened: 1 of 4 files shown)


appcues/src/main/java/com/appcues/action/appcues/LaunchExperienceAction.kt
```diff
@@ -45,7 +50,9 @@ internal class LaunchExperienceAction(
 
     override suspend fun execute() {
         if (experienceId != null) {
-            experienceRenderer.show(experienceId, getTrigger())
+            coroutineScope.launch {
```
andretortolano, line 53: here is the important change, basically launch a new scope before calling show(..)

appcues/src/test/java/com/appcues/action/appcues/LaunchExperienceActionTest.kt
```diff
@@ -23,6 +26,11 @@ internal class LaunchExperienceActionTest : AppcuesScopeTest {
     @get:Rule
     override val scopeRule = TestScopeRule()
 
+    @get:Rule
+    val dispatcherRule = MainDispatcherRule()
```
andretortolano, line 30: tests now require MainDispatcherRule since we are using AppcuesCoroutienScope

iujames (approved):
nice catch and fix!

appcues/src/main/java/com/appcues/statemachine/effects/ExperienceActionEffect.kt
```diff
@@ -8,7 +8,7 @@ import com.appcues.statemachine.SideEffect
 internal data class ExperienceActionEffect(private val actions: List<ExperienceAction>) : SideEffect {
 
     override suspend fun launch(processor: ActionProcessor): Action? {
-        processor.process(actions)
+        processor.processAsync(actions)
```
iujames, line 11: maybe just add a comment here for our future selves noting (1) this effect is only used for processing the post-flow-completion actions AFTER the experience has transitioned to Idle and (2) they must kick off their own new coroutine so that the state machine can complete the current transition, unblocked, and return back to the resting/idling state and release the mutex held on the current action processing instance (the current flow completion)

appcues/src/main/java/com/appcues/action/ActionProcessor.kt
```diff
@@ -44,6 +44,11 @@ internal class ActionProcessor(override val scope: AppcuesScope) : AppcuesCompon
         transformQueue(actions).toMutableList().forEach { it.execute() }
     }
 
+    // Calls Process on a different Job to unlock existing thread
```
iujames, line 47: and a comment here to say this is the version used by post flow completion actions - and update the comment on L34-35 please to remove the reference to post flow completion actions, as that one is only used by the navigation actions now.

PR #540, opened 2023-11-21
🐛 Properly handle link actions for invalid deep links

andretortolano:
bug found during exploration test for the 3.1.5 release, its very low severity since its an edge case, but for future versions we can handle invalid deeplinks, we can also display a log error in the debugger so customers can figure out why they are not navigating properly

image

### DIFF
--- a/appcues/src/main/java/com/appcues/action/appcues/LinkAction.kt
+++ b/appcues/src/main/java/com/appcues/action/appcues/LinkAction.kt
@@ -1,5 +1,6 @@
 package com.appcues.action.appcues
 
+import android.content.ActivityNotFoundException
 import android.net.Uri
 import com.appcues.Appcues
 import com.appcues.action.ExperienceAction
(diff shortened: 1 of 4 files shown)


appcues/src/main/java/com/appcues/action/appcues/LinkAction.kt
```diff
@@ -42,10 +46,12 @@ internal class LinkAction(
 
         if (isEligibleCustomTab) {
             linkOpener.openCustomTabs(uri)
-        } else {
+        } else try {
             // this will handle any in-app deep link scheme URLs OR any web urls that were
             // requested to open into the external browser application
             appcues.navigationHandler?.navigateTo(uri) ?: linkOpener.startNewIntent(uri)
+        } catch (exception: ActivityNotFoundException) {
```
iujames, line 53: should it catch any Exception at this level? The `ActivityNotFoundException` is specific to the `startNewIntent` implementation, but in theory, if they have a custom `navigationHandler` registered, they may throw some other Exception type and we can guard against that too?

iujames (approved):
Fix looks great, just a question on possible making more general exception handling.

Nice work adding tests

appcues/src/main/java/com/appcues/action/appcues/LinkAction.kt
andretortolano: I would say if they do their own implementation they need to handle it themselves, at least here we can avoid generic exception

iujames: Ok, sounds reasonable to me, thanks.

PR #543, opened 2023-11-28
🐛 Invalidate current step and re-present onConfigurationChanged

andretortolano:
https://app.shortcut.com/appcues/story/59914/anchored-tooltip-positioning-issue-on-rotation-to-landscape

During investigation it was discovered that sometimes we are also dismissing the current Experience, if for some reason when configuration changes we hit the onActivityChanged callback from ActivityMonitor (1), and then when the experience is able to persist through the configuration changing the current state of the state machine was the same was before (2) this is why we see this anchored tooltip issue in the first place.

So:

1 - ActivityMonitor now better differentiate between activity changes and configuration changes (new callback method too)
2 - Introduced new action `RestartStep` (open to name suggestions) that we can send to RenderingStepState to move to BeginningStepState and force the re-presentation through PresentationEffect (which will produce a new set of metadata)

### DIFF
--- a/appcues/src/main/java/com/appcues/monitor/AppcuesActivityMonitor.kt
+++ b/appcues/src/main/java/com/appcues/monitor/AppcuesActivityMonitor.kt
@@ -10,11 +10,15 @@ internal object AppcuesActivityMonitor : Application.ActivityLifecycleCallbacks 
     interface ActivityMonitorListener {
 
         fun onActivityChanged(activity: Activity)
+
+        fun onConfigurationChanged(activity: Activity)
     }
 
     private var _isPaused = false
     val isPaused: Boolean
         get() = _isPaused
+
+    private var notifyConfigurationChanged = false
 
     private var activityWeakReference: WeakReference<Activity>? = null
     val activity: Activity?
@@ -42,15 +46,32 @@ internal object AppcuesActivityMonitor : Application.ActivityLifecycleCallbacks 
 
     override fun onActivityResumed(activity: Activity) {
         _isPaused = false
-        if (this.activity != activity) {
-            this.activityWeakReference = WeakReference(activity)
 
+        if (this.activity == null) {
+            this.activityWeakReference = WeakReference(activity)
             // notifies all subscribers that activity has changed
             activityMonitorListener.forEach { it.onActivityChanged(activity) }
         }
+
+        if (notifyConfigurationChanged) {
+            // notifies all subscribers that activity has changed
+            activityMonitorListener.forEach { it.onConfigurationChanged(activity) }
+            notifyConfigurationChanged = false
+        }
     }
 
-    override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) = Unit
+    override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
+        // whenever we hit onActivityCreated + savedInstanceState is not null
+        // it means configuration has changed (same activity was re-created
+        if (savedInstanceState != null) {
+            notifyConfigurationChanged = true
+        } else {
+            if (this.activity != activity) {
+                this.activityWeakReference = null
+            }
+        }
+    }
+
     override fun onActivityStarted(activity: Activity) = Unit
     override fun onActivityPaused(activity: Activity) {
         _isPaused = true
--- a/appcues/src/main/java/com/appcues/statemachine/states/RenderingStepState.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/states/RenderingStepState.kt
@@ -5,12 +5,14 @@ import com.appcues.data.model.StepReference.StepOffset
 import com.appcues.statemachine.Action
 import com.appcues.statemachine.Action.EndExperience
 import com.appcues.statemachine.Action.MoveToStep
+import com.appcues.statemachine.Action.Reset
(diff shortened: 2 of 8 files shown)


appcues/src/main/java/com/appcues/ui/StateMachineDirectory.kt
```diff
@@ -33,9 +39,16 @@ internal class ModalStateMachineOwner(override val stateMachine: StateMachine) :
     override suspend fun reset() {
         stateMachine.stop(true)
     }
+
+    override suspend fun onConfigurationChanged() {
+        if (stateMachine.state is RenderingStepState) {
+            stateMachine.handleAction(Reset)
+        }
+    }
```
iujames, line 47: following along - this is only handled in the ModalStateMachineOwner (modals and tooltips) as we reset the overlay content with the customer app changes config. With embeds, not needed, as those live within a customer view that has already been reset, presumably, and controls the layout of our composition within it.

appcues/src/main/java/com/appcues/ui/presentation/ViewPresenter.kt
```diff
@@ -47,6 +47,10 @@ internal abstract class ViewPresenter(
         override fun onActivityChanged(activity: Activity) {
             viewModel?.onActivityChanged()
         }
+
+        override fun onConfigurationChanged(activity: Activity) {
+            viewModel?.onConfigurationChanged()
```
iujames, line 52: the `ViewPresenter` already has the `experienceRenderer` and the `renderContext`, do we need to send this through the `viewModel` at all? thinking we could short circuit that and just do 

```kotlin
coroutineScope.launch {
    experienceRenderer.onViewConfigurationChanged(renderContext)
}
```

right here?

appcues/src/main/java/com/appcues/statemachine/states/RenderingStepState.kt
```diff
@@ -31,6 +33,9 @@ internal data class RenderingStepState(
             } else {
                 toEndingStep(action)
             }
+            is Reset -> {
+                toBeginningStep()
+            }
```
iujames, line 38: seems ok to use existing `Reset` here to me. The PR description has some older info about `RestartStep`. Would be good to update and clean that up.

Maybe in the future it would use `Retry` from [here](https://github.com/appcues/appcues-android-sdk/pull/528/files#diff-a4ce2dcad8b214bb7929fd1f82a3c39b079064cd08a542c9fa209c5b8df0c3c4R19)? (or vice versa, not sure)

appcues/src/main/java/com/appcues/monitor/AppcuesActivityMonitor.kt
```diff
@@ -42,15 +46,32 @@ internal object AppcuesActivityMonitor : Application.ActivityLifecycleCallbacks
...
+            activityMonitorListener.forEach { it.onConfigurationChanged(activity) }
+            notifyConfigurationChanged = false
+        }
+    }
+
+    override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
+        // whenever we hit onActivityCreated + savedInstanceState is not null
+        // it means configuration has changed (same activity was re-created
+        if (savedInstanceState != null) {
+            notifyConfigurationChanged = true
```
iujames, line 67: this is the only part I really don't know definitively enough about to know if this is always going to work, hopefully so. I wish there were some more direct lifecycle callback specifically for configuration change, but maybe that just doesn't exist at the global lifecycle callback listener level.

iujames (approved):
awesome to have better handling on rotation! This seems like it would be a great future candidate for a UI test in the snapshot repo with Roborazzi, if possible.

appcues/src/main/java/com/appcues/ui/presentation/ViewPresenter.kt
andretortolano: it is important it goes through the viewModel because the viewModel is the one getting nulled when the experience is dismissed/removed from screen permanently. probably wont make much of difference but this way we ensure that we are only reporting configuration changed when we should for a given renderContext.

appcues/src/main/java/com/appcues/statemachine/states/RenderingStepState.kt
andretortolano: yeah we ok use Retry in the future, both naming here could be confusing for whats actually doing

appcues/src/main/java/com/appcues/monitor/AppcuesActivityMonitor.kt
andretortolano: there is actually a method inside that interface called:
```kotlin

        /**
         * Called when the Activity configuration was changed.
         * @hide
         */
        default void onActivityConfigurationChanged(@NonNull Activity activity) {
        }
```
        
        but I'm not able to override it for some reason

appcues/src/main/java/com/appcues/ui/presentation/ViewPresenter.kt
iujames: ah ok, so I guess it could null check `viewModel` here? or really it wont matter since the `Reset` action would do nothing if it was in Idling state? but this current implementation seems fine too - I was just thinking it may possibly be able to be simplified. I don't think it really needs to change though if this is most clear.

PR #544, opened 2023-11-29
🐛 Fix screenshot for API 26+ when target view that  doesn’t support hardware bitmaps.

andretortolano:
Customer was unable to take screenshot on a full compose app. I was able to reproduce by installing Appcues into [NowInAndroid](https://github.com/android/nowinandroid). the error message suggest that the method we were using for taking screenshots is probably not the most recommended, so for API 26+ I switched to an approach that uses PixelCopy.

+ When I changed to PixelCopy I noticed that the screenshot was being taken on the mainThread (probably) causing the visibility animation of the FAB to lag for some frames, so I changed our internal logic, we now wait for the fab to dismiss before taking the screen shot, this also means that we don't need to "prepare" and "restore" anymore as the FAB is no longer there.

Tested on API 22, 25, 26, 32

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
+++ b/appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
@@ -31,12 +31,10 @@ import com.appcues.logging.LogMessage
 import com.appcues.ui.ExperienceRenderer
 import com.appcues.util.ResultOf.Failure
 import com.appcues.util.ResultOf.Success
-import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.withContext
 
 internal class DebuggerViewModel(override val scope: AppcuesScope, debugMode: DebugMode) : ViewModel(), AppcuesComponent {
 
@@ -156,10 +154,15 @@ internal class DebuggerViewModel(override val scope: AppcuesScope, debugMode: De
         }
     }
 
-    fun onDismissAnimationCompleted() {
+    fun onDismissAnimationCompleted(debuggerState: MutableDebuggerState) {
         if (_uiState.value is Dismissing) {
             _uiState.value = Dismissed(uiState.value.mode)
             viewModelScope.cancel()
+        }
+
+        // fab is completely dismissed, we can take screenshot
+        if (_uiState.value is Expanded && _uiState.value.mode is ScreenCapture) {
+            captureScreen(debuggerState)
         }
     }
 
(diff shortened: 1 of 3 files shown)


appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
```diff
@@ -156,11 +154,16 @@ internal class DebuggerViewModel(override val scope: AppcuesScope, debugMode: De
...
+    fun onDismissAnimationCompleted(debuggerState: MutableDebuggerState) {
         if (_uiState.value is Dismissing) {
             _uiState.value = Dismissed(uiState.value.mode)
             viewModelScope.cancel()
         }
+
+        // fab is completely dismissed, we can take screenshot
+        if (_uiState.value is Expanded && _uiState.value.mode is ScreenCapture) {
+            captureScreen(debuggerState)
+        }
```
iujames, line 166: I'm assuming this all probably works fine, but I'm having a hard time following exactly really - this seems pretty subtle now. Let me see if my understanding is correct

1. when screen capture starts, it hides the FAB, this happens on L179 in DebuggerComposition - however, this no longer directly triggers the VM to sart the screen capture
2. here, on L164, we check for the dismissal complete, and if in a specific state (expanded in ScreenCapture mode) - we trigger the capture, a side-effect of the dismissal at this point
3. additional debuggerState is passed through the dismissal callbacks to make that available here for `captureScreen(state)`
4. we no longer need prepare(view) in ScreenCaptureProcessor, since it is already hidden. restore is removed as well - is this an issue? When does the FAB come back after the screen capture? The previous logic was to return to the camera icon state so you could take more captures.

iujames (commented):
The update to PixelCopy seems ideal. There seems to be some extra added complexity here around handling the dismissal of the debugger view during the snapshot. Maybe that is unavoidable - wanted to check and make sure I understood that part and see if there was any way to simplify.

appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
andretortolano: 1,2 ,3 - you go it

4. you can still do it, the prepare and restore was used to  basically remove the view for the screen capturing, but what is actually returning it to the FAB (Idle) is what happens after the dialog is dismissed.


https://github.com/appcues/appcues-android-sdk/assets/5244805/61b8721f-2c5e-4cc1-86bf-b8ced36a5e1d

iujames: ok next question i would have - does the `R.id.appcues_debugger_view` show up in the screen capture layout structure now? previously, that view was hidden so it would get skipped in the layout hierarchy capture. But now it is still there right? just the composition within it is dismissed?

iujames (approved):
definitely want to make this work for all customers, so no issue with the progress here. I find the new flow a bit subtle/confusing to follow. Maybe just some extra comments in there to indicate what is going on when the capture is taken, and how the dismissal flows into the capture would help in the future if/when we return to this.

PR #548, opened 2023-12-08
Fix issue with screen capture not working in some scenarios

iujames:
It was observed that screen capture could fail sometimes, seemingly randomly, and the log message seen was
```
java.lang.IllegalArgumentException: Window doesn't have a backing surface!
```

The error was seen on a tablet device, using rotation to landscape as well during capture (important later). This message is coming from a call to PixelCopy, which was [introduced recently](https://github.com/appcues/appcues-android-sdk/pull/544) in the 3.1.7 version. 

However, further research seems to show that the root cause was actually a bad current `Activity` reference, from the `AppcuesActivityMonitor`, due to some other recent 3.1.7 changes in https://github.com/appcues/appcues-android-sdk/pull/543. Specifically, on a rotation change, the `onActivityCreated` handler would not set the current activity to `null`, thus when it went into `onActivityResumed`, it was not updating the current activity with the reference to the new (different) Activity being passed in. This would eventually lead to a bad activity reference in the screen capture code, after a rotation, and then the error message seen above.

Three small related fixes here:
1. Ensure we track the current `Activity` correctly at all times in the `AppcuesActivityMonitor` (root cause)
2. Add some more exception handling + fallback in the bitmap capture logic - now, if the PixelCopy attempt fails, we'll at least try to capture using the older style method (creating a Bitmap and having a View draw to it). This can also fail, so catching that exception now as well
3. Fix the selector preview display on landscape captures - just a UI bug noticed as well during this.

The selector displays now look correct in both orientations (ignore the missing dialog backdrop here, thats just an emulator rendering issue, known).

| portrait | landscape |
| --- | --- |
|  Screenshot 2023-12-08 at 2 57 24 PM | Screenshot 2023-12-08 at 2 55 02 PM |

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/screencapture/GetCaptureUseCase.kt
+++ b/appcues/src/main/java/com/appcues/debugger/screencapture/GetCaptureUseCase.kt
@@ -61,55 +61,69 @@ internal class GetCaptureUseCase {
     // default capture screenshot when the current strategy returns null
     private suspend fun View.screenshot(window: Window): Screenshot? {
         return if (this.width > 0 && this.height > 0) {
-            val bitmap = awaitCaptureBitmap(this, window)
-            val canvas = Canvas(bitmap)
-            this.draw(canvas)
+            awaitCaptureBitmap(this, window)?.let {
+                val canvas = Canvas(it)
+                this.draw(canvas)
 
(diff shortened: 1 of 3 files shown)


mmaatttt (approved):

iujames:
> Should there be a shadow on the capture confirmation modal in your screenshots?

yeah, there is normally - this is some emulator artifact that we ran into a while back, and maybe only on my machine? but we've confirmed that it is not an issue on an real device.

PR #549, opened 2023-12-12
♻️  Replace current colors with LocalAppcuesTheme

andretortolano:
Currently our colors are spread all over the code, this is terrible way to manage themes and impossible to maintain light and dark mode, so this proposed approach defines a real AppcuesTheme that contains all abstract "objects" of interest like primary, secondary, background, etc... based on [Appcues branding guidelines](https://www.figma.com/file/8Sozh9JnseLTF5oWjvGfIp/Foundations?type=design&node-id=9-59&mode=design&t=56CipGcX1bXwpUOW-0). Some colors have changed from the original implementation since those were not following any guidelines, but we can surely go back to a specific color if it doesn't look right for some reason.

* Updated [experience-specs](https://github.com/appcues/appcues-mobile-experience-spec/pull/244)

| Screen                   | Light Mode Screenshot                                  | Dark Mode Screenshot                                   |
|--------------------------|--------------------------------------------------------|--------------------------------------------------------|
| Debugger (Main)          | Light Debugger | Dark Debugger |
| Debugger (Fonts)         | Light Debugger | Dark Debugger |
| Debugger (Event)         | Light Debugger | Dark Debugger |
| ScreenCapture            | Light ScreenCapture | Dark ScreenCapture |
| ScreenCapture (Error)    | Light ScreenCapture (Error) | Dark ScreenCapture (Error) |


appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
```diff
@@ -44,7 +44,7 @@ public fun ComposeContent(json: String, imageLoader: ImageLoader) {
     val response = MoshiConfiguration.moshi.adapter(PrimitiveResponse::class.java).fromJson(json)
     val primitive = response!!.mapPrimitive()
 
-    AppcuesTheme {
+    AppcuesExperienceTheme {
```
andretortolano, line 47: Open to suggestions for naming here, but I wanted to have AppcuesTheme being the theme that respects our branding colors, the Theme we have for experience is just clearing some colors from MaterialTheme so we can use Composable like the Surface and Text.

appcues/src/main/java/com/appcues/ui/theme/AppcuesColors.kt
```diff
@@ -1,23 +0,0 @@
-package com.appcues.ui.theme
-
-import androidx.compose.ui.graphics.Color
-
-internal object AppcuesColors {
```
andretortolano, line 5: Deleted since accessing colors directly made it impossible to switch between light and dark mode

iujames:
This is looking really nice, nice job! At a very high level glancing over the screenshots, the only question I have is the dark mode foreground color, when on top of a filled button or colored background. So, for example, the text and button outline on the failure toast - the toast is pink, so should the text and outline remain white in dark mode? Same thing for the "send to builder" filled button on the screen capture dialog.

andretortolano:
I'm going over small increments I will add to another commit so we can clearly see, but I think the idea is that in darkmode those colors are prepared to contrast with BLACK instead of WHITE. but can confirm with sam

appcues/src/debug/java/com/appcues/CompositionTestHelper.kt
iujames: this naming makes sense to me

appcues/src/main/java/com/appcues/debugger/ui/ds/ButtonComponents.kt
iujames, line 1: ds package for design system?

appcues/src/main/java/com/appcues/debugger/ui/theme/AppcuesThemeColors.kt
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/**
+ * This is official Appcues Theme that applies all appcues branding colors
+ * Not used by Experiences since we don't want to impose our colors in any experience,
+ *
+ * ONLY USE THIS FOR APPCUES SPECIFIC COMPOSITIONS LIKE OUR FAB(Floating Action Button)
+ */
+@Composable
+internal fun AppcuesTheme(content: @Composable () -> Unit) {
```
iujames, line 101: maybe pull this one out to its own file? I had trouble finding it and then saw it was in the colors file

appcues/src/main/java/com/appcues/debugger/ui/DebuggerPanel.kt
```diff
@@ -66,7 +66,7 @@ internal fun BoxScope.DebuggerPanel(debuggerState: MutableDebuggerState, debugge
         Spacer(
             modifier = Modifier
                 .fillMaxSize()
-                .background(AppcuesColors.DebuggerBackdrop)
+                .background(Color(color = 0x51000000))
```
iujames, line 69: should this be using a theme color?

iujames (approved):
Looks great! nice work, couple minor comments

appcues/src/main/java/com/appcues/debugger/ui/ds/ButtonComponents.kt
andretortolano: yes

appcues/src/main/java/com/appcues/debugger/ui/theme/AppcuesThemeColors.kt
andretortolano: will do

appcues/src/main/java/com/appcues/debugger/ui/DebuggerPanel.kt
andretortolano: I had it in there but I think this will never differ from dark to light mode or any other theme so decided to just hardcode since its black with some transparency

PR #550, opened 2023-12-14
♻️  Including below the fold target element

andretortolano:
Including extra lines and a new element we can target below the fold.

image


iujames:
What do you think of adding a page with a recycler view, similar to what was done on iOS here https://github.com/appcues/appcues-ios-sdk/pull/481

This is what it looks like, with 5 targetable buttons in scroll:

https://github.com/appcues/appcues-android-sdk/assets/19266448/cdc0676a-233d-4e5c-9791-6616ec205ebb

I think it would be ideal to have the test apps in parity as close as possible for testing.

andretortolano:
Closing this one, going for a different approach using recyclerView in a different page.

PR #551, opened 2024-01-08
Update identity verification note to point to central help doc

iujames:
Per recent customer feedback and [discussion](https://appcues.slack.com/archives/C031RLGS4F8/p1704495749255789) - moving to an approach where we do not provide a code snippet including `"appcues:user_id_signature"` that could mistakenly be copy/pasted, and rather a less explicit pointer to the main help doc for identity verification - for the less common use case where those details are needed.


andretortolano (approved):
not sure why branch os not green, probably just random failure on the jobs, not relevant

PR #554, opened 2024-01-24
♻️  Adding Query params for deeplink preview and show

andretortolano:
matching iOS https://github.com/appcues/appcues-ios-sdk/pull/489.

Enabling query params from builder deeplinks (preview and show) to be used by de SDK when calling the end points for preview and show.

this use case is currently aiming to solve the localeId properly related to localization feature.

### DIFF
--- a/appcues/src/main/java/com/appcues/data/model/Experience.kt
+++ b/appcues/src/main/java/com/appcues/data/model/Experience.kt
@@ -27,6 +27,7 @@ internal data class Experience(
     val requestId: UUID? = null,
     val error: String? = null,
     var renderErrorId: UUID? = null,
+    var previewQuery: Map<String, String> = mapOf(),
 ) {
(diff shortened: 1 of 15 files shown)


appcues/src/main/java/com/appcues/data/model/Experience.kt
```diff
@@ -27,6 +27,7 @@ internal data class Experience(
     val requestId: UUID? = null,
     val error: String? = null,
     var renderErrorId: UUID? = null,
+    var previewQuery: Map<String, String> = mapOf(),
```
andretortolano, line 30: I feel like this keeps it all more organized, so when on ViewPresenter we want to call refreshPreview we can just grab both the ID and previewQuery from the experience.

iujames (approved):
looking good!

PR #546, opened 2023-12-01
♻️ Remove CustomerAPI Interceptor

andretortolano:
* Remove CustomerAPI Interceptor
* Streamline screen capture process and decouple from the old "Capture" object into more purposeful objects

1. CustomerAPIService was dependent on an HttpInterceptor to build the request to the correct URL, this is unnecessary as we can just pass in as a parameter the url, maybe a little more verbose but brings simplicity and make so this whole logic is easier to test.

2. Created specific classes to describe HttpRequests in isolation.

3. Broken down ScreenCaptureProcessor into two UseCase classes for separation of concern + testability

PR #556, opened 2024-02-01
Fix screen capture bitmap handling with hardware accelerated views

iujames:
I believe this will fix the issue seen recently with apps using HW accelerated views trying to take screen captures. The error observed was:
```
java.lang.Exception: java.lang.IllegalArgumentException: Software rendering doesn't support hardware bitmaps
```

This error was coming from the lines removed in this PR.

What I believe happened was
* A fix was made in https://github.com/appcues/appcues-android-sdk/pull/544 that corrected the behavior, and worked, in version 3.1.7
* Some refactoring was then done in https://github.com/appcues/appcues-android-sdk/pull/546 and when the `GetCaptureUseCase` was created, a couple of lines were accidentally copied over into this spot where they should not be. These lines are from the `captureLegacyBitmap` implementation - which should only be used on devices < API 26, or as a fallback when PixelCopy fails to try optimistically to capture a view with no hardware acceleration. This reverted the fix in version 3.1.8 (and 3.1.9 current)

This change should get it back to the expected structure for HW accelerated cases.


andretortolano (approved):
💯

PR #557, opened 2024-02-02
✨ Adding _lastSeenAt autoProp to identify and event analytics

andretortolano:
Keeping in sync with Web Frontend change to include _lastSeenAt as _updatedAt is deprecated but is still used for targeting.

included:
* changes to autoPropertyDecorator.
* change to Debugger for recent events to use new property
* changes to unit tests

### DIFF
--- a/appcues/src/main/java/com/appcues/analytics/AutoPropertyDecorator.kt
+++ b/appcues/src/main/java/com/appcues/analytics/AutoPropertyDecorator.kt
@@ -22,7 +22,7 @@ internal class AutoPropertyDecorator(
     companion object {
 
         const val IDENTITY_PROPERTY = "_identity"
-        const val UPDATED_AT_PROPERTY = "_updatedAt"
+        const val LAST_SEEN_AT = "_lastSeenAt"
     }
 
     private var currentScreen: String? = null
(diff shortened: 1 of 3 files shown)


appcues/src/main/java/com/appcues/analytics/AutoPropertyDecorator.kt
```diff
@@ -55,7 +54,9 @@ internal class AutoPropertyDecorator(
             "userId" to storage.userId,
             "_isAnonymous" to storage.isAnonymous,
             "_localId" to storage.deviceId,
-            UPDATED_AT_PROPERTY to Date(),
+            "_updatedAt" to Date(),
+            // Last Seen At will deprecate _updatedAt which can't be entirely removed today since it's used for targeting
+            "_lastSeenAt" to Date(),
```
iujames, line 59: nit, but might be best to have a shared `now` variable that is used in both spots, so they are guaranteed to have identical value, like https://github.com/appcues/appcues-ios-sdk/pull/496/files#diff-ded0c79ea6f55775f64fabb6e0cb83e0d69f4cd0a0912523386714135db2450dR69

iujames (approved):

PR #558, opened 2024-02-08
🐛 Fix to React Native RTL

andretortolano:
Experience was not showing because the composition was losing state, I still did not figure out what exactly made so the experience was stuck, but it was related to how React Native keeps "refreshing" the layout direction, so by adding LayoutDirection.current to the main key clause we ensure to re-compose from scratch when a new value is passed in.


iujames (approved):
It makes sense for the content alignment. I don't quite understand how it would make an experience be non-visible, but if it tests out and works ok 👍 nice find an fix!

test failure looks like that flakey analytics test, unrelated https://app.circleci.com/pipelines/github/appcues/appcues-mobile-experience-spec/1109/workflows/d704b8bf-05dd-41cc-84a5-f1e752707c44/jobs/1352

andretortolano:
what was observed is that the Composition is losing state, I imagine the flags "contentVisible" and "backdropVisible" makes so the experience never shows. on further testing I was able to render single compositions in the same scenario, so its an issue related to our Composition and its states for sure.

+ the button werent working, so adding the key ensures a clean draw where everything works

PR #562, opened 2024-02-15
⬆️  Update compose from 1.5 to 1.6

andretortolano:
Migrating from compose 1.5 to 1.6

it was necessary to remove the MaterialTheme since they constraint the line height to a proper font style, since we dynamically decide on the font size and possibly custom font height, using MaterialTheme as our overall theme is no longer an option.

* Pending to figure out if there is a way to minimize the differences in text display from the material theme and the standard theme.

### DIFF
--- a/appcues/build.gradle
+++ b/appcues/build.gradle
@@ -11,7 +11,7 @@ apply from: 'copy-dependencies.gradle'
 apply from: 'dokka.gradle'
 
 ext.room_version = '2.5.2'
-ext.compose_version = '1.6.0'
+ext.compose_version = '1.6.2'
 ext.compose_compiler_version = '1.4.8'
 ext.coil_version = '2.4.0'
 ext.moshi_version = '1.15.0'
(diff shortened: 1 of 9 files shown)


appcues/build.gradle
```diff
@@ -11,7 +11,7 @@ apply from: 'copy-dependencies.gradle'
 apply from: 'dokka.gradle'
 
 ext.room_version = '2.5.2'
-ext.compose_version = '1.5.2'
+ext.compose_version = '1.6.0'
```
iujames, line 14: any value in us moving to the BOM approach for the compose packages at this point?

iujames (commented):
looking good so far, appreciate you working through these. Will wait and see how things look with the text sizing. Would be nice if we can make these changes with no significant visual difference detected in flow rendering.

appcues/build.gradle
andretortolano: I'm old school, I rather know which version we are on, also probably makes it easier for customers who are consulting our code maybe ? not sure

andretortolano:
+ Updated to 1.6.1 which includes some bug fixes https://developer.android.com/jetpack/androidx/releases/compose-foundation#1.6.1

iujames (approved):
🚀

PR #565, opened 2024-02-22
✨ Adding Appcues.setPushToken method

andretortolano:
Adding new API call to register push token and set it in Storage (shared preferences)

### DIFF
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -264,6 +264,15 @@ public class Appcues internal constructor(internal val scope: AppcuesScope) {
     }
 
     /**
+     *  Provide the Firebase Cloud Messaging (FCM) device token to Appcues.
+     *
+     *  @param token A globally unique token that identifies this device to FCM.
+     */
+    public fun setPushToken(token: String?) {
+        storage.pushToken = token
+    }
+
+    /**
      * Enables automatic screen tracking for Activities.
      */
     public fun trackScreens() {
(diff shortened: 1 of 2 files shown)


appcues/src/main/java/com/appcues/Storage.kt
```diff
@@ -13,6 +13,7 @@ internal class Storage(
 
     private enum class Constants(val rawVal: String) {
         DeviceId("appcues.deviceId"),
+        DeviceToken("appcues.deviceToken"),
```
andretortolano, line 16: I was referring to the existing reference from the  Hackaton but I was wondering if it would be clearer to name this PushToken

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -263,6 +263,15 @@ public class Appcues internal constructor(internal val scope: AppcuesScope) {
         }
     }
 
+    /**
+     *  Provide the APNs device token to Appcues.
+     *
+     *  @param token A globally unique token that identifies this device to APNs.
```
iujames, line 269: should update this comment to reflect Firebase Cloud Messaging (FCM) rather than APNs

appcues/src/main/java/com/appcues/Storage.kt
iujames: yes, lets please name this `PushToken` and use `pushToken` on L33. Landed on that consensus in chat I believe, and matches iOS side https://github.com/appcues/appcues-ios-sdk/blob/sdk4/Sources/AppcuesKit/Data/Storage.swift#L31

iujames (approved):
couple small touch ups but otherwise looks good

PR #566, opened 2024-02-22
✨ Adding _pushToken property

andretortolano:
* Adding "_pushToken" to device properties
* Improving visual representation of the "_device" property set for debugger.

### DIFF
--- a/appcues/src/main/java/com/appcues/analytics/AutoPropertyDecorator.kt
+++ b/appcues/src/main/java/com/appcues/analytics/AutoPropertyDecorator.kt
@@ -51,15 +51,16 @@ internal class AutoPropertyDecorator(
         "_deviceModel" to contextWrapper.getDeviceName(),
     )
 
-    private var deviceProperties = hashMapOf<String, Any>(
-        "_deviceId" to storage.deviceId,
-        "_language" to contextWrapper.getLanguage()
-        // token information on comes later on future task
-        // "_pushToken" to "UUID",
-        // "_pushSubscriptionStatus" to “subscribed”, “opted-in”, “unsubscribed”
-        // "_pushEnabled" to true, false
-        // "_pushEnabledBackground" to true, false
-    )
+    private val deviceProperties: Map<String, Any?>
+        get() = hashMapOf(
+            "_deviceId" to storage.deviceId,
+            "_language" to contextWrapper.getLanguage(),
+            "_pushToken" to storage.pushToken,
+            // token information on comes later on future task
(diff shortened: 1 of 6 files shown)


appcues/src/main/java/com/appcues/analytics/AutoPropertyDecorator.kt
```diff
@@ -100,7 +101,7 @@ internal class AutoPropertyDecorator(
             sessionPageviews += 1
         } else if (event.name == AnalyticsEvent.SessionStarted.eventName) {
             // on session start set _device key
-            attributes[DEVICE_PROPERTY] = hashMapOf<String, Any>().apply {
+            attributes[DEVICE_PROPERTY] = hashMapOf<String, Any?>().apply {
```
iujames, line 104: just double checking - this allows the JSON serializing to send an explicit `null` value for `_pushToken`, when unset, correct?

iujames (approved):

appcues/src/main/java/com/appcues/analytics/AutoPropertyDecorator.kt
andretortolano: yeah good catch, i was only testing on the Recents Events, but for it to carry over to the network call we need to make sure Moshi does serialize it. the simplest one seems to be adding the @SerializeNull to EventRequest attributes, open to ideas.

appcues/src/main/java/com/appcues/debugger/DebuggerRecentEventsManager.kt
```diff
@@ -229,26 +233,29 @@ private fun Map<String, Any>.filterOutScreenProperties(eventType: EventType): Ma
...
 }
 
-private fun Map<String, Any>.filterOutAutoProperties(): Map<String, Any> {
-    return filterNot { it.key == AutoPropertyDecorator.IDENTITY_PROPERTY }
-}
-
-private fun Map<String, Any>.filterOutMetricsProperties(): Map<String, Any> {
-    return filterNot { it.key == SdkMetrics.METRICS_PROPERTY }
+// filter out all internal properties so it doesnt show in "properties" section
+private fun Map<String, Any>.filterOutInternalProperties(): Map<String, Any> {
```
andretortolano, line 237: got detekt error (too many functions) in this file, since we are only using these filterOut in a single place I think its ok to consolidate all of those into one method.

appcues/src/main/java/com/appcues/analytics/AutoPropertyDecorator.kt
andretortolano: image

iujames (approved):

PR #567, opened 2024-02-22
✨ Adding device updated event

andretortolano:
* Adding device_updated event and track it on setPushToken call
* Event is set to track when there is a valid session
* Added Device type (debugger)


iujames:
changed the base of this for review, so it stacks on #566

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -270,6 +271,8 @@ public class Appcues internal constructor(internal val scope: AppcuesScope) {
      */
     public fun setPushToken(token: String?) {
         storage.deviceToken = token
+
+        analyticsTracker.track(AnalyticsEvent.DeviceUpdated.eventName, properties = null, interactive = false, isInternal = true)
```
iujames, line 275: checked and confirmed - this will only track if the session exists (user identified), that check is inside `AnalyticsTracker`

appcues/src/main/java/com/appcues/debugger/ui/DebuggerExt.kt
```diff
@@ -36,7 +36,8 @@ internal fun EventType?.toResourceId(): Int {
 
 internal fun String.toEventType(): EventType = when (this) {
     AnalyticsEvent.ScreenView.eventName -> SCREEN
-    AnalyticsEvent.SessionStarted.eventName -> SESSION
+    AnalyticsEvent.SessionStarted.eventName,
+    AnalyticsEvent.DeviceUpdated.eventName -> SESSION
```
iujames, line 40: looks like on the iOS side there is a new category `device` for this (and upcoming device_unregistered), would be good to keep consistent https://github.com/appcues/appcues-ios-sdk/pull/502/files#diff-142965116d57ab557fe6f255ae2cd3095c91b606e894a2c27b354d7fd491da9bR136

appcues/src/main/java/com/appcues/Appcues.kt
andretortolano: I think that makes sense and its aligned with what matt is doing on iOS

PR #570, opened 2024-03-04
✨ Adding error description for InvalidToken 400 during screen capture

andretortolano:
Adding more descriptive error message when the session token is invalid.

image

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
+++ b/appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
@@ -213,7 +213,7 @@ internal class DebuggerViewModel(override val scope: AppcuesScope, debugMode: De
             // saving a capture is only valid in screen capture mode with token
             is ScreenCapture -> {
                 viewModelScope.launch {
-                    when (saveCaptureUseCase(currentMode.token, capture)) {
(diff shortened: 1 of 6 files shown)


appcues/src/main/java/com/appcues/debugger/DebuggerViewModel.kt
```diff
@@ -226,7 +227,8 @@ internal class DebuggerViewModel(override val scope: AppcuesScope, debugMode: De
                                 onRetry = {
                                     _toastState.value = ToastState.Idle
                                     onScreenCaptureConfirm(capture)
-                                }
+                                },
+                                errorCode = if (response.reason is HttpError) response.reason.code else null
```
iujames, line 231: would this return any 400 received from any of the endpoints used in the chain of network calls to save a capture? Ideally, would want to have something more specific to the customer api call if possible? I think that endpoint is used on 2 of the 4 calls maybe if memory serves?

andretortolano: made so each step in there is unique when we get the exception, and then we can have particular checks to see what type of error that is.

mapping to an enum its so we dont have a lot of logic in the composition

iujames (approved):

PR #572, opened 2024-03-06
✨ Adding _pushPrimerEligible property

andretortolano:


appcues/src/main/java/com/appcues/analytics/AutoPropertyDecorator.kt
```diff
@@ -77,8 +81,19 @@ internal class AutoPropertyDecorator(
...
 
+    private fun getPushPrimerEligible(): Boolean {
+        val activity = AppcuesActivityMonitor.activity ?: return false
+        // VERSION guard because of Manifest.permission.POST_NOTIFICATIONS
+        return if (VERSION.SDK_INT >= VERSION_CODES.TIRAMISU) {
+            ActivityCompat.shouldShowRequestPermissionRationale(activity, Manifest.permission.POST_NOTIFICATIONS)
+        } else {
+            false
+        }
+    }
```
iujames, line 95: probably ok, but wondering if there is a way we could avoid having to check this on every single decorate call? I guess that gets complicated in some cases to know when to refresh it. Maybe this is low overhead and no big deal.

iujames (approved):

appcues/src/main/java/com/appcues/analytics/AutoPropertyDecorator.kt
andretortolano: open to ideas, I cant think of a way to keep this "in cache" in a reliable way

andretortolano: open to ideas, I cant think of a way to keep this "in cache" in a reliable way

PR #573, opened 2024-03-07
✨ Adding option for packageNames in AppcuesConfig

andretortolano:
Adding the property packageNames that allows customers to set different paths when we look for fonts in res/fonts

* Adding to the debugger scan was easy and straight forward
* Adding to the composition (Text and Syles) on the other hand required a lot of change 
    -> Created `LocalPackageNames` 

Improving logging when font coming from builder is not found:
image

also Adding informative block about missing fonts
image

### DIFF
--- a/appcues/src/main/java/com/appcues/AppcuesConfig.kt
+++ b/appcues/src/main/java/com/appcues/AppcuesConfig.kt
@@ -90,6 +90,12 @@ public data class AppcuesConfig internal constructor(
      */
     var additionalAutoProperties: Map<String, Any> = emptyMap()
 
+    /**
+     * Optional property that allows for setting custom package list for fonts, mostly used when your fonts are stored in
+     * a different package than your main app (multi-module apps)
+     */
+    var packageNames: List<String> = arrayListOf()
+
     // internally used for ui test on debug variant
     internal var imageLoader: ImageLoader? = null
     internal var configApiBasePath: String? = null
--- a/appcues/src/main/java/com/appcues/ui/extensions/StyleComponentExt.kt
+++ b/appcues/src/main/java/com/appcues/ui/extensions/StyleComponentExt.kt
@@ -7,10 +7,14 @@ import android.graphics.Typeface
 import android.os.Build.VERSION
 import android.os.Build.VERSION_CODES
 import androidx.annotation.RequiresApi
+import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.BiasAlignment
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.Font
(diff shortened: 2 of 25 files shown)


appcues/src/main/java/com/appcues/ui/extensions/StyleComponentExt.kt
```diff
@@ -56,11 +64,27 @@ internal fun ComponentStyle.getTextAlignment(): TextAlign? {
...
+    val packageNames = LocalPackageNames.current
+    val fontFamily = FontFamily.getSystemFontFamily(fontName)
+        ?: FontFamily.getFontResource(context, packageNames, fontName)
         ?: FontFamily.getFontAsset(context, fontName)
         ?: FontFamily.getSystemFont(fontName)
+
+    val logcues = LocalLogcues.current
+    LaunchedEffect(Unit) {
+        if (fontName != null && fontFamily == null) {
+            logcues.error(
```
iujames, line 79: I guess we don't have warnings, but this feels more like a warning, since things can still run with fallback fonts right?

appcues/src/main/java/com/appcues/ui/extensions/StyleComponentExt.kt
```diff
@@ -76,32 +100,50 @@ private fun FontFamily.Companion.getSystemFontFamily(fontName: String?): FontFam
...
+        }
+    }
+
+    // font was not found in resources
+    return null
+}
+
+// try to load a custom font from a resource in the host application
+@RequiresApi(VERSION_CODES.O)
+@SuppressLint("DiscouragedApi")
```
iujames, line 124: curious what uses DiscouragedApi? was that there before?

appcues/src/main/java/com/appcues/AppcuesConfig.kt
```diff
@@ -90,6 +90,12 @@ public data class AppcuesConfig internal constructor(
      */
     var additionalAutoProperties: Map<String, Any> = emptyMap()
 
+    /**
+     * Optional property that allows for setting custom package list for fonts, mostly used when your fonts are stored in
+     * a different package than your main app (multi-module apps)
+     */
+    var packageNames: List<String> = arrayListOf()
```
iujames, line 97: should it be something more specific like `fontPackageNames` ?

iujames (approved):
no objection with approach - couple minor comments

I added a commit with some suggested revisions to the doc just to try to make it as clear as possible. It was simpler than trying to comment here on the doc. Feel free to edit or revert if you don't think those are valid.

appcues/src/main/java/com/appcues/ui/extensions/StyleComponentExt.kt
andretortolano: getIdentifier since it can lead to errors

appcues/src/main/java/com/appcues/ui/extensions/StyleComponentExt.kt
andretortolano: makes sense

appcues/src/main/java/com/appcues/AppcuesConfig.kt
andretortolano: so I was considering leaving this as packageNames for in case in the future we need this for something else besides fonts.

PR #577, opened 2024-03-15
✨ Adding AppcuesFirebaseMessagingService

andretortolano:
* Added AppcuesFirebaseMessagingService - This allows customers to plugin their firebase "app" into their app, work as an abstraction for how messaging works when sent from Appcues only.
* Added Static method calls to AppcuesFirebaseMessagingService - Allows customers to have multiple messaging services (calling static methods that will return true if its a handled message or false when its ignored)
* Added Initial Debugger implementation for Push messages - Pending to add more error messages, but the initial structure checks if push message is sent
* Added XML config options - Initial options that customers can override with their own settings. 
* Modified sample app to implement AppcuesFirebaseMessagingService.


appcues/src/main/java/com/appcues/AppcuesFirebaseMessagingService.kt
```diff
@@ -0,0 +1,100 @@
...
+import java.net.URL
+
+public class AppcuesFirebaseMessagingService : FirebaseMessagingService() {
+
+    public companion object {
+
+        // Suppressing MissingPermission for POST_NOTIFICATION since this should be done on customer's end.
+        @SuppressLint("MissingPermission")
+        @JvmStatic
+        public fun handleMessage(context: Context, message: RemoteMessage): Boolean {
```
andretortolano, line 27: static method to allow for custom service to call similar and check if we can handle that message.

```kotlin
public void onMessageReceived(RemoteMessage remoteMessage) {
    super.onMessageReceived(remoteMessage);
    if (AppcuesFirebaseMessagingService.handleMessage(this, remoteMessage)) {
      // appcues handled message
    } else {
      // message from other messaging provider
    }
  }
```

appcues/src/main/java/com/appcues/AppcuesFirebaseMessagingService.kt
```diff
@@ -0,0 +1,100 @@
...
+                builder.setContentIntent(pendingIntent)
+            }
+
+            NotificationManagerCompat.from(context).notify(1, builder.build())
+
+            return true
+        }
+
+        @JvmStatic
+        public fun setToken(token: String) {
```
andretortolano, line 75: similar to handleMessage, if they have a custom service they can also setup the token using this

```kotlin
   override fun onNewToken(token: String) {
        super.onNewToken(token)

        AppcuesFirebaseMessagingService.setToken(token)
    }
```

appcues/src/main/java/com/appcues/AppcuesFirebaseMessagingService.kt
```diff
@@ -0,0 +1,100 @@
...
+            Appcues.pushToken = token
+        }
+    }
+
+    internal class AppcuesMessagingData(message: RemoteMessage) {
```
andretortolano, line 80: decided to organize our Firebase message data into a class, so we can control whats required and whats optional information we need to post the notification

samples/kotlin-android-app/src/main/AndroidManifest.xml
```diff
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
 
-    <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
```
andretortolano, line 5: file got some changes from re-formatting it.

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -131,6 +133,10 @@ public class Appcues internal constructor(internal val scope: AppcuesScope) {
      * @param properties Optional properties that provide additional context about the user.
      */
     public fun identify(userId: String, properties: Map<String, Any>? = null) {
+        if (pushToken != null && pushToken != storage.pushToken) {
+            storage.pushToken = pushToken
+        }
```
iujames, line 138: should it reset the `pushToken` value in companion to nil after this check?

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -67,6 +67,8 @@ public class Appcues internal constructor(internal val scope: AppcuesScope) {
          * provided by the SDK is based on Android View layout information.
          */
         public var elementTargeting: ElementTargetingStrategy = AndroidTargetingStrategy()
+
+        internal var pushToken: String? = null
```
iujames, line 71: what happens if we already have a session and the value is set? It seems we need a way to immediately notify any Appcues instance and let it set in storage? when this is called from the service. Otherwise it would only update if/when another identify happens on L137 correct?

appcues/src/main/java/com/appcues/Appcues.kt
iujames: checking understanding, we don't call `setPushToken(...)` directly since the session_started event that is coming here will handle the device update, and the separate device_updated event is not needed in this case.

appcues/src/main/java/com/appcues/AppcuesFirebaseMessagingService.kt
```diff
@@ -0,0 +1,88 @@
...
+
+        // required
+        val title: String = message.data["title"] ?: throw IllegalStateException("title not found")
+        val body: String = message.data["body"] ?: throw IllegalStateException("body not found")
+        val notificationId: String =
+            message.data["appcues_attachment_url"] ?: throw IllegalStateException("appcues_attachment_url not found")
```
iujames, line 73: I believe this should be looking for `appcues_notification_id` rather than attachment

appcues/src/main/java/com/appcues/AppcuesFirebaseMessagingService.kt
iujames: nice split out of that logic 👍

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -131,6 +133,10 @@ public class Appcues internal constructor(internal val scope: AppcuesScope) {
      * @param properties Optional properties that provide additional context about the user.
      */
     public fun identify(userId: String, properties: Map<String, Any>? = null) {
+        if (pushToken != null && pushToken != storage.pushToken) {
```
andretortolano, line 136: The idea here is that whenever we identify we first check if the value of pushToken (static) is different than null and also different than the previous storage value, if so we update it before the actual identify call (this will make so session_started will push the new value)

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -271,9 +277,11 @@ public class Appcues internal constructor(internal val scope: AppcuesScope) {
      *  @param token A globally unique token that identifies this device to FCM.
      */
     public fun setPushToken(token: String?) {
-        storage.pushToken = token
+        if (token != storage.pushToken) {
```
andretortolano, line 280: adding a safe-guard check, if token is the same as the existing we dont track DeviceUpdated

samples/kotlin-android-app/google-services.json
iujames, line 1: should it be injected in CI?

appcues/src/main/java/com/appcues/Appcues.kt
andretortolano: this value is not possible to set as its internal, the only way this is set is when we receive a response from the FirebaseMessagingService which should happen before the application onCreate since its inside of a service.

if user decides to refresh token and use the listener (like you did on the POC) then they can call appcues.setPushToken(token), which then will trigger device_updated

samples/kotlin-android-app/google-services.json
andretortolano: since our project is open source that would make sense to do that, I'm not sure how to properly set this up

appcues/src/main/java/com/appcues/Appcues.kt
iujames: I'm thinking about the fact that not all app creates would call identify, is that an issue? sometimes the app just launches and uses the user stored in cache already.

andretortolano: we could maybe move this to the AnalyticsTracker and make sure we do this before the actual session start. its a good callout

iujames: nice, I like that idea - always pick up any token that the service might have set on session start

samples/kotlin-android-app/google-services.json
iujames: I'm thinking we put it in our normal s3 storage and pull it in during the build, [just like we do the keystore](https://github.com/appcues/appcues-android-sdk/blob/main/.circleci/config.yml#L63-L65)

iujames: this file is now available to add in the circle config like that, from 
```
      - aws-s3/copy:
          from: 's3://appcues-mobile-certs/appcues-kotlin-example-google-services.json'
          to: './samples/kotlin-android-app/google-services.json'
```
and then I suppose add to the `.gitignore` to keep it locally

appcues/src/main/java/com/appcues/debugger/DebuggerStatusManager.kt
```diff
@@ -328,4 +360,27 @@ internal class DebuggerStatusManager(
             }
         }
     }
+
+    suspend fun checkPushValidation() {
+        if (pushConfigured == true) return
+
+        pushErrorText = null
+        pushValidationToken = "push-token"
```
iujames, line 368: is this or the null value in L376 or 381 needed? It seems like it is really validated when the receive happens on L54 AppcuesFirebaseMessenger? I'm sort of confused by the validation flow here.

appcues/src/main/java/com/appcues/AppcuesFirebaseMessagingService.kt
```diff
@@ -0,0 +1,125 @@
...
+                return false
+            }
+
+            if (data.test) {
+                // during testing we just want to validate that push message came through
+                context.startActivity(
+                    Intent(Intent.ACTION_VIEW, Uri.parse("appcues-${data.appId}://sdk/debugger/push-token")).apply {
+                        flags = Intent.FLAG_ACTIVITY_NEW_TASK
+                    }
+                )
```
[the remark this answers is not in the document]
iujames: Also, I think we want this to be the content intent on the tap of the push? not just triggered immediately on receive. As the test push says to tap on it to verify installation - which would then link into the main app handler.

appcues/src/main/java/com/appcues/util/NotificationBuilderExt.kt
```diff
@@ -0,0 +1,99 @@
...
+        )
+    } ?: run {
+        setStyle(
+            NotificationCompat.BigTextStyle()
+                .bigText(data.body)
+        )
+    }
+
+    setSmallIcon(R.drawable.appcues_notification_small_icon)
+    setColor(ContextCompat.getColor(context, R.color.appcues_notification_color))
```
iujames, line 69: should these be here? or should they use the host application defaults?

appcues/src/main/res/values/appcues.xml
iujames, line 1: I see the host application overrides for these - do we need to provide these at all inside the sdk itself? Maybe that is a requirement for default values.

samples/kotlin-android-app/src/main/res/values/appcues.xml
iujames, line 1: TODO: will need to update our main installation docs to reflect adding this file and providing info on the supported values.

iujames (commented):
Looks like everything progressing along nicely - I think there are really two areas overlapping in this PR that make it a little larger and harder to process: (1) the firebase service handler and related duties and (2) the debugger validation tool. It might be simpler to understand if the debugger validation pieces were split out into their own PR, but I know that might also be a fair bit of extra work at this point.

appcues/src/main/java/com/appcues/util/NotificationBuilderExt.kt
andretortolano: so this is the "default" value  that they can override in their XML, but I think this is a nice setup, if they dont do anything it will just work with our default values (appcues_color set to 5C5CFF) and the small notification icon, both of which they can replace with their own.

I made it this way because there is no good way for me to get a valid icon, or even the color to set automatically as default

samples/kotlin-android-app/src/main/res/values/appcues.xml
andretortolano: yes makes sense, I plan to add this somewhere in the docs, and then we should have all possible options they can customize in a single file like this.

appcues/src/main/java/com/appcues/debugger/DebuggerStatusManager.kt
```diff
@@ -288,44 +315,65 @@ internal class DebuggerStatusManager(
...
-                    deepLinkValidationToken = null
-                    deepLinkConfigured = false
-                    deepLinkErrorText = contextWrapper.getString(R.string.appcues_debugger_status_check_deep_link_error_handler)
-                    updateData()
-                }
-            } else {
-                deepLinkConfigured = false
-                deepLinkErrorText = contextWrapper.getString(R.string.appcues_debugger_status_check_deep_link_error_manifest)
+        // create new random id and send it to the server
+        val token = "push-${UUID.randomUUID()}"
```
iujames, line 365: It worked ok to have a non-uuid string here with a `push-` prefix?

appcues/src/main/java/com/appcues/util/NotificationBuilderExt.kt
```diff
@@ -0,0 +1,105 @@
...
+    val intent = if (data.test) {
+        // during testing we just want to validate that push message came through
+        Intent(Intent.ACTION_VIEW, Uri.parse("appcues-${data.appId}://sdk/debugger/${data.id}"))
+    } else {
+        Intent(Intent.ACTION_VIEW, Uri.parse("appcues-${data.appId}://sdk/notification")).apply {
+            putExtra("notification_id", data.id)
+            putExtra("forward_deeplink", data.deeplink)
+            putExtra("show_content", data.experienceId)
+        }
+    }
```
iujames, line 82: sorry but I think we may still have some details to work out here since the `test` value will be true on both the test push AND any non-published content like builder preview pushes.

I think we need to handle these scenarios
1. the notification ID matches what we had sent for a test - this is the case we mark the test successful in the debugger
2. the notification ID has `test: true` - this is some other preview type of push, it should still execute as normal with deep links, flow links, etc - for preview - but should NOT track analytics
3. all others - normal case - should execute any linked content AND track analytics

andretortolano: its done now see if that makes sense

appcues/src/main/java/com/appcues/debugger/DebuggerStatusManager.kt
```diff
@@ -353,7 +353,7 @@ internal class DebuggerStatusManager(
     }
 
     suspend fun checkPushValidation() {
-        if (pushStatus == SUCCESS) return
+        if (pushStatus == SUCCESS || pushStatus == LOADING) return
```
iujames, line 356: prob ok to allow another test on SUCCESS case? if they wanted to retry and check anything about their app UI. Not a big deal but any reason to prevent that?

appcues/src/main/res/values/strings.xml
```diff
@@ -31,6 +31,10 @@
     <string name="appcues_debugger_status_check_deep_link_instruction">Tap to check configuration</string>
     <string name="appcues_debugger_status_check_deep_link_error_manifest">Error 1: AndroidManifest intent-filter scheme missing</string>
     <string name="appcues_debugger_status_check_deep_link_error_handler">Error 2: Appcues SDK not receiving links</string>
+    <string name="appcues_debugger_status_check_push_title">Appcues Push</string>
+    <string name="appcues_debugger_status_check_push_instruction">Tap to check configuration</string>
+    <string name="appcues_debugger_status_check_push_cancel">Tap to clear notifications</string>
+    <string name="appcues_debugger_status_check_push_error_1">Error 1: x</string>
```
andretortolano, line 37: will be worked on later (errors)

appcues/src/main/java/com/appcues/debugger/DebuggerStatusManager.kt
andretortolano: yes, also decided to prefix with test-push and test-deeplink for readability.

appcues/src/main/java/com/appcues/DeepLinkHandler.kt
```diff
@@ -86,6 +96,26 @@ internal class DeepLinkHandler(
         }
     }
 
+    private fun processNotification(extras: Bundle) {
+        analyticsTracker.track(
+            PushOpened.eventName,
+            properties = mapOf("notification_id" to extras.getString("notification_id", null)),
```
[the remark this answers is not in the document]
andretortolano: the only property that needs to be decorated after (Autoprop decorator) is device_id.

appcues/src/main/java/com/appcues/analytics/AutoPropertyDecorator.kt
```diff
@@ -132,6 +133,9 @@ internal class AutoPropertyDecorator(
             AnalyticsEvent.DeviceUpdated.eventName -> {
                 attributes[DEVICE_PROPERTY] = deviceProperties
             }
+            AnalyticsEvent.PushOpened.eventName -> {
+                attributes[DEVICE_ID] = storage.deviceId
```
andretortolano, line 137: when decorating PushOpened we include the DEVICE_ID (cannot be accessed when we receive the message or when we handle the deeplink)

appcues/src/main/java/com/appcues/analytics/AutoPropertyDecorator.kt
```diff
@@ -28,6 +28,7 @@ internal class AutoPropertyDecorator(
         const val IDENTITY_PROPERTY = "_identity"
         const val DEVICE_PROPERTY = "_device"
         const val LAST_SEEN_AT = "_lastSeenAt"
+        const val DEVICE_ID = "device_id"
```
iujames, line 31: I think unused now

appcues/src/main/res/values/strings.xml
```diff
@@ -31,6 +31,10 @@
     <string name="appcues_debugger_status_check_deep_link_instruction">Tap to check configuration</string>
     <string name="appcues_debugger_status_check_deep_link_error_manifest">Error 1: AndroidManifest intent-filter scheme missing</string>
     <string name="appcues_debugger_status_check_deep_link_error_handler">Error 2: Appcues SDK not receiving links</string>
+    <string name="appcues_debugger_status_check_push_title">Appcues Push</string>
+    <string name="appcues_debugger_status_check_push_instruction">Tap to check configuration</string>
+    <string name="appcues_debugger_status_check_push_loading_instruction">Tap on the test notification in system tray</string>
+    <string name="appcues_debugger_status_check_push_error_1">Error 1: x</string>
```
iujames, line 37: is this one used?

iujames (approved):
LGTM - lets move forward and iterate on it as needed! Great start to test out push

Note to selves: the `sdk4` branch doesn't have the codecov checks so we'll need to cycle back and look at that and consider how to keep our test coverage numbers up - in future follow ups

appcues/src/main/res/values/strings.xml
iujames: oh nm, i see other resolved comment now.

PR #586, opened 2024-04-01
✨ Adding request-push action

andretortolano:
Adding appcues/request-push action 

Creates a new transparent activity just so we can call permission request and listen to the return of the REQUEST_CODE, which is internally implemented by this new API `ActivityResultLauncher `

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/RequestPermissionActivity.kt
+++ b/appcues/src/main/java/com/appcues/ui/RequestPermissionActivity.kt
@@ -1,0 +1,96 @@
+package com.appcues.ui
+
+import android.Manifest
+import android.content.Context
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.net.Uri
+import android.os.Build.VERSION
+import android.os.Build.VERSION_CODES
+import android.os.Bundle
+import android.provider.Settings
+import androidx.activity.result.contract.ActivityResultContracts.RequestPermission
+import androidx.activity.result.contract.ActivityResultContracts.StartActivityForResult
+import androidx.appcompat.app.AppCompatActivity
+import androidx.core.app.NotificationManagerCompat
+import androidx.core.content.ContextCompat
+import kotlinx.coroutines.CompletableDeferred
+
+internal class RequestPermissionActivity : AppCompatActivity() {
+
+    companion object {
+
+        private const val EXTRA_PERMISSION_KEY = "EXTRA_PERMISSION"
+
+        // since there can only be one review activity running at a time, this companion level
+        // deferred can be used to await completion
+        var completion: CompletableDeferred<Boolean>? = null
+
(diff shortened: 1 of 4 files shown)


appcues/src/main/java/com/appcues/ui/RequestPermissionActivity.kt
```diff
@@ -0,0 +1,96 @@
...
+            }
+            // POST_NOTIFICATION did not exist before API 33, so instead we direct user to open the settings page
+            VERSION.SDK_INT < VERSION_CODES.TIRAMISU &&
+                permissionKey == Manifest.permission.POST_NOTIFICATIONS &&
+                !NotificationManagerCompat.from(this).areNotificationsEnabled() -> {
+                val intent = Intent().apply {
+                    action = Settings.ACTION_APPLICATION_DETAILS_SETTINGS
+                    data = Uri.fromParts("package", packageName, null)
+                }
+                settingsLauncher.launch(intent)
```
iujames, line 76: ok i see that the `settingsLauncher` and `permissionLauncher` internally call the completion and finish so these cases should be covered, in terms of ensuring the completion handler is invoked

appcues/src/main/java/com/appcues/ui/RequestPermissionActivity.kt
```diff
@@ -0,0 +1,96 @@
...
+    override fun finish() {
+        super.finish()
+        // remove exit animation from this activity
+        if (VERSION.SDK_INT >= VERSION_CODES.UPSIDE_DOWN_CAKE) {
+            overrideActivityTransition(OVERRIDE_TRANSITION_CLOSE, 0, 0)
+        } else {
+            @Suppress("DEPRECATION")
+            overridePendingTransition(0, 0)
+        }
+    }
```
iujames, line 95: is there any way that the activity could get into here outside of the expected path? Wondering if there is a way the completion callback would not be invoked.

iujames (approved):

appcues/src/main/java/com/appcues/ui/RequestPermissionActivity.kt
andretortolano: so I tested different apis right.

before 33 you dont actually have the permission, thats why we need to check the API level and instead of showing the permission dialog we send them to the settings page.

both scenarios (requestin permission and settings page) when coming back here will resolve and finish() + complete the task,  the other possible scenario is when they already have permission, or in case permission returns false right away (also handled by the launcher)

I could not think of any other scenarios surrounding this for now

iujames: what about a deep link coming in and redirecting the app to another activity or something like that? could it leave an action in a hung state?

andretortolano: not able to reproduce any issue of this sort but as a safety measure for now i will add a delayedJob that cancels this activity if it stays in resumed state for long (should never happen)

PR #588, opened 2024-04-04
♻️  Updating sample app to support show experience

andretortolano:
simple addition of query param on sample app deeplink to support showing content after moving to a particular screen. like so:

```kotlin
deeplink-android "appcues-example://profile?experience=dc3c3e23-cc34-4720-94b8-6a1db104dc7d"
```


iujames (approved):
no issue if this went into `main` really, but also probably no pressing reason so keeping it on `sdk4` ok with me too.

PR #590, opened 2024-04-05
♻️  Adding properties to push data & improve naming

andretortolano:
including properties: `notification_version` and `workflow_version` as part of notification data model.

also whenever they are present we want to include both in the push_opened track.

+ Renaming `id` to `notificationId` and  old `notificationId` to `androidNotificationId` for clarification

### DIFF
--- a/appcues/src/main/java/com/appcues/AppcuesFirebaseMessagingService.kt
+++ b/appcues/src/main/java/com/appcues/AppcuesFirebaseMessagingService.kt
@@ -53,7 +53,7 @@ public class AppcuesFirebaseMessagingService : FirebaseMessagingService() {
             if (ActivityCompat.checkSelfPermission(context, permission.POST_NOTIFICATIONS) != PackageManager.PERMISSION_GRANTED)
                 return false
 
-            val isCheckPush = data.test && data.id.startsWith("test-push")
+            val isCheckPush = data.test && data.notificationId.startsWith("test-push")
 
             context.getNotificationBuilder()
                 .setupNotification(message)
@@ -78,10 +78,12 @@ public class AppcuesFirebaseMessagingService : FirebaseMessagingService() {
 
     @JsonClass(generateAdapter = true)
     internal data class AppcuesMessagingData(
-        @Json(name = "notification_id")
-        val id: String,
         val title: String,
         val body: String,
+        @Json(name = "notification_id")
+        val notificationId: String,
+        @Json(name = "notification_version")
+        val notificationVersion: Long? = null,
         @Json(name = "account_id")
         val accountId: String,
         @Json(name = "app_id")
@@ -92,6 +94,8 @@ public class AppcuesFirebaseMessagingService : FirebaseMessagingService() {
         val workflowId: String?,
         @Json(name = "workflow_task_id")
         val workflowTaskId: String?,
+        @Json(name = "workflow_version")
+        val workflowVersion: Long? = null,
         @Json(name = "deep_link_url")
         val deeplink: String?,
         @Json(name = "attachment_url")
@@ -101,9 +105,9 @@ public class AppcuesFirebaseMessagingService : FirebaseMessagingService() {
         @Json(name = "category")
         val category: String?,
         @Json(name = "android_notification_id")
-        val notificationId: String?,
+        val androidNotificationId: String?,
         @Json(name = "test")
-        val test: Boolean = false,
+        val test: Boolean = false
     )
 
     override fun onMessageReceived(remoteMessage: RemoteMessage) {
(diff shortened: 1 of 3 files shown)


appcues/src/main/java/com/appcues/DeepLinkHandler.kt
```diff
@@ -27,9 +27,11 @@ internal class DeepLinkHandler(scope: AppcuesScope) {
...
     companion object {
 
-        private const val NOTIFICATION_ID_EXTRA = "ID"
-        private const val NOTIFICATION_WORKFLOW_ID = "WORKFLOW_ID"
-        private const val NOTIFICATION_WORKFLOW_TASK_ID = "WORKFLOW_TASK_ID"
+        private const val NOTIFICATION_ID_EXTRA = "NOTIFICATION_ID"
+        private const val NOTIFICATION_VERSION_EXTRA = "NOTIFICATION_VERSION"
+        private const val NOTIFICATION_WORKFLOW_ID_EXTRA = "WORKFLOW_ID"
+        private const val NOTIFICATION_WORKFLOW_TASK_ID_EXTRA = "WORKFLOW_TASK_ID"
+        private const val NOTIFICATION_WORKFLOW_VERSION_EXTRA = "WORKFLOW_TASK_ID"
```
iujames, line 34: copy/paste bug here on the string value

appcues/src/main/java/com/appcues/DeepLinkHandler.kt
```diff
@@ -142,12 +146,14 @@ internal class DeepLinkHandler(scope: AppcuesScope) {
         if (!extras.getBoolean(NOTIFICATION_TEST_EXTRA)) {
             analyticsTracker.track(
                 PushOpened.eventName,
-                properties = mapOf(
+                properties = mapOf<String, Any?>(
                     "notification_id" to extras.getString(NOTIFICATION_ID_EXTRA, null),
-                    "workflow_id" to extras.getString(NOTIFICATION_WORKFLOW_ID, null),
-                    "workflow_task_id" to extras.getString(NOTIFICATION_WORKFLOW_TASK_ID, null),
+                    "push_notification_version" to extras.getInt(NOTIFICATION_VERSION_EXTRA).let { if (it == 0) null else it },
```
iujames, line 151: I think `0` could be a valid version in some case? I'm not sure, but wondering if there is a reason we should filter that out here.

iujames: it looks like an unpublished draft has version 0. Can we use -1 and filter that out? -1 should never be an actual value.

appcues/src/main/java/com/appcues/AppcuesFirebaseMessagingService.kt
```diff
@@ -78,10 +78,12 @@ public class AppcuesFirebaseMessagingService : FirebaseMessagingService() {
...
     @JsonClass(generateAdapter = true)
     internal data class AppcuesMessagingData(
-        @Json(name = "notification_id")
-        val id: String,
         val title: String,
         val body: String,
+        @Json(name = "notification_id")
+        val notificationId: String,
+        @Json(name = "notification_version")
+        val notificationVersion: Int? = null,
```
iujames, line 86: was testing some things and discovered this, and L98 below need to be `Long?` actually - they are timestamp values

appcues/src/main/java/com/appcues/DeepLinkHandler.kt
```diff
@@ -142,12 +151,14 @@ internal class DeepLinkHandler(scope: AppcuesScope) {
         if (!extras.getBoolean(NOTIFICATION_TEST_EXTRA)) {
             analyticsTracker.track(
                 PushOpened.eventName,
-                properties = mapOf(
+                properties = mapOf<String, Any?>(
                     "notification_id" to extras.getString(NOTIFICATION_ID_EXTRA, null),
-                    "workflow_id" to extras.getString(NOTIFICATION_WORKFLOW_ID, null),
-                    "workflow_task_id" to extras.getString(NOTIFICATION_WORKFLOW_TASK_ID, null),
+                    "push_notification_version" to extras.getInt(NOTIFICATION_VERSION_EXTRA, -1).let { if (it == -1) null else it },
```
iujames, line 156: and then this needs to change to `getLong` and compare to `-1L` (and L159 below)

appcues/src/main/java/com/appcues/AppcuesFirebaseMessagingService.kt
iujames: yes, confirmed

PR #595, opened 2024-04-11
♻️  Storing deferred actions when processing notification deeplink in no-session run

andretortolano:
First pass of this task. here we are storing in Room table the deferred actions (currently Track, Deeplink and ShowContent) when we see that sessionMonitor has no active session, else we run those normally.

+ Separated some piece of DeeplinkHandler for clarity and to solve some detekt warnings (Complex methods)

This is the big chunk, want to make sure this makes sense before moving forward, the idea here is that we will be storing deferred actions based on appId and userId (to account for different sessions and multi-instances of appcues),

and then later we are going to implement the other piece which after the identify we will see if there is any elements in that list and run those, also clear those elements from the list afterwards.

### DIFF
--- a/appcues/src/main/java/com/appcues/push/PushDeeplinkHandler.kt
+++ b/appcues/src/main/java/com/appcues/push/PushDeeplinkHandler.kt
@@ -1,0 +1,102 @@
+package com.appcues.push
+
+import android.content.Intent
+import android.net.Uri
+import android.os.Bundle
+import com.appcues.AppcuesCoroutineScope
+import com.appcues.AppcuesFirebaseMessagingService.AppcuesMessagingData
+import com.appcues.SessionMonitor
+import com.appcues.Storage
+import com.appcues.data.PushRepository
+import com.appcues.di.component.AppcuesComponent
+import com.appcues.di.component.inject
+import com.appcues.di.scope.AppcuesScope
+import com.appcues.di.scope.inject
+import kotlinx.coroutines.launch
+
+internal class PushDeeplinkHandler(
+    override val scope: AppcuesScope,
+) : AppcuesComponent {
+
+    companion object {
+
+        private const val NOTIFICATION_ID_EXTRA = "ID"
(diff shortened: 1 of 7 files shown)


appcues/src/main/java/com/appcues/push/PushDeeplinkHandler.kt
```diff
@@ -0,0 +1,102 @@
...
+            "workflow_version" to extras.getLong(NOTIFICATION_WORKFLOW_VERSION_EXTRA, -1L).let { if (it == -1L) null else it },
+            "device_id" to storage.deviceId
+        ).filterValues { it != null }.mapValues { it.value as Any }
+
+        val deeplink = extras.getString(NOTIFICATION_FORWARD_DEEPLINK_EXTRA, null)
+        val experienceId = extras.getString(NOTIFICATION_SHOW_CONTENT_EXTRA, null)
+
+        val action = PushOpenedAction(storage.userId, properties, deeplink, experienceId)
+
+        if (sessionMonitor.hasSession()) {
```
andretortolano, line 88: when hasSession we run process, else we defer til processDeferred is called (next up on identify)

appcues/src/main/java/com/appcues/push/PushOpenedProcessor.kt
```diff
@@ -0,0 +1,56 @@
...
+    private val appcues by scope.inject<Appcues>()
+    private val logcues by scope.inject<Logcues>()
+
+    private val deferredAction: MutableList<PushOpenedAction> = arrayListOf()
+
+    fun defer(action: PushOpenedAction) {
+        deferredAction.add(action)
+    }
+
+    suspend fun processDeferred(userId: String) {
```
andretortolano, line 32: processDeferred will run the first item found (single item contains all information needed (properties, deeplink and experienceId)

and will clear the entire list to ensure we are not stacking old actions from other users here.

iujames (approved):

PR #596, opened 2024-04-15
♻️  Processing deferred notification actions during identify call

andretortolano:
Last piece for this task, consuming deferred action for given userId during the identify call, similarly to whats done on iOS [here](https://github.com/appcues/appcues-ios-sdk/blob/sdk4/Sources/AppcuesKit/Appcues.swift#L404)

* Minor changes to adjust unit tests.

### DIFF
--- a/appcues/src/main/java/com/appcues/SessionMonitor.kt
+++ b/appcues/src/main/java/com/appcues/SessionMonitor.kt
@@ -27,7 +27,7 @@ internal class SessionMonitor(
     val sessionId: UUID?
         get() = _sessionId
 
-    val isExpired: Boolean
+    private val isExpired: Boolean
         get() = lastActivityAt?.let {
             TimeUnit.MILLISECONDS.toSeconds(Date().time - it.time) >= sessionTimeout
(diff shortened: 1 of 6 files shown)


appcues/src/main/java/com/appcues/SessionMonitor.kt
```diff
@@ -42,7 +42,7 @@ internal class SessionMonitor(
     }
 
     fun hasSession(): Boolean {
-        val isValid = sessionId != null || isExpired.not()
+        val isValid = sessionId != null && isExpired.not()
```
iujames, line 45: oh woops, was this from the last PR? just double checking we dont have any bug here in `main`

iujames (approved):

appcues/src/main/java/com/appcues/SessionMonitor.kt
andretortolano: that was fom the last PR (sorry about that), I found it when I was running unit tests. which i will be improving later too.

PR #597, opened 2024-04-16
♻️  Improving debugger push check

andretortolano:
Another pass on debugger check for push configuration.

Improving how errors are displayed, doing some pre-checks before sending the request.


appcues/src/main/java/com/appcues/data/remote/NetworkRequest.kt
```diff
@@ -24,4 +25,22 @@ internal object NetworkRequest {
         } catch (exception: JsonDataException) {
             null
         }
+
+    internal suspend fun <Success> executeV2(apiCall: suspend () -> Success): ResultOf<Success, RemoteError> =
```
iujames, line 29: 😬 this is unfortunate to have to introduce v2 variants and have callsites know which one to use.

Wondering if we can have a better abstraction layer here - i.e. he lower level networking code can handle the different response bodies as needed, but just pass back a mapped error model with status code and message to the caller? Is there a way to keep the callsites uniform but handle different error structures in the networking layer?

andretortolano: its possible as we can possibly do the same thing we did with the Experience response, have a parent class but that would actually mean a bigger refactor on the current usage.

personally i think we need to intentionally know the possible return for a particular endpoint, maybe the naming is not helping here. but we dont need to have a geneic mapping for errors

iujames: ok, yeah i can see how that would be a large refactor. More specific naming for now would be nice, maybe `executePushRequest` and `PushErrorResponse` ?

PR #605, opened 2024-04-29
♻️  Auto collecting push token whenever appcues is instanciated

andretortolano:
small improvement I would like to try, so on new installations of firebase it seems like the "onNewToken" is called, but then after that its never called again (unless the token changes), so this means customers would need to always get the current token and set in order to ensure we have that information set in the SDK, or else they could update Appcues on an already existing FirebaseMessagingService installation and never get setPushToken called.

### DIFF
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -24,6 +24,7 @@ import com.appcues.trait.ExperienceTrait
 import com.appcues.trait.ExperienceTraitLevel
 import com.appcues.trait.TraitRegistry
 import com.appcues.ui.ExperienceRenderer
+import com.google.firebase.messaging.FirebaseMessaging
 import kotlinx.coroutines.launch
 import kotlin.DeprecationLevel.ERROR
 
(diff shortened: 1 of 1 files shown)


appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -46,7 +47,21 @@ public fun Appcues(
...
         config = AppcuesConfig(accountId, applicationId).apply { config?.invoke(this) }
-    ).get()
+    ).get<Appcues>().also { appcues ->
+        try {
+            // this is necessary to ensure that not only when we get a new token but whenever we initialize the application appcues should know
+            // whats the latest available token
+            FirebaseMessaging.getInstance().token.addOnCompleteListener {
+                // ensures we have token set
+                if (it.isSuccessful) {
+                    appcues.setPushToken(it.result)
```
iujames, line 57: wondering if there might be a better place to stick this around when a new session is starting (SessionMonitor) -- this is very early here, and it would result in an attempt to start a session to track analytics for the device update. It is likely the user is not known here, or is the user from the previous session that may need to log in again? This would create a session and send the analytics before the app had a chance to reset if they had a session timeout?

Just want to think through the analytics timing and see if there is a good spot during the normal session startup that would fit.

andretortolano: it makes sense, what if we change to this instead:

```kotlin
FirebaseMessaging.getInstance().token.addOnCompleteListener {
    // ensures we have token set
    if (it.isSuccessful) {
        // store token on static pushToken that will be used on the next session start
        Appcues.pushToken = it.result
    }
}
```

iujames: nice, yeah I think that would work well here

PR #610, opened 2024-05-02
Revise automatic push token collection

iujames:
Posting my idea here to slightly revise the auto token collection from #605 . I believe the benefits may be
1. The debugger will see the token (from storage) even if a user has not been identified yet, to avoid any confusion like we had today on whether the token was actually being captured for push
2. Avoid any potential race condition where a session start would occur before the Firebase callback that sets the global token variable (temp storage)

### DIFF
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -41,27 +41,40 @@ public fun Appcues(
     accountId: String,
     applicationId: String,
     config: (AppcuesConfig.() -> Unit)? = null,
-): Appcues = Bootstrap
+): Appcues {
     // This creates the Scope and initializes the Appcues instance within, then returns the Appcues instance
     // ready to go with the necessary dependency configuration in its scope.
-    .createScope(
-        context = context,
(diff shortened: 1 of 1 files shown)


appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -41,28 +41,41 @@ public fun Appcues(
...
+        // whats the latest available token
+        FirebaseMessaging.getInstance().token.addOnCompleteListener {
+            // ensures we have token set
+            if (it.isSuccessful) {
+                if (sessionMonitor.hasSession()) {
+                    // can call setPushToken directly if we are already in a session, so the device_updated event is tracked
+                    appcues.setPushToken(it.result)
+                } else {
+                    // store token on static pushToken that will be used on the next session start to track device props
+                    storage.pushToken = it.result
```
iujames, line 68: This reference to storage causes [this test](https://github.com/appcues/appcues-android-sdk/blob/main/appcues/src/test/java/com/appcues/AppcuesTest.kt#L63) to fail, as it does not use the mocked storage when the new scope is created here. Need to think about this one. I don't think it is a _real_ issue, but breaks that test.

andretortolano (approved):
nice increment on the token auto collection

PR #611, opened 2024-05-02
♻️  Adding new push verification check for the manifest.xml

andretortolano:
Adding extra check that verifies that there is a Service filtered by the firebase messaging Action.

if there is another one not the default FirebaseMessagingService it means they are either using Appcues's or a custom one (which is what we want)

+ Re-ordered existing error messages to match the sequence they are being checked.

### DIFF
--- a/appcues/src/main/res/values/strings.xml
+++ b/appcues/src/main/res/values/strings.xml
@@ -34,10 +34,11 @@
     <string name="appcues_debugger_status_check_push_title">Appcues Push</string>
     <string name="appcues_debugger_status_check_push_instruction">Tap to check configuration</string>
     <string name="appcues_debugger_status_check_push_loading_instruction">Tap on the test notification in system tray</string>
-    <string name="appcues_debugger_status_check_push_error_no_token">Error 1: No Push token registered with Appcues</string>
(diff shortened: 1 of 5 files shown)


appcues/src/main/res/values/strings.xml
```diff
@@ -38,6 +38,7 @@
     <string name="appcues_debugger_status_check_push_error_no_permission">Error 2: Notification permission required. Tap to enable through settings</string>
```
iujames, line 38: side note, i noticed Error 2 here happens before Error 1 in the check sequence - should we have our errors ordered to match the order of the checks?

iujames (approved):
helpful addition!

appcues/src/main/res/values/strings.xml
andretortolano: they were added in the order that i thought of them basically, but if you agree with the order they are being checked  i can rename the strings to match that

iujames: yeah i was thinking order them to match the checks - since it might be confusing to get Error 2, fix it, then get Error 1 - feel like you are going backwards. minor thing

PR #612, opened 2024-05-06
♻️  Enhancing preview feedback on backend error

andretortolano:


appcues/src/main/java/com/appcues/DeepLinkHandler.kt
```diff
@@ -105,7 +105,7 @@ internal class DeepLinkHandler(scope: AppcuesScope) {
                 }
             }
 
-            pushDeeplinkHandler.processLink(segments, extras, query) -> true
+            pushDeeplinkHandler.processLink(activity, segments, extras, query) -> true
```
andretortolano, line 108: adding context so we can access strings and make the toast

iujames (approved):
looks good - heads up that my error handling changes in pushserv and api are hopefully going on today to ensure the client is getting back responses as expected in all cases - was missing the "app not configured for push" item, previously.

Would like to test this out after those updates to in just to check (same on iOS side I'm working on).

Also - does the full error response body show in the debugger? Would like to make sure that is there as well, so they can see the actual error - i.e. device not push enabled, or app not configured, etc.

andretortolano:
image
this is how its showing in detailed log.

PR #613, opened 2024-05-13
✨ Adding google services key (protected key)

andretortolano:
syncing up the approach used for the RN project, we include the google-services.json while hiding the key as GOOGLE_SERVICES_KEY placeholder

* Found an issue with the latest android image, changing from default to a static reference for now until we take a proper look at whats happening there.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,8 +26,8 @@ orbs:
 # -------------------------
 executors:
   android_compatible:
-    machine: 
-      image: android:default
+    machine:
+      image: android:2024.01.1
     resource_class: large
     working_directory: ~/code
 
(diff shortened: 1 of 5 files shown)


andretortolano:
(tests are passing locally)

.circleci/config.yml
```diff
@@ -63,20 +63,17 @@ commands:
...
-      - aws-s3/copy:
-          from: 's3://appcues-mobile-certs/appcues-kotlin-example-google-services.json'
-          to: './samples/kotlin-android-app/google-services.json'
       - run:
           name: Generate Gradle cache key
           command: ./checksum.sh /tmp/checksum.txt
       - restore_cache:
           name: Restore Gradle cache
-          key: gradle-{{ checksum "/tmp/checksum.txt" }}
+          key: 1-gradle-{{ checksum "/tmp/checksum.txt" }}
```
iujames, line 71: the prefix here ended up probably not needed, but also harmless - no issue if kept in I think.

iujames (approved):

iujames:
> (tests are passing locally)

looks like its that old flakey analytics test 😔 fine to ignore, but wish we could sort that one out.

.circleci/config.yml
andretortolano: i was considering undoing it, but since we already burned the 1-* i think we need to keep now so in the future if we think its cache issue we go for 2-*

Issue #616, opened 2024-05-22
ModalStateMachineOwner.onUiIdle java.lang.OutOfMemoryError com.appcues.ui ModalStateMachineOwner.kt:116


mahmed1987:
We get this crash in our production app for a few devices. 

I have printed the stacktrace below

<img width="850" alt="image" src="https://github.com/appcues/appcues-android-sdk/assets/26518062/a70ceeee-4786-4c9a-b18b-05e3f8765adc">

This is the OS breakdown of the crash
<img width="546" alt="image" src="https://github.com/appcues/appcues-android-sdk/assets/26518062/e29380a5-8615-4ca1-b5f7-9b424d516086">

This is the device breakdown
<img width="550" alt="image" src="https://github.com/appcues/appcues-android-sdk/assets/26518062/5aa7cc8f-a0b7-4903-ab12-d1b6928714a1">

iujames:
Hi @mahmed1987 , thank you for letting us know about this issue. We believe it is possible that the use of a Timer could be leading to memory issues in some cases. One idea we have is to change this to use a coroutine and a delay - @andretortolano proposed this change in #617 if you have any feedback to share on that approach. Thanks!

mahmed1987:
@iujames  thanks for quick resolution of the issue . Any idea when this fix would be included in your release cycle ?

andretortolano:
hey @mahmed1987 we plan on doing a small patch that include the fix this afternoon

iujames:
@mahmed1987 confirming that @andretortolano did release the 3.1.14 version yesterday, with this update, and you should see that available now in Maven Central. Thank you @andretortolano !

mahmed1987:
Thanks a bunch good folks. I hope this would resolve the situation. Will update our app with the new dependency and wait for the results

mahmed1987:
@iujames  I forgot to mention that i am using the the appcues dependency through segment 

I am using this to include the appcues sdk in the code 

com.appcues:segment-appcues:3.1.14.

I guess 3.1.14 must be available for com.appcues:appcues but for the above dependency with segment I get the error 

"Caused by: org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find com.appcues:segment-appcues:3.1.14.
" during dependency resolution .

iujames:
@mahmed1987 got it, thank you for this additional info. We can update our Segment plugin to match, today.

iujames:
[...]

PR #617, opened 2024-05-22
🐛 Fix Timer memory leak on anchored tooltip retry

andretortolano:
Reported in issue #616 

Replacing timer with debounce logic using coroutines since Timer appears to cause memory leak. [REF1](https://androidtrainningcenter.blogspot.com/2013/12/handler-vs-timer-fixed-period-execution.html)

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/ModalStateMachineOwner.kt
+++ b/appcues/src/main/java/com/appcues/ui/ModalStateMachineOwner.kt
@@ -18,10 +18,9 @@ import com.appcues.statemachine.states.IdlingState
 import com.appcues.statemachine.states.RenderingStepState
 import com.appcues.ui.utils.getParentView
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
-import java.util.Timer
-import java.util.TimerTask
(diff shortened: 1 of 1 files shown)


appcues/src/main/java/com/appcues/ui/ModalStateMachineOwner.kt
```diff
@@ -113,11 +112,10 @@ internal class ModalStateMachineOwner(
         // cancel any previous
         uiIdleDebounceTimer?.cancel()
 
-        uiIdleDebounceTimer = Timer().schedule(delay = SCROLL_DEBOUNCE_MILLISECONDS) {
-            // set to know to allow new Timer
-            uiIdleDebounceTimer = null
```
iujames, line 118: note: we are losing the setting of this value back to null here, but looking at the code, it seems like we don't actually need that, correct? The prior comment just made it sound like it was needed, but perhaps that was an old comment.

iujames (approved):
thanks! hopefully this will improve any potential memory issues here. Good find!

appcues/src/main/java/com/appcues/ui/ModalStateMachineOwner.kt
andretortolano: it was a typo on the old comment, but its not required

PR #622, opened 2024-06-05
✨ Adding delay action

andretortolano:
Adding support for the new @appcues/delay action:

```json
{
  "on": "navigate",
  "type": "@appcues/delay",
  "config": {
    "duration": 3000
  }
}
```

### DIFF
--- a/appcues/src/main/java/com/appcues/action/appcues/DelayAction.kt
+++ b/appcues/src/main/java/com/appcues/action/appcues/DelayAction.kt
@@ -1,0 +1,22 @@
+package com.appcues.action.appcues
+
+import com.appcues.action.ExperienceAction
+import com.appcues.data.model.AppcuesConfigMap
+import com.appcues.data.model.getConfigOrDefault
+import kotlinx.coroutines.delay
+
+internal class DelayAction(
+    override val config: AppcuesConfigMap,
+) : ExperienceAction {
+
+    companion object {
+
+        const val TYPE = "@appcues/delay"
+    }
+
+    private val duration = config.getConfigOrDefault("duration", 0)
+
+    override suspend fun execute() {
+        delay(duration.toLong())
+    }
+}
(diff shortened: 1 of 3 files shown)


appcues/src/main/java/com/appcues/action/appcues/DelayAction.kt
```diff
@@ -0,0 +1,22 @@
...
+internal class DelayAction(
+    override val config: AppcuesConfigMap,
+) : ExperienceAction {
+
+    companion object {
+
+        const val TYPE = "@appcues/delay"
+    }
+
+    private val duration = config.getConfigOrDefault("duration", 0)
```
mmaatttt, line 17: fwiw on iOS if there's no `duration` value, I'm failing the init of the delay action so that it gets skipped. Different than defaulting to 0 here, but the end result will be the same.

mmaatttt (approved):

appcues/src/main/java/com/appcues/action/appcues/DelayAction.kt
andretortolano: yeah. thinking here there was not harm in just handling with default 0

appcues/src/test/java/com/appcues/action/appcues/DelayActionTest.kt
```diff
@@ -0,0 +1,52 @@
...
+        assertThat(DelayAction.TYPE).isEqualTo("@appcues/delay")
+    }
+
+    @Test
+    fun `delay SHOULD call delay with given value of 300ms `() = runTest {
+        // GIVEN
+        val action = DelayAction(mapOf("duration" to 300))
+
+        launch {
+            val workDuration = testScheduler.timeSource.measureTime {
```
iujames, line 31: TIL `measureTime` cool - and it is always the exact ms of the delay? thats sort of amazing but useful here

iujames (approved):

appcues/src/test/java/com/appcues/action/appcues/DelayActionTest.kt
andretortolano: yes. [reference](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/#:~:text=To%20test%20regular%20suspend%20functions,of%20delaying%20%2F%2F%20...%20%7D)

PR #624, opened 2024-06-20
:arrow_up: Update compose to use bom, side effects include updating k…

Shakezulla57:
Update Compose to 1.6.8 and use bill of materials to handle versioning. In order to do this, Kotlin also need to be updated to 1.9.24. In order to update Kotlin, ksp needed to be used for Room instead of kapt 😄. I also updated Moshi to use ksp since kapt is deprecated.

### DIFF
--- a/appcues/build.gradle
+++ b/appcues/build.gradle
@@ -3,6 +3,7 @@ plugins {
     id 'kotlin-android'
     id 'org.jetbrains.dokka'
     id 'kotlin-kapt'
+    id 'com.google.devtools.ksp'
 }
 
 apply from: 'artifacts-android.gradle'
@@ -11,10 +12,13 @@ apply from: 'copy-dependencies.gradle'
 apply from: 'dokka.gradle'
 
 ext.room_version = '2.5.2'
-ext.compose_version = '1.6.2'
-ext.compose_compiler_version = '1.4.8'
+ext.compose_compiler_version = '1.5.14'
 ext.coil_version = '2.4.0'
 ext.moshi_version = '1.15.0'
+
+ksp {
+    arg('room.schemaLocation', "$projectDir/schemas")
+}
 
 android {
     compileSdk 34
@@ -29,12 +33,6 @@ android {
         testApplicationId = "com.appcues.test"
 
         consumerProguardFiles "consumer-rules.pro"
-
-        javaCompileOptions {
-            annotationProcessorOptions {
-                arguments += ["room.schemaLocation": "$projectDir/schemas".toString()]
-            }
-        }
     }
 
     resourcePrefix 'appcues_'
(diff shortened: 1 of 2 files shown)


andretortolano:
Thanks @Shakezulla57 for this pull request, we really appreciate your contribution to our sdk, right now we are limited to do some of those updates because of our need to keep it compatible with some versions of cross-platforms frameworks like ReactNative.

I'm curious to know the reason behind this pull request, i imagine you are trying to update Compose version on your own app and we are hitting some compatibility issue?

Shakezulla57:
> I'm curious to know the reason behind this pull request, i imagine you are trying to update Compose version on your own app and we are hitting some compatibility issue?

That's exactly right, we're on the latest version of Compose and are looking to use your sdk. Below is the exception I'm getting. My assumption is a Compose mismatch.

``
java.lang.NoSuchMethodError: No static method HorizontalPager-xYaah8o(Landroidx/compose/foundation/pager/PagerState;Landroidx/compose/ui/Modifier;Landroidx/compose/foundation/layout/PaddingValues;Landroidx/compose/foundation/pager/PageSize;IFLandroidx/compose/ui/Alignment$Vertical;Landroidx/compose/foundation/gestures/snapping/SnapFlingBehavior;ZZLkotlin/jvm/functions/Function1;Landroidx/compose/ui/input/nestedscroll/NestedScrollConnection;Lkotlin/jvm/functions/Function4;Landroidx/compose/runtime/Composer;III)V in class Landroidx/compose/foundation/pager/PagerKt; or its super classes (declaration of 'androidx.compose.foundation.pager.PagerKt' appears in /data/data/com.***.**/code_cache/.overlay/base.apk/classes.dex)
``

It may be possible to update the Compose libraries without updating the compiler, which may avoid the Kotlin update.

andretortolano:
I did some local testing around this error and interestingly enough the version 1.6.8 appears to be working ok (no breaking chances in the compose api). I also ended up testing version 1.7.0-beta03 which in fact has some breaking changes we will need to sort out when time is right.

could you confirm for me what are your dependencies around compose and material so i can give it another shot trying to replicate this?

also not sure if there is more to this error log but if you could see if there is a line pointing to a specific code line that would be helpful

Shakezulla57:
@andretortolano I just retested with 1.6.8 and everything is working fine. I must've missed a dependency or a gradle sync when I was going back and forth in my troubleshooting. The issue I'm seeing is with the material3 1.3.0-beta02 artifact. Unfortunately we're using a few things from the beta so I can't downgrade.

You probably saw this error with your testing, but here it is.
```
at com.appcues.ui.modal.FullScreenModalKt$FullScreenModal$1$1$1$2.invoke(FullScreenModal.kt:78)
at com.appcues.ui.modal.FullScreenModalKt$FullScreenModal$1$1$1$2.invoke(FullScreenModal.kt:77)
```

Is it worth me updating this PR to use 1.7.0 and fixing what needs to be fixed, or do you have a timeline for when you might do an update?

andretortolano:
@Shakezulla57 unfortunately we cant yet support compose 1.7.0 as it is in beta, and probably after its official version is released we are going to need some extra time to evaluate migrating as many customers uses compose right now and that would probably break on their end as well.

I think i tracked down and it comes down to the deprecation of "rememberRipple", we should now use "ripple" for better performance it seems. here in [this branch](https://github.com/appcues/appcues-android-sdk/compare/main...support/material-1.7.0) is a simple draft of how this migration would roll.

at this time there is one possible workaround I imagine you can try.

1. Apply this changes to your forked Appcues

2. remove the signing lines from 106-109 from publish-maven.gradle
```
//    signing {
//        useInMemoryPgpKeys(signingKey, signingKeyPwd)
//        sign publishing.publications.release
//    }
```

3. update appcues.properties to add a custom classifier of your choice 
```
VERSION_MAJOR = 3
VERSION_MINOR = 2
VERSION_PATCH = 0
VERSION_CLASSIFIER = material17
```

4. sync project and run task :appcues:publishToMavenLocal
image

5. in your project add as mavenLocal() as your last dependency resolution repository

6. add your custom appcues dependency (in this case here 3.2.0-material17)

Hopefully that would get you unblocked to continue validating appcues even though you are on the latest version of compose.

Shakezulla57:
Thanks @andretortolano. I've got it working locally.

Shakezulla57:
@andretortolano Have you considered providing a beta release that depends on beta versions of compose? I can imagine since compose and specifically material 3 are still young, appcues consumers may want to utilize new components and performance enhancements as soon as possible. Thanks for your help btw.

andretortolano:
Hi @Shakezulla57, we talked this over with the team here, and at this time we are not able to support prerelease versions of the Appcues SDK dedicated to using prerelease versions of dependency libraries. We understand the need here and your use case, and hope that having our SDK as open source here fully supports any level of testing and modification you would like to do. However, since our SDK uses prerelease versions already to preview our own functionality - such as our 4.0 alpha/beta versions in progress now for push notifications - we are unable to also support the other cases around external dependency changes, directly. We also release our SDK products in several other products like Segment plugins and cross platform plugins, thus we are unable to support multiple dependency variations in prerelease across all of these products effectively.

We will certainly continue to monitor our dependency versions and pick up these updates when they are released and we can safely support them for our customer base across all of our SDK products. Thank you for using Appcues and for your careful examination and input for us to consider with our Android SDK here.

Shakezulla57:
@iujames Thanks for the follow up.

PR #625, opened 2024-06-24
Merging main (3.2.0) into sdk4

andretortolano:


andretortolano:
even though I merged its giving conflicts (i think this is because we did rebase before) so i changed some settings for the branch protection to allow me and james to force push (required after rebase) when needed.

will also keep commits in chronological order instead of the merge.

PR #629, opened 2024-07-24
💥 Removing deprecated Anonymous call

andretortolano:
Removing deprecated anonymous call as part of sdk4 breaking changes


iujames (approved):
nice clean up!

PR #632, opened 2024-09-03
🐛 fix state machine when another activity is started on top

andretortolano:
Reported by a customer in [#mobile-support ](https://appcues.slack.com/archives/C03JYPH1NJZ/p1724931068889339) it looks like an oversight on our part for some of the possible outcomes from creating an overlay activity on top of an existing activity that was presenting an experience.

changes: 

* When transitioning from [RenderingStep to EndingExperience](https://github.com/appcues/appcues-android-sdk/compare/fix/overlay_view_experience?expand=1#diff-9a2951280d6c94900b1fb1c73e24089e62d7f593ef819d695183f858be5a74cdR92) we will always pass in the AwaitDismissEffect to ensure that it will always fire up a new ui state in AppcuesViewModel.

*  Since the reported issue was this error message `java.lang.IllegalStateException: Method removeObserver must be called on the main thread` when [removing](https://github.com/appcues/appcues-android-sdk/compare/fix/overlay_view_experience?expand=1#diff-eed3d6c94ac3b8e438725716643f71dd0a8e01f23818b047709860a6be149b15R131) the view we will ensure to do all the processing from the main thread

* In some scenarios when there is another activity on top the view was not properly getting removed, this is because when we call "getParentView" we are looking for the topMost decorView available, when multiple activities are in display we can get a different decorView from the first one we got when its time to do cleanup, to avoid this we are storing the view that was used to show the experience.

### DIFF
--- a/appcues/src/main/java/com/appcues/statemachine/states/RenderingStepState.kt
+++ b/appcues/src/main/java/com/appcues/statemachine/states/RenderingStepState.kt
@@ -90,18 +90,20 @@ internal data class RenderingStepState(
     }
 
     private fun toEndingExperience(action: EndExperience): Transition {
+        // passing in the effect on EndingStepState to ensure the UI will dismiss
+        // even though we might not be waiting for the completion before moving to
+        // the next step.
+        val awaitDismissEffect = AwaitDismissEffect(action)
         return if (action.destroyed) {
             // this means the activity hosting the experience was destroyed externally (i.e. deep link) and we should
             // immediately transition to EndingExperience - not rely on the UI to do it for us (it's gone)
-            next(EndingStepState(experience, flatStepIndex, action.markComplete, null), ContinuationEffect(action))
+            next(EndingStepState(experience, flatStepIndex, action.markComplete, awaitDismissEffect), ContinuationEffect(action))
         } else {
             // otherwise, its a natural end of experience from an in-experience action / dismiss
             // and should be communicated to the UI layer to dismiss itself.
             //
             // instead of using sideEffect we pass EndExperience on EndingStep
             // then AppcuesViewModel will continue to EndExperience when appropriate
-            val awaitDismissEffect = AwaitDismissEffect(action)
-
             next(EndingStepState(experience, flatStepIndex, action.markComplete, awaitDismissEffect), awaitDismissEffect)
         }
     }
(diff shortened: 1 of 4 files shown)


appcues/src/main/java/com/appcues/ui/presentation/ViewPresenter.kt
```diff
@@ -40,6 +42,7 @@ internal abstract class ViewPresenter(
     private val appcuesViewTreeOwner: AppcuesViewTreeOwner by inject()
     private val appcuesConfig: AppcuesConfig by inject()
 
+    private var parentView: ViewGroup? = null
```
andretortolano, line 45: Looking at this again after mulling it over I think its best to change this to a weakReference instead (avoid any future issue since its a view)

iujames: yeah that sounds like a good safeguard

appcues/src/main/java/com/appcues/statemachine/states/RenderingStepState.kt
```diff
@@ -90,18 +90,20 @@ internal data class RenderingStepState(
...
     private fun toEndingExperience(action: EndExperience): Transition {
+        // passing in the effect on EndingStepState to ensure the UI will dismiss
+        // even though we might not be waiting for the completion before moving to
+        // the next step.
+        val awaitDismissEffect = AwaitDismissEffect(action)
         return if (action.destroyed) {
             // this means the activity hosting the experience was destroyed externally (i.e. deep link) and we should
             // immediately transition to EndingExperience - not rely on the UI to do it for us (it's gone)
-            next(EndingStepState(experience, flatStepIndex, action.markComplete, null), ContinuationEffect(action))
+            next(EndingStepState(experience, flatStepIndex, action.markComplete, awaitDismissEffect), ContinuationEffect(action))
```
iujames, line 100: checking understanding - the `EndingStepState.awaitDismissEffect` is still optional, and thus no changes in the `AppcuesViewModel` implementation where it does the null check. However, it should really never be null there, in practice.

The one spot it could still be null is when transitioning in `EndingStepState.toBeginningStepState` where it is checking if a new container will be presented or not.

andretortolano: Thats right, so one thing we missed when we last talked is that if we had the uiState dismiss to go through the container dismisses, this is bad for multiple steps (failed snapshot tests), one thing we could do to make it clearer is to create pass in a "isDestroyed" boolean on EndingStepState to do this. decided to keep the State as is but open to change.

iujames: ok sounds good, and good catch. seems ok to me to move this fix along to resolve the observed issue, thanks!

iujames (approved):

PR #633, opened 2024-09-09
🐛 Add required sanatize for all events + UUID sanatize

andretortolano:
from support [thread](https://appcues.slack.com/archives/C03JYPH1NJZ/p1725461076933259) we need to sanatize all events

+ sanatize UUID to string value


iujames (approved):
nice!

PR #635, opened 2024-09-25
Update Compose and some other dependencies

iujames:
These are the updates I think are needed for Compose 1.7+ to work, plus misc other version updates to bring up to latest. This appears to have no cascading effects on the Kotlin version (due to staying on same Compose Compiler version 1.4.8) or on anything like Android targetSdk - so, I think rather safe for rolling out across our platforms, but that is what I'll be testing next.

Without this update, any app that begins using Compose 1.7 will hit a crash at runtime with an error like:
```
java.lang.NoSuchMethodError: No static method HorizontalPager-xYaah8o(Landroidx/compose/foundation/pager/PagerState;Landroidx/compose/ui/Modifier;Landroidx/compose/foundation/layout/PaddingValues;Landroidx/compose/foundation/pager/PageSize;IFLandroidx/compose/ui/Alignment$Vertical;Landroidx/compose/foundation/gestures/snapping/SnapFlingBehavior;ZZLkotlin/jvm/functions/Function1;Landroidx/compose/ui/input/nestedscroll/NestedScrollConnection;Lkotlin/jvm/functions/Function4;Landroidx/compose/runtime/Composer;III)V in class Landroidx/compose/foundation/pager/PagerKt; or its super classes (declaration of 'androidx.compose.foundation.pager.PagerKt' appears in /data/app/~~klOqzmhdPE9hWF1ejvR47g==/com.appcues.samples.kotlin-N61GjtIaKqkxVkXcboelvw==/base.apk)
```

This appears to be due to the change in 1.7.0 noted [here](https://developer.android.com/jetpack/androidx/releases/compose-foundation#1.7.0-alpha02), where `HorizontalPager` was promoted to stable. Thus, something in how this loads at runtime was changed (I'm not sure exactly what) due to [these](https://android-review.googlesource.com/c/platform/frameworks/support/+/2888286/5/compose/foundation/foundation/src/commonMain/kotlin/androidx/compose/foundation/pager/Pager.kt) changes, even though the actual API itself appears unchanged. 

There are some other deprecations, some of which were errors that also needed code changes to make the Compose upgrade, and those are included here.

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/ui/CaptureConfirmDialog.kt
+++ b/appcues/src/main/java/com/appcues/debugger/ui/CaptureConfirmDialog.kt
@@ -19,14 +19,13 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.shape.RoundedCornerShape
-import androidx.compose.foundation.text.ClickableText
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.ContentAlpha
 import androidx.compose.material.OutlinedTextField
 import androidx.compose.material.Surface
 import androidx.compose.material.Text
 import androidx.compose.material.TextFieldDefaults
-import androidx.compose.material.ripple.rememberRipple
+import androidx.compose.material.ripple
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.derivedStateOf
@@ -45,14 +44,15 @@ import androidx.compose.ui.graphics.asImageBitmap
 import androidx.compose.ui.graphics.drawscope.DrawScope
 import androidx.compose.ui.graphics.drawscope.Stroke
 import androidx.compose.ui.layout.ContentScale
-import androidx.compose.ui.platform.LocalUriHandler
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.text.LinkAnnotation
 import androidx.compose.ui.text.SpanStyle
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.withLink
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import androidx.compose.ui.window.Dialog
@@ -104,7 +104,7 @@ private fun Header(debuggerViewModel: DebuggerViewModel) {
                     onClick = { debuggerViewModel.closeExpandedView() },
                     role = Role.Button,
                     interactionSource = remember { MutableInteractionSource() },
-                    indication = rememberRipple(bounded = false, 24.dp),
+                    indication = ripple(bounded = false, 24.dp),
                     onClickLabel = stringResource(id = string.appcues_screen_capture_dismiss)
                 )
                 .drawBehind {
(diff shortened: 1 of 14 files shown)


appcues/src/main/java/com/appcues/debugger/ui/CaptureConfirmDialog.kt
```diff
@@ -104,7 +104,7 @@ private fun Header(debuggerViewModel: DebuggerViewModel) {
                     onClick = { debuggerViewModel.closeExpandedView() },
                     role = Role.Button,
                     interactionSource = remember { MutableInteractionSource() },
-                    indication = rememberRipple(bounded = false, 24.dp),
+                    indication = ripple(bounded = false, 24.dp),
```
iujames, line 107: `rememberRipple` --> `ripple` was a required change due to an error level deprecation https://developer.android.com/develop/ui/compose/touch-input/user-interactions/migrate-indication-ripple#migrate-remember-ripple

appcues/src/main/java/com/appcues/debugger/ui/CaptureConfirmDialog.kt
```diff
@@ -225,22 +225,22 @@ private fun CaptureContents(debuggerViewModel: DebuggerViewModel, capture: Captu
 
 @Composable
 private fun TextWebLink(text: String, @Suppress("SameParameterValue") url: String) {
-    val uriHandler = LocalUriHandler.current
-    ClickableText(
```
iujames, line 229: migrating away from `ClickableText` to `Text` with a `LinkAnnotation` within was a deprecation warning https://joebirch.co/android/migrating-from-the-clickabletext-composable-to-linkannotation/

appcues/src/main/java/com/appcues/debugger/ui/DebuggerComposition.kt
```diff
@@ -15,12 +15,12 @@ import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.layout.onSizeChanged
 import androidx.compose.ui.platform.LocalDensity
-import androidx.compose.ui.platform.LocalLifecycleOwner
 import androidx.compose.ui.platform.testTag
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.Lifecycle.Event
 import androidx.lifecycle.Lifecycle.Event.ON_ANY
 import androidx.lifecycle.LifecycleEventObserver
+import androidx.lifecycle.compose.LocalLifecycleOwner
```
iujames, line 23: adding the new `androidx.lifecycle:lifecycle-runtime-compose` dependency and changing this import was a deprecation warning

appcues/src/main/java/com/appcues/debugger/ui/DebuggerFloatingActionEvents.kt
```diff
@@ -122,7 +122,7 @@ private fun LazyItemScope.Item(event: DebuggerEventItem, isAnchoredStart: Boolea
                 .clip(RoundedCornerShape(6.dp))
                 .background(theme.background)
                 .padding(2.dp)
-                .animateItemPlacement(),
+                .animateItem(),
```
iujames, line 125: fix a deprecation warning on this one

appcues/src/main/java/com/appcues/trait/appcues/CarouselTrait.kt
```diff
@@ -31,7 +30,6 @@ internal class CarouselTrait(
         const val TYPE = "@appcues/carousel"
     }
 
-    @OptIn(ExperimentalFoundationApi::class)
```
iujames, line 34: as noted in PR description - HorizontalPager stuff is no longer marked experimental, so this is removed

iujames:
I assume this is going to cause some snapshot test differences as well, will look into that.

mmaatttt (approved):

appcues/src/main/java/com/appcues/ui/utils/AppcuesMargins.kt
```diff
@@ -60,3 +63,53 @@ private class MarginValuesModifier(
...
+// we need our own.
+//
+// This implementation still implements the same PaddingValues interface, so there is minimal impact to
+// existing call sites.
+//
+// the change to check negative values was in this update:
+// https://android-review.googlesource.com/c/platform/frameworks/support/+/3021136/3/compose/foundation/foundation-layout/src/commonMain/kotlin/androidx/compose/foundation/layout/Padding.kt
+@Stable
+@Suppress("FunctionNaming") // matching Compose naming for PaddingValues here
+internal fun MarginValues(all: Dp): PaddingValues = MarginValues(all, all, all, all)
```
andretortolano, line 80: I believe this is the way to go here too, we had this done for the Modifier too, not sure why they felt the need to add this restriction to the PaddingValues implementation as well

PR #641, opened 2024-09-30
Revert dependency versions to avoid kotlin version requirement change

iujames:
After analyzing our dependency usage a bit more, I discovered that reverting coil (image loading), moshi (json) and navigation-compose (debugger nav) components back to previous versions before #635  allows us to update to Compose 1.7.2 without imposing a new requirement of Kotlin 1.9. I don't see any compelling reason to force this upgrade, so I propose we keep those as-is. This lessens the burden of integration on some cross platform frameworks as well. Our minimum Kotlin version is 1.8.22 [as determined by the Compose compiler](https://developer.android.com/jetpack/androidx/releases/compose-kotlin) currently.

There does not appear to be any way to avoid the AGP 8+ requirement at this point, as it seems Compose 1.7 requires that. Hopefully this is acceptable, as that version has been out for a while, but it may require some additional integration work for some apps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ The SDK is a Kotlin library for sending user properties and events to the Appcue
 
 ### Prerequisites
 
-Your application's `build.gradle` must have a `compileSdkVersion` of 34+ and `minSdkVersion` of 21+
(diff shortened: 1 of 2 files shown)


mmaatttt (approved):
I’m sure some customers will appreciate the minimal changes

PR #642, opened 2024-10-02
Add support for swipe to dismiss on slideouts from edge

iujames:
This adds swipe to dismiss behavior similar to what was done for iOS in https://github.com/appcues/appcues-ios-sdk/pull/564

The core functionality builds upon the `.anchoredDraggable` Modifier that was added in Compose a few versions back https://fvilarino.medium.com/exploring-jetpack-compose-anchored-draggable-modifier-5fdb21a0c64c

We conditionally add this modifier if (A) the transition on the modal trait is `slide` (a slideout) and (B) the transition is from an edge (not center). The modifier defines an initial anchor point where the content is presenting, and then a dismissal anchor that represents the content being fully off screen. In the case that the content is not dismissible (the `ignoreBackdropTap: true` case on the `@appcues/skippable` trait) - the dismissal anchor moves to just a small fraction from the initial state, to provide bounce/rejection feedback.

**Dismissible case**

https://github.com/user-attachments/assets/3ed52075-a9d1-4cf2-b7c0-0b74d65b4892   

**Non-dismissible case**

https://github.com/user-attachments/assets/87b7a9e3-02ee-426f-bebb-a834932a3b54

One complication was vertical scrolling, when trying to swipe up or down on a slideout from top or bottom, respectively. If the content has vertical scroll enabled, then the gesture is captured by the scroll handler, and will not initiate the drag. The workaround used here was to conditionally add scroll, only if the container size determined that it needed to take the full maxHeight available in the layout - an indicator that the content is either larger than the maxHeight or exactly the same. This allows the common case for slideouts to work fine - not maxHeight content.

The previous `Experience.isSkippable(stepIndex)` is renamed to `allowDismissal`, and moved to the `Step` where it is determined. This is just an attempt at code clarity as it really represents the concept of dismissing a step from a gesture, either the back button or this new swipe action, and not just the presence of the skippable trait.

### DIFF
--- a/appcues/src/main/java/com/appcues/ui/composables/StepComposition.kt
+++ b/appcues/src/main/java/com/appcues/ui/composables/StepComposition.kt
@@ -2,6 +2,7 @@ package com.appcues.ui.composables
 
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
+import androidx.compose.foundation.layout.BoxWithConstraints
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.padding
@@ -10,11 +11,14 @@ import androidx.compose.foundation.verticalScroll
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.key
+import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.composed
+import androidx.compose.ui.layout.onGloballyPositioned
 import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.unit.Dp
 import com.appcues.data.model.Step
 import com.appcues.trait.StepDecoratingTrait.StepDecoratingType
 import com.appcues.trait.StickyContentPadding
@@ -32,7 +36,7 @@ internal fun Step.ComposeStep(
     key(id) {
         CompositionLocalProvider(
             LocalAppcuesActions provides actions,
-            LocalExperienceStepFormStateDelegate provides formState
+            LocalExperienceStepFormStateDelegate provides formState,
         ) {
             // used to get the padding values from step decorating trait and apply to the Column
             val density = LocalDensity.current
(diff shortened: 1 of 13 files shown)


appcues/src/main/java/com/appcues/ui/composables/StepComposition.kt
```diff
@@ -119,11 +125,31 @@ private fun Step.ComposeStickyContent(
...
+            .onGloballyPositioned { coordinates ->
+                contentHeight.intValue = coordinates.size.height
+            }
+            .let {
+                if (contentHeight.intValue < maxHeight) {
+                    it
+                } else {
+                    it.verticalScroll(rememberScrollState())
+                }
+            }
```
iujames, line 152: This is the change for the vertical scroll handling - conditionally adding scroll only if the container height is not less than max available height. It is more complex to measure the actual content within, but this container measurement should provide effectively the same result.

appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
iujames, line 1: all of the key changes for the AnchoredDraggable usage are in this file, the modal container for slideouts

appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
```diff
@@ -69,6 +95,8 @@ internal fun DialogModal(
 ) {
     val isDark = isSystemInDarkTheme()
     val density = LocalDensity.current
+    val viewModel = LocalViewModel.current
```
iujames, line 98: it appears that due to our snapshot testing set up, I cannot currently access the composition local for `LocalViewModel` here 😬 need to look into options

appcues/src/main/java/com/appcues/ui/composables/StepComposition.kt
iujames: snapshot tests are showing there is something not quite right here as well with content height sizing, looking into.

appcues/src/main/java/com/appcues/ui/modal/DialogModal.kt
iujames: reworked this to use an abstraction layer for testing in https://github.com/appcues/appcues-android-sdk/pull/642/commits/f482f85b2a1490e277bf0540267f2267d54be745

mmaatttt (approved):

PR #643, opened 2024-10-04
Allow interaction with underlying app when backdrop is disabled

iujames:
stacks on #642 

Similar change to iOS update in https://github.com/appcues/appcues-ios-sdk/pull/567 - if there is no `@appcues/backdrop` trait, then interactions with the app beneath our content are not blocked.

implications:
* Skippable trait will no longer cover the screen with a tap to dismiss handler, if no backdrop exists
* Target interaction will no longer capture input, if no backdrop exists
* Backdrop keyhole will not attempt to render, if no backdrop exists
* Effects trait will continue to operate as-is, so you could have confetti showing while also allowing interaction with the app behind that confetti

The above updates are accomplished by adding a new `isBlocking` property on any `BackdropDecoratingTrait`, to allow it to declare that it blocks input -- currently only true for `@appcues/backdrop`. When the step is rendering, we first check if any of the `BackdropDecoratingTraits` are blocking, and pass that information through to each trait being applied to the backdrop, so it can decide whether or not to enable certain functionality, as noted above.

### DIFF
--- a/appcues/src/main/java/com/appcues/data/mapper/step/StepMapper.kt
+++ b/appcues/src/main/java/com/appcues/data/mapper/step/StepMapper.kt
@@ -44,7 +44,7 @@ internal class StepMapper(
             id = from.id,
             content = responseContent.mapPrimitive(),
             presentingTrait = presentingTrait,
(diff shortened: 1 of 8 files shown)


mmaatttt (approved):
Quite elegant!

PR #648, opened 2024-10-17
Support improved testability with injectable CoroutineScope

iujames:
These updates came out of some investigation into how to make our automated tests more stable. The key change is allowing the tests to inject a `CoroutineScope` (i.e. `UnconfinedTestDispatcher`) instead of using the `AppcuesCoroutineScope`, to make the execution of our internal coroutines for analytics reporting more reliable during test runs. ([ref](https://developer.android.com/kotlin/coroutines/test?authuser=1#creating-your-own-testscope))

There were some other details found along the way that also tripped up tests, and I'll note those inline.

There is a corresponding change in the spec repo to use this new feature in the tests.

Posting this for visibility and to test in CI, but we can also wait for Andre on this one I think. Would be good to get his opinions on changes here.

### DIFF
--- a/appcues/src/debug/java/com/appcues/ExperienceTestHelper.kt
+++ b/appcues/src/debug/java/com/appcues/ExperienceTestHelper.kt
@@ -3,7 +3,9 @@ package com.appcues
 import android.content.Context
 import coil.ImageLoader
 import com.appcues.di.Bootstrap
+import com.appcues.di.definition.ScopedDefinition
 import com.appcues.di.scope.get
+import kotlinx.coroutines.CoroutineScope
 
 /**
  * custom instance call that allows for custom imageLoader (used by coil)
@@ -15,11 +17,13 @@ public fun Appcues(
     accountId: String,
     applicationId: String,
     imageLoader: ImageLoader?,
+    coroutineScope: CoroutineScope?,
     config: (AppcuesConfig.() -> Unit)? = null,
 ): Appcues = Bootstrap.createScope(
     context = context,
-    config = AppcuesConfig(accountId, applicationId).apply {
-        config?.invoke(this)
-        this.imageLoader = imageLoader
-    }
-).get()
+    config = AppcuesConfig(accountId, applicationId).apply { config?.invoke(this) }
+).also { scope ->
+    // optional test overrides
+    imageLoader?.let { scope.define(ImageLoader::class, ScopedDefinition { imageLoader }, true) }
+    coroutineScope?.let { scope.define(CoroutineScope::class, ScopedDefinition { coroutineScope}, true) }
+}.get()
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -25,6 +25,7 @@ import com.appcues.trait.ExperienceTraitLevel
 import com.appcues.trait.TraitRegistry
 import com.appcues.ui.ExperienceRenderer
 import com.google.firebase.messaging.FirebaseMessaging
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 
 /**
@@ -110,7 +111,7 @@ public class Appcues internal constructor(internal val scope: AppcuesScope) {
     private val activityScreenTracking by scope.inject<ActivityScreenTracking>()
     private val deepLinkHandler by scope.inject<DeepLinkHandler>()
     private val debuggerManager by scope.inject<AppcuesDebuggerManager>()
-    private val appcuesCoroutineScope by scope.inject<AppcuesCoroutineScope>()
+    private val appcuesCoroutineScope by scope.inject<CoroutineScope>()
     private val analyticsPublisher by scope.inject<AnalyticsPublisher>()
     private val pushOpenedProcessor by scope.inject<PushOpenedProcessor>()
 
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -140,6 +140,7 @@ public class Appcues internal constructor(internal val scope: AppcuesScope) {
 
         appcuesCoroutineScope.launch {
             analyticsTracker.analyticsFlow.collect {
+                @Suppress("TooGenericExceptionCaught")
                 try {
                     analyticsPublisher.publish(analyticsListener, it)
                 } catch (ex: Exception) {
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -140,7 +140,12 @@ public class Appcues internal constructor(internal val scope: AppcuesScope) {
 
         appcuesCoroutineScope.launch {
             analyticsTracker.analyticsFlow.collect {
-                analyticsPublisher.publish(analyticsListener, it)
+                try {
+                    analyticsPublisher.publish(analyticsListener, it)
+                } catch (_: Exception) {
+                    // ignore any exception that occurs in client app code that is observing analytics,
+                    // so we always continue collecting on the analyticsFlow
+                }
             }
         }
 
--- a/appcues/src/main/java/com/appcues/Appcues.kt
+++ b/appcues/src/main/java/com/appcues/Appcues.kt
@@ -142,9 +142,10 @@ public class Appcues internal constructor(internal val scope: AppcuesScope) {
             analyticsTracker.analyticsFlow.collect {
                 try {
                     analyticsPublisher.publish(analyticsListener, it)
-                } catch (_: Exception) {
+                } catch (ex: Exception) {
                     // ignore any exception that occurs in client app code that is observing analytics,
                     // so we always continue collecting on the analyticsFlow
+                    logcues.error(ex)
                 }
             }
         }
--- a/appcues/src/main/java/com/appcues/AppcuesConfig.kt
+++ b/appcues/src/main/java/com/appcues/AppcuesConfig.kt
@@ -1,6 +1,5 @@
 package com.appcues
 
-import coil.ImageLoader
 import com.appcues.LoggingLevel.NONE
 
 /**
@@ -101,7 +100,4 @@ public data class AppcuesConfig internal constructor(
      * a different package than your main app (multi-module apps)
      */
     var packageNames: List<String> = arrayListOf()
-
-    // internally used for ui test on debug variant
-    internal var imageLoader: ImageLoader? = null
 }
--- a/appcues/src/main/java/com/appcues/MainModule.kt
+++ b/appcues/src/main/java/com/appcues/MainModule.kt
@@ -1,5 +1,6 @@
 package com.appcues
 
+import coil.ImageLoader
 import com.appcues.action.ActionProcessor
 import com.appcues.action.ActionRegistry
 import com.appcues.data.AppcuesRepository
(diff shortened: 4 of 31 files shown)


appcues/src/debug/java/com/appcues/ExperienceTestHelper.kt
```diff
@@ -15,11 +17,13 @@ public fun Appcues(
...
-    config = AppcuesConfig(accountId, applicationId).apply {
-        config?.invoke(this)
-        this.imageLoader = imageLoader
-    }
-).get()
+    config = AppcuesConfig(accountId, applicationId).apply { config?.invoke(this) }
+).also { scope ->
+    // optional test overrides
+    imageLoader?.let { scope.define(ImageLoader::class, ScopedDefinition { imageLoader }, true) }
+    coroutineScope?.let { scope.define(CoroutineScope::class, ScopedDefinition { coroutineScope}, true) }
```
iujames, line 28: this is test only code (in `src/debug` folder) -- allows the tests to pass in a CoroutineScope, in addition to the ImageLoader option we had before. Now, it uses a new `override: true` option on the `scope.define` call that I added. This allows our normal code to be totally unaware of these test specifics, and just do the override right here.

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -110,7 +111,7 @@ public class Appcues internal constructor(internal val scope: AppcuesScope) {
     private val activityScreenTracking by scope.inject<ActivityScreenTracking>()
     private val deepLinkHandler by scope.inject<DeepLinkHandler>()
     private val debuggerManager by scope.inject<AppcuesDebuggerManager>()
-    private val appcuesCoroutineScope by scope.inject<AppcuesCoroutineScope>()
+    private val appcuesCoroutineScope by scope.inject<CoroutineScope>()
```
iujames, line 114: lots of changes will be like this - where instead of resolving `AppcuesCoroutineScope` from the DI container, it will be the higher level `CoroutineScope` interface.

appcues/src/main/java/com/appcues/AppcuesConfig.kt
```diff
@@ -101,7 +100,4 @@ public data class AppcuesConfig internal constructor(
      * a different package than your main app (multi-module apps)
      */
     var packageNames: List<String> = arrayListOf()
-
-    // internally used for ui test on debug variant
-    internal var imageLoader: ImageLoader? = null
```
iujames, line 106: no longer needed, due to the new test override approach

appcues/src/main/java/com/appcues/MainModule.kt
```diff
@@ -53,6 +55,6 @@ internal object MainModule : AppcuesModule {
 
         factory { StateMachine(actionProcessor = get(), lifecycleTracker = get(), onEndedExperience = it.next()) }
 
-        scoped { get<AppcuesConfig>().imageLoader ?: get<ImageLoaderWrapper>().build() }
+        scoped<ImageLoader> { get<ImageLoaderWrapper>().build() }
```
iujames, line 58: this file now just provides the defaults, which tests may override later.

appcues/src/main/java/com/appcues/analytics/AnalyticsTracker.kt
```diff
@@ -96,8 +96,6 @@ internal class AnalyticsTracker(
     }
 
     private fun updateAnalyticsFlow(type: AnalyticType, isInternal: Boolean, activity: ActivityRequest) {
-        appcuesCoroutineScope.launch {
-            _analyticsFlow.emit(TrackingData(type, isInternal, activity))
-        }
+        _analyticsFlow.tryEmit(TrackingData(type, isInternal, activity))
```
iujames, line 99: updated this flow to have a max size buffer (L22 above) and now call `tryEmit` here, which is non-suspending.

what I ran into, without this, is that analytics can get sent into the flow out of order, when using the [`UnconfinedTestDispatcher` ](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/-unconfined-test-dispatcher.html)

this is because we have multiple spots that can trigger calls into this class during flow execution - step interaction actions and experience lifecycle observer.

it _seems_ safe here to just use the non-suspending version with a large buffer, since we'll never get anywhere near that buffer size in practice, and the consumers of this flow are quite simple - our own debugger, and the analytics publisher that sends to any attached listeners.

appcues/src/main/java/com/appcues/analytics/ExperienceLifecycleTracker.kt
```diff
@@ -39,18 +37,18 @@ internal class ExperienceLifecycleTracker(
...
 
-    fun start(stateMachine: StateMachine, onEndedExperience: (Experience) -> Unit, dispatcher: CoroutineDispatcher = Dispatchers.IO) {
+    fun start(stateMachine: StateMachine, onEndedExperience: (Experience) -> Unit) {
         // ensure any existing observers are stopped before starting new ones
         stop()
 
-        stateJob = appcuesCoroutineScope.launch(dispatcher) { stateMachine.observeState(onEndedExperience) }
-        errorJob = appcuesCoroutineScope.launch(dispatcher) { stateMachine.observeErrors() }
+        stateJob = appcuesCoroutineScope.launch { stateMachine.observeState(onEndedExperience) }
+        errorJob = appcuesCoroutineScope.launch { stateMachine.observeErrors() }
```
iujames, line 51: removed the `launch(Dispatchers.IO)` here as it causes the tests to hang and not complete analytics handling as expected when running in Robolectric. It seems like it is not really needed? I'm not sure, but the actual network calls that happen later are already using `Dispatchers.IO`.

appcues/src/main/java/com/appcues/data/remote/NetworkRequest.kt
```diff
@@ -22,6 +23,8 @@ internal object NetworkRequest {
             exception.response()?.errorBody()?.source()?.let {
                 MoshiConfiguration.moshi.adapter(ErrorResponse::class.java).fromJson(it)
             }
+        } catch (exception: IOException) {
+            null
```
iujames, line 27: discovered in tests, which sometimes mock responses with no body, that an `IOException` is raised for unexpected end of file - so just covering that case here as well - would be expected to only really happen in test cases, but in any case, we don't want to allow an exception to bubble up while just trying to parse an error message.

appcues/src/main/java/com/appcues/di/scope/AppcuesScope.kt
```diff
@@ -12,8 +12,8 @@ internal data class AppcuesScope(
 
     private val definitions = hashMapOf<KClass<*>, Definition<*>>()
 
-    fun <T : Any> define(clazz: KClass<T>, definition: Definition<T>) {
-        if (definitions.contains(clazz)) throw DefinitionException("definition already registered for class $clazz")
+    fun <T : Any> define(clazz: KClass<T>, definition: Definition<T>, override: Boolean = false) {
+        if (!override && definitions.contains(clazz)) throw DefinitionException("definition already registered for class $clazz")
```
iujames, line 16: new option for `override` used by tests only

appcues/src/main/java/com/appcues/ui/ExperienceRenderer.kt
```diff
@@ -131,7 +133,7 @@ internal class ExperienceRenderer(override val scope: AppcuesScope) : AppcuesCom
         return newExperience.priority == NORMAL && state !is IdlingState
     }
 
-    suspend fun show(qualificationResult: QualificationResult) {
+    suspend fun show(qualificationResult: QualificationResult) = withContext(Dispatchers.Main) {
```
iujames, line 136: an important spot here where `show(..)` can be called during network response handling, and if we are using a test CoroutineScope, we still have to ensure that it gets back to the Main thread here for the UI work. This is not an issue in normal usage, as the AppcuesCoroutineScope is already using Dispatchers.Main

appcues/src/main/java/com/appcues/ui/composables/CompositionRemembers.kt
```diff
@@ -49,7 +50,9 @@ internal fun LaunchOnHideAnimationCompleted(block: () -> Unit) {
     with(remember { mutableStateOf(isContentVisible) }.value) {
         // if hide animation is completed
         if (isIdle && currentState.not()) {
-            block()
+            SideEffect {
+                block()
+            }
```
iujames, line 55: testing showed that this call to `dismiss()` that happens during hide can throw an exception, if the underlying Android View that is presenting our content (ViewPresenter) is removed while the content is composing (inside a Composable). I can no longer find the resource I had, but basically, if you are removing the ComposeView out from under a Composition, it needs to happen in a SideEffect so it runs after the Composition completes.

appcues/src/main/java/com/appcues/ui/presentation/ViewPresenter.kt
```diff
@@ -129,7 +129,7 @@ internal abstract class ViewPresenter(
 
     fun remove() {
         // ensure all this is running on the main thread
-        Handler(Looper.getMainLooper()).post {
+        postIfNeeded {
```
iujames, line 132: seems like we almost always do not need to post to the main thread here, and it would be complex to make this suspending, so I updated to this check for the main thread and only post if needed.  Tests are not stable without this, and they are already on the main thread when the dismissal occurs.

appcues/src/main/java/com/appcues/Appcues.kt
```diff
@@ -139,7 +140,12 @@ public class Appcues internal constructor(internal val scope: AppcuesScope) {
 
         appcuesCoroutineScope.launch {
             analyticsTracker.analyticsFlow.collect {
-                analyticsPublisher.publish(analyticsListener, it)
+                try {
+                    analyticsPublisher.publish(analyticsListener, it)
+                } catch (_: Exception) {
+                    // ignore any exception that occurs in client app code that is observing analytics,
+                    // so we always continue collecting on the analyticsFlow
+                }
```
iujames, line 148: this is a small enhancement we talked about a little while back - passing a call over to the customer listener is going into unknown code, so we should just safeguard for any exception and carry on, so we don't kill this coroutine and stop the analytics flow.

andretortolano: any idea how we could track if exceptions are getting thrown here? would be nice if at least the debugger was notified that. something happened here so if they see it while using with the debugger open they can report it and we can patch it up? (for later PR)

appcues/src/main/java/com/appcues/di/scope/AppcuesScope.kt
andretortolano: I like the approach, give the extra flexibility on our DI structure

appcues/src/main/java/com/appcues/Appcues.kt
iujames: yeah great call, we can at least send it through Logcues I assume

appcues/src/main/java/com/appcues/ui/presentation/ViewPresenter.kt
andretortolano: Thats right, this was to help with some inconsistencies when we had multiple activities (particularly transparent)

andretortolano (approved):

appcues/src/main/java/com/appcues/Appcues.kt
iujames: Simple update in this commit https://github.com/appcues/appcues-android-sdk/pull/648/commits/701ec8cf7e55bdc37336849703effb0119e8a968

So they show up in the detailed log like this:

Screenshot 2024-10-21 at 10 08 13 AM

thanks

PR #652, opened 2024-11-08
🐛 Cleanup icons-extension and use local references

andretortolano:
Reported by customer, there was an increase of SDK size because during our Material Compose update we included icons-extension which is heavy.

```
implementation "androidx.compose.material:material-icons-extended:$compose_version"
```

It is possible to keep the library and exclude all icons but the ones we are using, for a simple solution we decided to have our own local icons and avoid including an extra dependency instead.

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/ui/AppcuesSearchView.kt
+++ b/appcues/src/main/java/com/appcues/debugger/ui/AppcuesSearchView.kt
@@ -10,14 +10,13 @@ import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.text.BasicTextField
 import androidx.compose.foundation.text.KeyboardActions
 import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.material.Icon
 import androidx.compose.material.Text
-import androidx.compose.material.icons.Icons
-import androidx.compose.material.icons.filled.Close
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.MutableState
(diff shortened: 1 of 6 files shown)


appcues/src/main/java/com/appcues/debugger/ui/AppcuesSearchView.kt
```diff
@@ -75,7 +73,10 @@ internal fun AppcuesSearchView(
             .border(1.dp, border.value, RoundedCornerShape(cornerDp.value))
             .background(LocalAppcuesTheme.current.background)
             .height(height)
-            .clickable(interactionSource = MutableInteractionSource(), indication = null) { focusRequester.requestFocus() },
+            .clickable(
+                interactionSource = remember { MutableInteractionSource() },
```
andretortolano, line 77: fixing [warning](https://googlesamples.github.io/android-custom-lint-rules/checks/UnrememberedMutableInteractionSource.md.html) about MutableInteractionSource created outside of remember.

appcues/src/main/java/com/appcues/debugger/ui/AppcuesSearchView.kt
```diff
@@ -136,27 +136,26 @@ private fun BoxScope.SearchViewOverlay(
...
-                isFocusOn.value = false
-                keyboardController?.hide()
-                focusManager.clearFocus()
-            }
-            .padding(8.dp)
-
         Icon(
-            modifier = iconModifier,
-            imageVector = Icons.Default.Close,
+            modifier = Modifier
```
andretortolano, line 140: Small change here, before using `Modifier.Companion` for no reason

appcues/src/main/java/com/appcues/debugger/ui/icons/CloseIcon.kt
```diff
@@ -0,0 +1,31 @@
+package com.appcues.debugger.ui.icons
+
+import androidx.compose.ui.graphics.vector.ImageVector
+
+private var filled: ImageVector? = null
+
+@Suppress("UnusedReceiverParameter", "MagicNumber")
```
andretortolano, line 7: UnusedReceiverParameter because of "DebuggerIcons.Filled"
and MagicNumber because of all the path values.

iujames: could be nice to have the rationale in a code comment for later reference on those

iujames (approved):
nice update to remove the extra dependency + size

PR #659, opened 2025-02-03
✅ Enable screen shots tests for debugger

andretortolano:
Small increment on Screenshot tests, debugger screens now have fixed version number and dates, so they can be screenshot tested.

### DIFF
--- a/appcues/src/main/java/com/appcues/AppcuesConfig.kt
+++ b/appcues/src/main/java/com/appcues/AppcuesConfig.kt
@@ -100,4 +100,6 @@ public data class AppcuesConfig internal constructor(
      * a different package than your main app (multi-module apps)
      */
     var packageNames: List<String> = arrayListOf()
+
+    internal var isSnapshotTesting: Boolean = false
 }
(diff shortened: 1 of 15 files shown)


appcues/src/main/java/com/appcues/debugger/DebuggerRecentEventsManager.kt
```diff
@@ -111,12 +114,21 @@ internal class DebuggerRecentEventsManager(
         lastEventId++
     }
 
+    private fun ActivityRequest.getIdentifyTimeStamp(): Long {
+        return if (!appcuesConfig.isTesting) {
+            // Unsure why we need to do this, we could just get current date or ActivityRequest.timeStamp
```
iujames, line 119: I don't know either, it seems like the ActivityRequest timestamp would be accurate for this.

appcues/src/main/java/com/appcues/AppcuesConfig.kt
```diff
@@ -100,4 +100,6 @@ public data class AppcuesConfig internal constructor(
      * a different package than your main app (multi-module apps)
      */
     var packageNames: List<String> = arrayListOf()
+
+    internal var isTesting: Boolean = false
```
iujames, line 104: wondering if naming like `isSnapshotTesting` would make the usage of this even more clear

iujames (approved):
LGTM, nice to have a pretty isolated fix here to enable snapshot testing the Debugger UI in general - will be a nice coverage improvement.

appcues/src/main/java/com/appcues/debugger/DebuggerRecentEventsManager.kt
andretortolano: not sure why we do it like this, I think we can clean it up and use the activity request timestamp instead too.

PR #660, opened 2025-03-07
Trim any leading hyphen on form response generated property names

iujames:
The hyphen in the scenario described in the ticket was due to replacing of special char `*` with empty space ` `, then splitting on empty space, but not filtering out empty strings, then joining with a `-`. In this case, joining an empty string to the front with a hypen.

Fix here is to (A) filter out empty strings in the join, and (B) trim any leading hyphen for good measure.

| before | after | 
| --- | --- |
|  before |  after |

### DIFF
--- a/appcues/src/main/java/com/appcues/util/StringExt.kt
+++ b/appcues/src/main/java/com/appcues/util/StringExt.kt
@@ -6,9 +6,11 @@ internal fun String.toSlug() = lowercase(Locale.getDefault())
     .replace("\n", " ")
     .replace("[^a-z\\d\\s]".toRegex(), " ")
     .split(" ")
+    .filter { it.isNotEmpty() }
     .joinToString("-")
     .replace("-+".toRegex(), "-")
     .trimEnd('-')
+    .trimStart('-')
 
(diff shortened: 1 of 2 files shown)


appcues/src/main/java/com/appcues/util/StringExt.kt
```diff
@@ -6,9 +6,11 @@ internal fun String.toSlug() = lowercase(Locale.getDefault())
     .replace("\n", " ")
     .replace("[^a-z\\d\\s]".toRegex(), " ")
     .split(" ")
+    .filter { it.isNotEmpty() }
     .joinToString("-")
     .replace("-+".toRegex(), "-")
```
mmaatttt, line 11: I think the filter makes this regex redundant since there should never be `--` any more? Probably fine to leave though.

mmaatttt (approved):

appcues/src/main/java/com/appcues/util/StringExt.kt
iujames: yeah, I agree it should really be redundant now. I'll leave it just to be safe I suppose.

PR #663, opened 2025-03-17
simplify primitive deserialization

iujames:
This change resolves an issue that a customer observed where minified builds (Proguard) would fail to render an experience. The issue resulted in a JSON deserialization error, from Moshi, the JSON library we use.

```
com.squareup.moshi.JsonDataException: Expected one of [STACK, BOX, CUSTOM_COMPONENT, BUTTON, TEXT, IMAGE, EMBED, SPACER, TEXT_INPUT, OPTION_SELECT, BLOCK] but was text at path $.steps[0].children[0].content.items[0].items[0].content.errorLabel.type
```

Tracing through the problem and the customer provided build configuration. I was able to reproduce by using the `-dontshrink` option in a sample app proguard-rules.pro file, and building against a matching Kotlin version `2.1.0` (our core sdk still builds against the older `1.8.22`) - using a release config build with `isMinifyEnabled = true`.

Here is what appears to be happening
* We use a `PolymorphicJsonAdapterFactory` from Moshi [here](https://github.com/appcues/appcues-android-sdk/blob/main/appcues/src/main/java/com/appcues/data/MoshiConfiguration.kt#L82-L93), to deserialize the correct primitive type, based on the `"type"` property in the JSON coming in
* The `PrimitiveResponse` base class has a [type](https://github.com/appcues/appcues-android-sdk/blob/main/appcues/src/main/java/com/appcues/data/remote/appcues/response/step/primitive/PrimitiveResponse.kt#L21) field that references this [Type enum](https://github.com/appcues/appcues-android-sdk/blob/main/appcues/src/main/java/com/appcues/data/remote/appcues/response/step/primitive/PrimitiveResponse.kt#L24-L36) in the same file
* That `PrimitiveResponse.Type` has a `jsonName` field that is the text representation in the JSON model that maps to each enum value - ex: `"text"` maps to enum value `TEXT`.
* As we deserialize, most of the base `PrimitiveResponse` values work ok still, but eventually we get to a nested `TextPrimitiveResponse` -- see the path in the error `$.steps[0].children[0].content.items[0].items[0].content.errorLabel` - this corresponds to an `OptionSelectPrimitiveResponse.errorLabel` [here](https://github.com/appcues/appcues-android-sdk/blob/main/appcues/src/main/java/com/appcues/data/remote/appcues/response/step/primitive/PrimitiveResponse.kt#L110) - which is a nested `TextPrimitiveResponse` -- a specific derived type, not the base class `PrimitiveResponse` that the Moshi adapter is configured for
* When it tries to deserialize this type (I believe using a default adapter for that type, not the custom adapter for the base type), for some reason it is unable to map `"text"` the the `type` property (the enum) on the base class -- and this only happens in the minified build with the `-dontshrink` option. Something about how Proguard is optimizing/renaming/moving things is breaking this

The fix: we don't actually need this `type` property on the base class at all, it is never used anywhere else in the code. We can simplify our `PolymorphicJsonAdapterFactory` to just pick the correct primitive type based on the string value, and remove this `Type` enum entirely. This solves the problem.

Why did this work before? I don't know. It seems to work in our test app with an earlier version of Kotlin, even when that proguard rule is applied. So something changed, and it is unclear if the prior behavior was simply due to luck at this point. But, I think this implementation is simpler and does not rely on any unclear deserialization side effects - resolves the issue.

### DIFF
--- a/appcues/src/main/java/com/appcues/data/MoshiConfiguration.kt
+++ b/appcues/src/main/java/com/appcues/data/MoshiConfiguration.kt
@@ -80,16 +80,16 @@ internal object MoshiConfiguration {
 
     private fun getPrimitiveFactory(): Factory {
         return PolymorphicJsonAdapterFactory.of(PrimitiveResponse::class.java, "type")
-            .withSubtype(BoxPrimitiveResponse::class.java, PrimitiveResponse.Type.BOX.jsonName)
-            .withSubtype(CustomComponentPrimitiveResponse::class.java, PrimitiveResponse.Type.CUSTOM_COMPONENT.jsonName)
-            .withSubtype(ButtonPrimitiveResponse::class.java, PrimitiveResponse.Type.BUTTON.jsonName)
-            .withSubtype(EmbedPrimitiveResponse::class.java, PrimitiveResponse.Type.EMBED.jsonName)
-            .withSubtype(ImagePrimitiveResponse::class.java, PrimitiveResponse.Type.IMAGE.jsonName)
-            .withSubtype(OptionSelectPrimitiveResponse::class.java, PrimitiveResponse.Type.OPTION_SELECT.jsonName)
-            .withSubtype(StackPrimitiveResponse::class.java, PrimitiveResponse.Type.STACK.jsonName)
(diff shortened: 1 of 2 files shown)


mmaatttt (approved):
Wow, nice find and fix!

PR #666, opened 2025-04-14
Update to AGP 8.9.1

iujames:
Staying current here on build tooling, also for my local builds the current codebase is very slow to build in recent releases of Android Studio - this update seems to help that.


mmaatttt (approved):
I'm updated to `Android Studio Meerkat | 2024.3.1 Patch 1` now, so looking good!

PR #669, opened 2025-04-25
Improve push open handling

iujames:
It was discovered that on cross platform frameworks, the push open tracking was not working as expected. The reason for this is the usage of Intent `extras` to pass the push details. 

The simplest approach I could think of to improve this was to just move to a URL query parameter approach to pass the extras, which accomplishes the same result but is cross-platform friendly with things like React Native linking - where the URL is already passed back to the JS side of the app and processed through normal deep link handling (passing into our SDK). This seems to work fine and allow us to avoid any other changes on the x-plat sides.

Here is the end result in React Native using an updated local library for the native dependency - Push opens, tracks analytics, and launches a linked flow:

https://github.com/user-attachments/assets/cde4b4f7-e680-46cd-950e-c6c8846492eb

Also did some touch up on the debugger recent events view here, to (A) give a friendly name for the `Push Opened` event and (B) fix up some details around how the floating events appear and fade out, to make this behave more smoothly.

Before this change, the visibility changes were not quite right, and things would hang until a new event came through. Also, all items were hidden at once, when the final item expired, not on a per item basis - video below shows the before:

https://github.com/user-attachments/assets/4e777df3-dc10-4a63-858f-cf1adfc6fb8a

After this change, the items start to fade correctly after 1 sec, with a smooth fade, then disappear after 1 more second - on a per item basis. Video below shows the after:

https://github.com/user-attachments/assets/0db28456-9dd0-47b7-a2a6-c5197bd79958

### DIFF
--- a/appcues/src/main/java/com/appcues/push/PushDeeplinkHandler.kt
+++ b/appcues/src/main/java/com/appcues/push/PushDeeplinkHandler.kt
@@ -3,7 +3,6 @@ package com.appcues.push
 import android.content.Context
 import android.content.Intent
 import android.net.Uri
-import android.os.Bundle
 import android.widget.Toast
 import com.appcues.AppcuesFirebaseMessagingService.AppcuesMessagingData
 import com.appcues.R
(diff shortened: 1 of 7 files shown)


appcues/src/main/java/com/appcues/push/PushDeeplinkHandler.kt
iujames, line 1: This file is the key change here. No longer using `extras` on the intent triggered from a push tap, but encoding them in query parameters - cross-platform framework friendly approach.

mmaatttt (approved):
Nice solution!

PR #670, opened 2025-04-25
Update publishing scripts for Maven Central Portal migration

iujames:
This is a required migration before ~June this year to change how Android libraries are published, per https://central.sonatype.org/faq/what-is-different-between-central-portal-and-legacy-ossrh/ . I've already handled the steps noted there to do the `Self-Service Migration` and our `com.appcues` namespace is all set up on the new portal now.

To use this new publishing destination, we need to remove our old publish script and replaced it with some new updates that use a different plugin https://github.com/vanniktech/gradle-maven-publish-plugin, to accomplish similar tasks as before, but publish to the Maven Central Portal instead of Legacy OSSRH. There is [no official plugin](https://central.sonatype.org/publish/publish-portal-gradle/) for this, but this one seems to be the popular choice.

Usage of the plugin is documented here https://vanniktech.github.io/gradle-maven-publish-plugin/central/ , and there are new environment variables needed for the secrets with [names defined here in docs](https://vanniktech.github.io/gradle-maven-publish-plugin/central/#secrets). The key we use for our account stays the same as before though.

I did a little file cleanup/simplification where I could along the way, and a test publish of an alpha artifact worked as expected. The command that ends up running it all is below, but this will just work in the updated release script included.
```bash
./gradlew publishToMavenCentral
```


appcues/build.gradle
```diff
@@ -124,3 +129,35 @@ dependencies {
     testImplementation 'androidx.arch.core:core-testing:2.2.0'
     testImplementation 'app.cash.turbine:turbine:1.0.0'
 }
+
+mavenPublishing {
```
iujames, line 133: for some reason, it seems this block _must_ be in this build.gradle file, not applied from another file like we did before. Thus, some of the common property usage was split out in the `appcues-properties.grandle` and used here to keep the additions as minimal as possible. The old `publish-maven.gradle` is then no longer needed at all.

iujames:
can ignore the one test failure here, it is the item that is fixed in the other PR / test repo (other branch)

mmaatttt (approved):

PR #671, opened 2025-04-25
Update test date precision

iujames:
Use exact millisecond precision on the test data so the snapshot tests don't have small differences on [this screen](https://github.com/appcues/appcues-mobile-experience-spec/blob/main/verification-suites/android/snapshots/DebuggerTests/testDebuggerPanel.3.png)


iujames:
closing this one and just including the one-liner for tests in this PR https://github.com/appcues/appcues-android-sdk/pull/669 to test it all together.

PR #681, opened 2025-07-18
Capture target elements from web views

mmaatttt:
Sibling to https://github.com/appcues/appcues-ios-sdk/pull/597

### DIFF
--- a/appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
+++ b/appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
@@ -6,6 +6,7 @@ import android.graphics.Rect
 import android.util.Log
 import android.view.View
 import android.view.ViewGroup
+import android.webkit.WebView
 import androidx.compose.ui.semantics.SemanticsNode
 import androidx.compose.ui.semantics.SemanticsOwner
 import androidx.compose.ui.semantics.SemanticsProperties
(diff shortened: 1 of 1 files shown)


appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
```diff
@@ -176,6 +193,64 @@ private fun View.asCaptureView(screenBounds: Rect): ViewElement? {
...
+
+private suspend fun WebView.evaluateJavascript(script: String): List<Map<String, Any>> {
+    val completion = CompletableDeferred<List<Map<String, Any>>>()
+    evaluateJavascript(script) { result ->
+        val listType: Type = Types.newParameterizedType(
+            List::class.java,
+            Types.newParameterizedType(Map::class.java, String::class.java, Any::class.java)
+        )
+        val adapter = MoshiConfiguration.moshi.adapter<List<Map<String, Any>>>(listType)
+        val parsed: List<Map<String, Any>> = adapter.fromJson(result)!!
```
iujames, line 248: took a quick peek to see if there was some simpler way to express this for the decoding, but no luck! I guess this is the way.

iujames (approved):
looking great!

appcues/src/main/java/com/appcues/debugger/screencapture/AndroidTargetingStrategy.kt
mmaatttt: Reminds me I was going to revisit the non-null assertion. I think this would be preferable?
```kt
val parsed: List<Map<String, Any>> = adapter.fromJson(result) ?: emptyList()
```

iujames: yeah agreed that would be a safer approach

PR #691, opened 2025-09-15
Update safe area handling

mmaatttt:
If we set the safe area layout margins at the root, that means the backdrop will not extend edge-to-edge. Instead we can manually apply the safe area depending on the modal pattern type.


mmaatttt:
After opening the PR I discovered I can just do `WindowInsets.systemBars.asPaddingValues()` instead of the more complex [approach](https://github.com/appcues/appcues-android-sdk/blob/18f5174d6b58016a27576e87b9a91843cf0beea5/appcues/src/main/java/com/appcues/trait/appcues/TargetElementTrait.kt#L46) I found elsewhere

iujames (approved):
makes sense, double checking - do tooltips already apply insets like this?

mmaatttt:
@iujames the tooltip trait uses the updated `OverlayViewPresenter` so it works now:

Screenshot_20250916_100311

PR #696, opened 2025-09-22
Fix tooltip padding calculations relative to the safe area boundaries

iujames:
Resolves issue found in testing where a tooltip pointed at an element on the bottom of the screen may not be positioned far enough in the vertical direction.

| before | after |
| --- | --- |
|  Screenshot_20250922-110028 (1) |  Screenshot_20250922-123600 |


mmaatttt (approved):
The comments are helpful!

PR #714, opened 2026-07-23
:bug: Fix NoSuchMethodError on devices missing PackageManager.*Flags.of()

jcz0r-hcp:
# Fix `NoSuchMethodError` crash on OEM Android 13+ devices missing `PackageManager.*Flags.of()`

## What

Guards the three API-33 helpers in `PackageManagerExt.kt` (`getPackageInfoCompat`, `getActivityInfoCompat`, `resolveActivityCompat`) so a missing `*Flags.of(long)` factory method falls back to the pre-33 overload instead of crashing the host app.

## Why

`PackageManagerExt` calls the API-33 `PackageManager.ComponentInfoFlags.of(long)` / `PackageInfoFlags.of(long)` / `ResolveInfoFlags.of(long)` factories, guarded only by `Build.VERSION.SDK_INT >= TIRAMISU`.

Some OEM Android 13 builds report `SDK_INT == 33` but ship a `/system/framework/framework.jar` that does **not** contain those `of(long)` factory methods. On such devices the call throws `java.lang.NoSuchMethodError`.

This is reached on every app launch through automatic screen tracking: `ActivityScreenTracking.onActivityStarted` -> `getActivityInfoCompat`. That callback only catches `PackageManager.NameNotFoundException`, and `NoSuchMethodError` is an `Error` (not an `Exception`), so it is unguarded and becomes a **fatal crash on the first Activity start at cold launch** - the user cannot open the app at all.

Observed stack trace:

```
java.lang.NoSuchMethodError: No static method of(J)Landroid/content/pm/PackageManager$ComponentInfoFlags;
in class Landroid/content/pm/PackageManager$ComponentInfoFlags; or its super classes
(declaration of 'android.content.pm.PackageManager$ComponentInfoFlags' appears in /system/framework/framework.jar)
    at com.appcues.util.PackageManagerExtKt.getActivityInfoCompat (PackageManagerExt.kt:23)
    at com.appcues.analytics.ActivityScreenTracking.onActivityStarted (ActivityScreenTracking.kt:34)
    at android.app.Application.dispatchActivityStarted (Application.java:614)
    ...
```

Reproduced on **Samsung Galaxy A03s (`SM-A037U`), Android 13**. This is the well-known class of OEM defect where a device advertises API 33 but ships an incomplete `framework.jar` missing some of the API-33 additions.

## How

Each helper now wraps the API-33 path in `try/catch (NoSuchMethodError)` and falls back to the deprecated int-flags overload, which still works on the affected firmware. No behavior change on devices where the API-33 methods are present.

```kotlin
if (sdkInt >= Build.VERSION_CODES.TIRAMISU) {
    try {
        getActivityInfo(componentName, PackageManager.ComponentInfoFlags.of(flags.toLong()))
    } catch (_: NoSuchMethodError) {
        @Suppress("DEPRECATION") getActivityInfo(componentName, flags)
    }
} else {
    @Suppress("DEPRECATION") getActivityInfo(componentName, flags)
}
```

## Tests

Adds `PackageManagerExtTest` (JUnit + MockK + Truth, no new deps) covering, for each helper:

- API-33 path throws `NoSuchMethodError` -> falls back to the legacy overload (fails without this change).
- API-33 path succeeds -> uses the new overload (no regression).
- below API 33 -> uses the legacy overload.

## Notes for reviewers

- The helpers gained an `internal`-only, defaulted `sdkInt: Int = Build.VERSION.SDK_INT` parameter purely as a test seam, since the project's unit tests do not run under Robolectric (no way to drive the `SDK_INT` branch otherwise). Public/consumer behavior is unchanged. Happy to swap this for another approach if you prefer.
- The catch is scoped narrowly to `NoSuchMethodError` to match the reported failure; it could be widened to `LinkageError` if you want to also cover `NoSuchFieldError`-style OEM gaps.
- Optional defense in depth (not included here to keep the change minimal): broaden the `catch` in `ActivityScreenTracking.onActivityStarted` from `NameNotFoundException` to `Throwable`, so a `PackageManager` failure can never crash the host app from a lifecycle callback.

## Checklist

- [x] Branch targets `main`
- [x] Bug fix includes tests that fail without the fix
- [x] `internal` API only; no public API change
- [ ] `./gradlew detekt` (please run in CI - not executed in my environment)
- [ ] Documentation - n/a (internal helper)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Android compatibility on API 33+ devices when framework flag factory methods are missing or unavailable due to OEM/framework differences.
  * Added safer handling to prevent runtime crashes and ensure the appropriate legacy behavior is used as a fallback.

* **Tests**
  * Added new unit tests validating correct behavior across API levels and both cases: presence vs absence of the newer flag APIs.


iujames (approved):
Thanks for the explanation and additional safeguards here @jcz0r-hcp - this looks good to me. We can work on publishing an updated patch version with this update. Appreciate your contribution!

